PR #15, opened 2019-04-15
decompose Components for keeping browser bot session

guinii:
Add new function in order to decompose all types of components and store it in the localStorage of the browser

### DIFF
--- a/packages/botonic-react/src/components/carousel.jsx
+++ b/packages/botonic-react/src/components/carousel.jsx
@@ -1,7 +1,26 @@
 import React from 'react'
+import { Message } from './message'
 import { isBrowser, isNode } from '@botonic/core'
 
 export const Carousel = props => {
+  let content = ''
+  if (isBrowser())
+    content = (
+      <div
+        style={{
+          paddingTop: '10px',
+          display: 'flex',
+          flexDirection: 'row',
+          overflowX: 'auto',
+          maxWidth: '400px',
+          fontFamily: 'Arial, Helvetica, sans-serif'
+        }}
+      >
+        {props.children}
+      </div>
+    )
+  return <Message type='carousel'>{content}</Message>
+} /*
   const renderBrowser = () => (
     <div
       style={{
@@ -21,4 +40,4 @@ export const Carousel = props => {
 
   if (isBrowser()) return renderBrowser()
   else if (isNode()) return renderNode()
-}
+}*/
--- a/packages/botonic-react/src/components/carousel.jsx
+++ b/packages/botonic-react/src/components/carousel.jsx
@@ -4,7 +4,7 @@ import { isBrowser, isNode } from '@botonic/core'
 
 export const Carousel = props => {
   let content = ''
-  if (isBrowser())
+  if (isBrowser()) {
     content = (
       <div
         style={{
@@ -18,7 +18,7 @@ export const Carousel = props => {
       >
         {props.children}
       </div>
-    )
+    ) }
   return <Message type='carousel'>{content}</Message>
 } /*
   const renderBrowser = () => (
--- a/packages/botonic-react/src/components/message.jsx
+++ b/packages/botonic-react/src/components/message.jsx
@@ -2,6 +2,7 @@ import React, { useState, useEffect, useContext } from 'react'
 
 import { isBrowser, isNode } from '@botonic/core'
 import { WebchatContext, RequestContext } from '../contexts'
+import { decomposeComponent } from '../utils'
 import { Button } from './button'
 import { Reply } from './reply'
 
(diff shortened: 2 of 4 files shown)


packages/botonic-react/src/components/carousel.jsx
```diff
@@ -1,7 +1,26 @@
 import React from 'react'
+import { Message } from './message'
 import { isBrowser, isNode } from '@botonic/core'
 
 export const Carousel = props => {
+  let content = ''
+  if (isBrowser())
```
dpinol:
```suggestion
  if (isBrowser()) {
```

packages/botonic-react/src/components/carousel.jsx
```diff
@@ -1,7 +1,26 @@
...
+          display: 'flex',
+          flexDirection: 'row',
+          overflowX: 'auto',
+          maxWidth: '400px',
+          fontFamily: 'Arial, Helvetica, sans-serif'
+        }}
+      >
+        {props.children}
+      </div>
+    )
```
dpinol:
```suggestion
    ) }
```

packages/botonic-react/src/utils.js
```diff
@@ -8,6 +11,47 @@ import { Video } from './components/video'
 import { Document } from './components/document'
 import { Location } from './components/location'
 
+export function decomposeComponent(component) {
+  let componentJSON = null
+  let carousel = []
+  try {
+    component[0].props.children.map((e, i) => {
```
dpinol: noob question:
is it possible that component has more than 1 element?

packages/botonic-react/src/utils.js
```diff
@@ -8,6 +11,47 @@ import { Video } from './components/video'
 import { Document } from './components/document'
 import { Location } from './components/location'
 
+export function decomposeComponent(component) {
```
dpinol: "component" argument is actually a children array, right?
I think it's not obvious what/why this method does.
Would childrenToJson be a good name?
Would it make sense to create a ComponentJSON class with a to toJson and fromJson (with the msgToBotonic code) methods?

packages/botonic-react/src/components/message.jsx
```diff
@@ -28,13 +29,13 @@ export const Message = props => {
   const textChildren = React.Children.toArray(children).filter(
```
dpinol: not sure if it would make sense to move this method within decomposeComponent

packages/botonic-react/src/utils.js
```diff
@@ -8,6 +11,47 @@ import { Video } from './components/video'
...
+        title: c[1].props.children,
+        subtitle: c[2].props.children,
+        button: c[3].props
+      }
+    })
+  } catch (e) {}
+  if (carousel.length) {
+    componentJSON = carousel
+  }
+  try {
```
dpinol: Maybe we can have a single try/catch for all cases?
Would a switch be more clear?

packages/botonic-react/src/utils.js
guinii: I have tried different methods and I could not reproduce a case when a component had more than 1 element was with the Location, but i just refactor the Location component because we are only using a URL button and the component had trash code. Now, all the components should only have 1 element

packages/botonic-react/src/utils.js
guinii: I think childrenToJson it's not a good name, because in this case we are giving a children Component to decompose it into JSON format, but not always has to be a children. For example, in the future, we maybe want to transform a component into a componentJSON, not always has to be a children.

I think It's a good approach to create a ComponentJSON in order to encapsulate all this logic, I'll add it for a new PR!

packages/botonic-react/src/components/message.jsx
guinii: This method it's used in the message component in order to display only the component without buttons and/or replies, it doesn't make any modification in the Component or the childrens. So I think it's okai to leave it here so we are using it only in the render of the Message component.

packages/botonic-react/src/utils.js
guinii: Sure, I am going to refactor the code with a switch and we will only need a single try/catch. Good point!

packages/botonic-react/src/components/carousel.jsx
dpinol: forget about this one if you havent' applied, not sure if "prettier" wants it like this :-)

packages/botonic-react/src/utils.js
dpinol: cool, thanks

packages/botonic-react/src/utils.js
dpinol: ok, thanks

packages/botonic-react/src/components/message.jsx
dpinol: ok

packages/botonic-react/src/utils.js
dpinol: 👍

packages/botonic-react/src/components/carousel.jsx
guinii: Yes, prettier accept it! I will add it now ;)

packages/botonic-react/src/utils.js
```diff
@@ -14,42 +14,41 @@ import { Location } from './components/location'
...
+      case 'embed':
+        componentJSON = component[0].props.src
+        break
+      case 'a':
+        componentJSON = component[0].props.href
+        break
+    }
+    return componentJSON
+  } catch (e) {
+    console.log(`Error decomposing Component ${e}`)
```
dpinol: maybe it would help adding the component type in the log

PR #16, opened 2019-04-16
feat: Support for whatsapp webviews

Khaeshah:
Redirects to chat when closing a webview on WhatsApp

### DIFF
--- a/packages/botonic-react/src/webview.jsx
+++ b/packages/botonic-react/src/webview.jsx
@@ -41,7 +41,7 @@ class App extends React.Component {
             }
         }
         if (this.state.session.user.provider === 'whatsappnew') {
-            location.href = 'https://wa.me/' + this.state.session.user.imp_id, '_self';
+            location.href = 'https://wa.me/' + this.state.session.user.imp_id;
         }
         else window.WebviewSdk.close(() => {}, err => console.log(err))
     }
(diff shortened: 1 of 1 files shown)


packages/botonic-react/src/webview.jsx
```diff
@@ -40,7 +40,10 @@ class App extends React.Component {
                 console.log(e)
             }
         }
-        window.WebviewSdk.close(() => {}, err => console.log(err))
+        if (this.state.session.user.provider === 'whatsappnew') {
+            window.open('https://wa.me/' + this.state.session.user.imp_id, '_self');
```
ericmarcos: use location.href

packages/botonic-react/src/webview.jsx
```diff
@@ -40,7 +40,10 @@ class App extends React.Component {
                 console.log(e)
             }
         }
-        window.WebviewSdk.close(() => {}, err => console.log(err))
+        if (this.state.session.user.provider === 'whatsappnew') {
```
dpinol: in the future, it would be nice to have constants for the provider names to avoid repeating the whatapp-gate :-)

packages/botonic-react/src/webview.jsx
```diff
@@ -40,7 +40,10 @@ class App extends React.Component {
...
+            location.href = 'https://wa.me/' + this.state.session.user.imp_id, '_self';
```
ericmarcos: Possible bug: remove `, '_self'`

Khaeshah: Solved

PR #17, opened 2019-04-16
Webchat SDK connected to Hubtype

guinii:
In this PR you will find that the webchat is getting the bot responses throw a post call into Hubtype API and a socket listens to the response.
Once we get the response, Botonic transforms the data into Botonic Components that are rendered in the webchat.

The pusher connected to Hubtype is identified by a UUID created for the user, and the appId that we can obtain when we integrate in 'app.hubtype.com' the bot with the Botonic webchat SDK.

We can test this integration in the 'serve' command passing the appId argument in the file 'app.js' as an argument.
   ex: const appId = 'YOUR_APP_ID'
           //export default new App({ appId })
Another way is when you have deployed your bot and you want to embed it into your html page, you have to inject the code you will find when you integrate your bot in 'app.hubtype.com'


packages/botonic-react/src/utils.js
```diff
@@ -82,19 +86,50 @@ export function msgToBotonic(msg) {
...
+      return (
+        <Reply key={i} payload={el.payload}>
+          {el.text}
+        </Reply>
+      )
+    })
+  }
+  if (msg.keyboard) {
+    msg.keyboard.map(el => {
+      return <Reply payload={el.data}>{el.label}</Reply>
```
dpinol: noob question: do we need a key attribute here?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -56,6 +73,14 @@ export const Webchat = props => {
       }
       if (devSettings) updateDevSettings(devSettings)
     } catch (e) {}
+    if (appId) {
+      pusher = new Pusher('da85029877df0c827e44')
+      if (!Object.keys(pusher.channels.channels).length) {
+        //TODO: remove my endpoint, generic one with socket
```
dpinol: is this done? otherwise, should we create a ticket to track it?

packages/botonic-react/src/utils.js
ericmarcos: yes, it's a React thing: https://reactjs.org/docs/lists-and-keys.html

packages/botonic-react/src/app.jsx
```diff
@@ -17,24 +17,29 @@ export class App {
...
     defaultTyping,
     defaultDelay
   }) {
+    this.plugins = loadPlugins(plugins)
+    this.theme = theme
+    this.defaultTyping = defaultTyping || 0
+    this.defaultDelay = defaultDelay || 0
+    this.locales = locales
+    this.integrations = integrations
+    if (appId) {
```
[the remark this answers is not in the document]
ericmarcos: you're right, this will need a refactor at some point, but it's ok for now.

packages/botonic-react/src/app.jsx
```diff
@@ -53,7 +58,8 @@ export class App {
     else return <Webchat botonicApp={this} theme={themeOptions} />
   }
 
-  render(dest, webchatOptions) {
+  render({ dest, webchatOptions, appId }) {
```
ericmarcos: Dangerous: we're breaking the original signature of this function. I agree it's better to pass an object so we can better handle optional params, but this can break projects with older versions of botonic.

packages/botonic-react/src/components/button.jsx
```diff
@@ -63,3 +63,16 @@ export const Button = props => {
...
 }
+
+Button.serialize = buttonProps => {
+  let payload = buttonProps.payload
+  if (buttonProps.path) payload = `__PATH_PAYLOAD__${buttonProps.path}`
+  return {
+    button: {
+      payload,
+      url: buttonProps.url,
+      webview: buttonProps.webview,
```
ericmarcos: Webview is a react component which is not serializable. Not sure how to deal with this. Maybe it would be reasonable to leave it null for now, seems to me that refreshing the screen and buttons not keeping the link to webview is not that big deal. We can look into it in future versions.

packages/botonic-react/src/components/element.jsx
```diff
@@ -20,3 +20,41 @@ export const Element = props => {
...
   else if (isNode()) return renderNode()
 }
+
+Element.serialize = elementProps => {
+  let element = Object.assign(
+    {},
+    ...elementProps.children
+      .filter(c => c && c.type && c.type.name != 'Button')
+      .map(c => c.type.serialize && c.type.serialize(c.props))
+  )
```
ericmarcos: This whole thing is confusing. The result of [...].filter.map() is an Array, then we're spreading it into an object?

packages/botonic-react/src/components/message.jsx
```diff
@@ -31,12 +32,17 @@ export const Message = props => {
...
     useEffect(() => {
-      let decomposedChildren = decomposeComponent(textChildren)
+      let decomposedChildren = json
       let message = {
         id: state.id,
         type,
         data: decomposedChildren ? decomposedChildren : textChildren,
         from,
+        buttons: buttons.map(b => ({
+          payload: b.props.payload || b.props.path,
```
ericmarcos: path is not the same as payload, did you test this?

packages/botonic-react/src/components/text.jsx
```diff
@@ -1,8 +1,29 @@
 import React from 'react'
 import { Message } from './message'
 
+const serialize = textProps => {
+  let text = textProps.children
+  /* As text message can have multiple children ( a text with buttons or quickreplies )
+    we only want to store the text, we don't want to store the buttons or quickreplies.
```
ericmarcos: We don't want to store quick replies, ok. But buttons? I think we should store them, no?

packages/botonic-react/src/webchat/webchat.jsx
guinii: Yes, it's done. I should remove the TODO comment

packages/botonic-react/src/app.jsx
guinii: You are right, It breaks the older versions because we are expecting specific params. We can make as before, ( render(dest, webchatOptions, appId ) and treat the logic inside the function.

packages/botonic-react/src/components/button.jsx
guinii: Right now we can leave it null. Maybe in the future we can try to store the path where we are importing this webview ( need to investigate how to get it from the react component, if posible )

packages/botonic-react/src/components/element.jsx
guinii: Yes, because we an element must be a single object, with their attributes. If an element is an array, when we try to decompose a carousel, we will obtain an array composed of multiple arrays ( one of only one length for every element ): Ex Carousel[ Element1 [{...atributes}], Element2 [{...at}], ...]

packages/botonic-react/src/components/text.jsx
guinii: Yes, we want to store quickreplies and buttons too, but we don't want to decompose here in the text, because in the webchat, we are already dealing with the decomposition of quickreplies and buttons.

packages/botonic-react/src/components/message.jsx
guinii: I will test it!

packages/botonic-react/src/app.jsx
ericmarcos: Maybe we can pass appId as a webchatOption? In this way we keep the original signature and don't add more params to the function. I also think it make sense from a semantic point of view.

guinii: I just made a new push with this logic applied!

dpinol: please rebase to master (my branches are on top of this)

guinii: The branch is up to date with master

ericmarcos:
In the last commit I've made a big refactor with these main goals:

1. To break apart different functionality that was being managed from the same botonic-react `App` class.
2. To remove a circular dependency between App and Webchat classes.
3. To clearly define 4 targets and better separate the code needed by each one: node, browser webchat, browser dev and browser webviews. This helps webpack eliminate dead code and bundle only what's really needed. A good side effect is that libraries that made node bundles crash (because they rely on  `window` or  `document`) are no  longer a problem as  they're not  included in the node  bundle.
4. To move to botonic-core a part of the logic that used to be in botonic-react `App`, as it's framework agnostic  and we could reuse it for other frameworks like Vue or Angular.
5. To better handle webchat theme and introduce webchat callbacks (onInit, onOpen, onClose, onMessage).

Before:

```
┌───────────┐         ┌─────────────┐
│           │         │  Webchat /  │
│    App    ◀─────────▶ WebchatDev  │
│           │         │             │
└───────────┘         └─────────────┘
```

After:

```
     ┌───────────┐   ┌───────────┐   ┌───────────┐  ┌───────────┐
     │           │   │           │   │           │  │           │
     │  NodeApp  │   │  DevApp   │   │WebchatApp │  │WebviewApp │
     │           │   │           │   │           │  │           │
     └───────────┘   └───────────┘   └───────────┘  └───────────┘
           │               │               │                     
           │    ┌──────────┴──────┐        │                     
           │    │                 │        │                     
┌──────────▼────▼──────┐   ┌──────▼──────┐ │   ┌───────────────┐ 
│       ReactBot       │   │             │ │   │               │ 
│   (botonic-react)    │   │ WebchatDev  │ ├──▶│HubtypeService │ 
│                      │   │             │ │   │               │ 
│ ┌──────────────────┐ │   └─────────────┘ │   └───────────────┘ 
│ │                  │ │          │        │                     
│ │     CoreBot      │ │   ┌──────▼──────┐ │                     
│ │  (botonic-core)  │ │   │             │ │                     
│ │                  │ │   │   Webchat   │◀┘                     
│ └──────────────────┘ │   │             │                       
└──────────────────────┘   └─────────────┘                       
```

NOTES:
* HubtypeService encapsulates the communication with Hubtype for the Webchat target. This  piece could be moved to botonic-core too as  it's independent of react.
* The code as it is right now is not backwards compatible. However, as all new classes have new names, I think we can safely re-introduce the old ones to make it compatible again.
* I've introduced a new file in botonic-react: `entry.js` which acts as the entry point of a botonic project, instead of relying on an `app.js` on the user's project directory. This hides this complexity from the user, but it also adds some degree of _magic_ and I'm not sure  what's best.
* We could move the base webpack config to botonic-react too, then in the user's project have a lightweight  webpack.conf.js file that extends the base one and is easily customisable.
* I've introduced a new webpack env var: `BOTONIC_TARGET` that is used by entry.js to select the components that are needed and let webpack eliminate the dead code. I think we could also use this instead of `isDev`, `isBrowser`, `isNode` functions to take further advantage of tree shaking.

PR #18, opened 2019-04-30
Plugin/dynamo project setup

dpinol:
PoC for event tracking plugin to Dynamo. Tracking info will be used for Massimo reporting 
Dynamo access using dynamo-data-mapper ORM


packages/botonic-plugin-track-dynamo/tests/helpers/dynamo.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * At least Dynamo gets the keys from ~/.aws/credentials, but it does not get the region from ~/.aws/config
+ * @param conf
+ */
+export function testConfig(
+  conf?: ServiceConfigurationOptions
+): ServiceConfigurationOptions {
+  if (!conf || !conf.secretAccessKey || !conf.accessKeyId) {
+    let credentials = os.homedir() + '/.aws/credentials';
```
guinii: Right now I don't know if in the lambda function ( is where this code will be executed when the bot is deployed ) we have in this route the  AWS credentials. Maybe we should use Environment Variables that can be stored in the lambda function.

dpinol: this is only for the unit tests. For production, the idea is injecting the keys from the bot definition through the plugin constructor options

ericmarcos:
Good work!

The only thing I'd change is the name of the plugin. I wouldn't tie it to a particular use-case, in this case we're using it to track events, yes, but this plugin should be the place where we centralise all functionality regarding the dynamo integration. I propose `botonic-plugin-dynamodb`, also to keep naming consistency with other plugins like `botonic-plugin-segment`, etc.

dpinol:
@ericmarcos makes sense, I created another PR because I rebased from a different base branch See https://github.com/hubtype/botonic/pull/26

PR #21, opened 2019-05-08
Contentful/memoize

dpinol:
Cache contentful calls to enable looking for keywords in memory in next PR.


dpinol:
https://martinfowler.com/bliki/TwoHardThings.html
caches make everything much complex. 
In this case it's not worth because it takes approx 250ms to get all text & carousel entries. In the future we may cache only this

PR #22, opened 2019-05-10
feat(plugin-dialogflow): Allow doing api calls directly

Khaeshah:
Function that allows calling dialogflow manually inside a bot to change the input of it. 

Example:
```
let newInput = 'this is a new text'
await dialogflow.apiCall(newInput, { input, session, lastRoutePath })
```
It updates the input of the bot on success.

### DIFF
--- a/packages/botonic-plugin-dashbot/package-lock.json
+++ b/packages/botonic-plugin-dashbot/package-lock.json
@@ -1,6 +1,6 @@
 {
-  "name": "botonic-plugin-dashbot",
-  "version": "0.8.0",
+  "name": "@botonic/plugin-dashbot",
+  "version": "0.8.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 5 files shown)


ericmarcos:
There's a few changes I'd make:
- Naming: `apiCall ` looks too generic to me (what API call??), I'd rename it to the actual endpoint that we're calling: `query`. In the future we might want to add other calls.
- Separation of concerns: I'd think as this functions as a low level access to Dialogflow for the Botonic developer, so they can use it in any way they want. I think this method should just call the dialogflow API with the params that are passed and return the raw json response from dialogflow. This function should not know anything about inputs, sessions and such.
- Have the `pre` function refactored to use the new method.

Khaeshah:
`apiCall` is renamed to `query`, and it is separated in a way that developers can call `query` with the required parameters and get the response.

The `pre` function is refactored to use the new method. See [6073129](https://github.com/hubtype/botonic/commit/6073129130df56a39ec8dfd8bf3d32184b74ef59)

PR #23, opened 2019-05-13
doc: update windows instructions

Khaeshah:


docs/2.getting-started/1.installation.mdx
```diff
@@ -55,7 +55,7 @@ In order to develop with Botonic on Windows, you can achieve it by following thi
 
 **Requirements:**
 
-* **[NodeJS](https://nodejs.org/en/download/)**: tested on v8.12.0
+* **[NodeJS](https://nodejs.org/en/download/)**: tested on v11.9.0
```
ericmarcos: It doesn't work with previous versions?

Khaeshah: It works, maybe it would be better to say "tested up to v11.9.0"

ericmarcos: Explicit is better than implicit. Let's put `tested on v8.12, [...] and 11.9`

PR #24, opened 2019-05-13
Typescript/compatibility

dpinol:
- refactor of router to have smaller methods
- compile core & react with babel
- types for typescript for some core & react symbols
- eslint & prettier config


packages/botonic-core/index.d.ts
```diff
@@ -0,0 +1,6 @@
+export interface Route {
+  path: string;
+  payload?: RegExp| string;
+  text?: string;
+  action: React.ReactNode;
+}
```
guinii: If this component is representing the 'router' of Botonic, we need ti add intent variable, I don't remember right now but we had the option of passing an attribute called 'type', where we defined the type of the input, like: 'type: image', or 'type: text'. I don't know if it's necessary to declare it, but to keep in mind.

packages/botonic-react/index.d.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export const RequestContext: React.Context<any>;
+
+export interface Session {
+  __locale: string;
+  __retries?: number;
+}
+
+// plugins
+export interface PluginPreInput {
```
guinii: The plugin, in the pre input, why can't return a response? As it's done in the PluginPostInput?

dpinol: it's actually the type for the pre() method. I added comment in future PR

packages/botonic-core/index.d.ts
dpinol: right! thanks
I'm adding this in future PR
```
type StringMatcher = RegExp| string | ((data: string) => boolean);

export interface Route {
  path: StringMatcher;
  payload?: StringMatcher;
  text?: StringMatcher;
  type?: StringMatcher;
  intent?: StringMatcher;
  session?: (session: any) => boolean;
  action: React.ReactNode;
}
```

PR #31, opened 2019-05-16
Case rating when getting back from human handoff

dpinol:
* add call to backend to store a case rating
* add flag to cms.Text to show buttons as replies


ericmarcos:
Update API call with backend changes (from case to chat)

dpinol:
> Update API call with backend changes (from case to chat)

Was already done in 7d264ab

PR #32, opened 2019-05-20
Nlp/stemmer

dpinol:


packages/botonic-plugin-contentful/src/nlp/node-nlp.ts
```diff
@@ -0,0 +1,13 @@
+//import NlpUtil from 'node-nlp/lib/nlp/nlp-util';
+import stemmer = require('node-nlp/lib/nlp/stemmers/natural/porter-stemmer-es');
```
guinii: Is there a better way to import stemmer?

packages/botonic-plugin-contentful/src/nlp/keywords.ts
```diff
@@ -1,4 +1,5 @@
-import { substringIsBlankSeparated, normalize } from './tokens';
+import { normalize } from './node-nlp';
+import { substringIsBlankSeparated } from './tokens';
 
 class CandidateWithKeywords<M> {
```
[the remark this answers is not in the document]
dpinol: in our case, we use it with an adhoc class ButtonsWithKeywords, but the nice thing of generics (aka templates) can be used with any class (or class with restrictions)
https://github.com/hubtype/botonic/blob/e37a33aafbc7a771a2f4940b973b50c56421e64d/packages/botonic-plugin-contentful/src/keywords/index.ts#L17

I'm a fan of generics. See https://www.typescriptlang.org/docs/handbook/generics.html Soon or later I'll do a presentation

packages/botonic-plugin-contentful/src/nlp/node-nlp.ts
dpinol: looks like "import stemmer from 'node-nlp/lib/nlp/stemmers/natural/porter-stemmer-es';" also works. Good spot! fixed
If you meant about the long path, see my comment at the d.ts file:
"// We use directly the ES Stemmer instead of using NlpUtil to avoid loading dependencies of so many languages
// (there are even errors with 'fs' library and with Japanese dependencies)
"

PR #55, opened 2019-06-03
Read a schedule (eg. for Desk) from Contentful plugin

dpinol:


packages/botonic-plugin-contentful/src/cms/callback.ts
```diff
@@ -0,0 +1,116 @@
+import { CMS, ModelType } from './cms';
+import { Model } from './model';
+import escapeStringRegexp = require('escape-string-regexp');
```
ericmarcos: import + require?

dpinol: umm, no idea why I had it. I changed it to in my current branch
`import escapeStringRegexp from 'escape-string-regexp';
`

PR #56, opened 2019-06-03
feature(react): call new Plugin's init method from loaded

dpinol:
Proposal to allow one plugin to use other plugins.

Plugin would be able to implement an "init(pluginInstances)" method so that they can store them and use them when necessary.
I think it could be interesting that the init method was async so that the plugin can done some async initialization (like connecting to an external service), so that it does not need to be in a lazy fashion. But not implemented because it was not required here.
In the future it would be nice also to pass the bot_id to this method, in case the plugin requires different configuration for each bot integration.

### DIFF
--- a/packages/botonic-react/package.json
+++ b/packages/botonic-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/react",
-  "version": "0.9.0-beta.1",
+  "version": "0.9.0-beta.2",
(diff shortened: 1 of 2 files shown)


ericmarcos:
Could `plugins` be a new param in the pre/post functions and be used there?
I'd try to avoid having constructor + init if possible.

PR #68, opened 2019-06-13
refactor(plugin-dialogflow): Default version for dialogflow queries

Khaeshah:


packages/botonic-plugin-dialogflow/src/index.js
```diff
@@ -3,10 +3,16 @@ import axios from 'axios'
...
+  async query(data, lang, sessionId, version = this.defaultVersion) {
+    // Data, lang, sessionId and version are required parameters
+    // See https://dialogflow.com/docs/reference/agent/query for more information
+
+    if (this.options.version) {
+      version = this.options.version
+    }
     return await axios({
       headers: {
         Authorization: 'Bearer ' + this.options.token
```
ericmarcos: Try to use template strings with backtick \` as much as you can. In this case: `` `Bearer ${this.options.token}` ``

packages/botonic-plugin-dialogflow/src/index.js
```diff
@@ -16,7 +22,7 @@ export default class BotonicPluginDialogflow {
         query: data,
         lang: lang,
         sessionId: sessionId,
-        v: this.options.version
+        v: version
```
ericmarcos: Try to put related logic together, or as close as possible. In this case I have to look at 3 different places in order to know what the value of `version` will be: 1) constructor sets  `defaultVersion`. 2) query() signature uses defaultVersion. 3) there's an `if` that sets `version` before axios call. 4) version is used in axios.

In this case, the module is very simple and it's not a problem. But when modules grow, this is one of the main reasons they become hard to read and to follow the logic.

Instead, we could just do:

`v: this.options.version || this.defaultVersion`

PR #73, opened 2019-06-17
Add types to core/react. storeCaseRating

dpinol:
* chores: Add types for typescript projects which use botonic core & react. Now core & react can be compiled (added dev dependencies), but only its sources will be exported, and so it should not affect pure JS projects.
* chores: eslint & prettier config for core/react
* fix: carousel was not passing its props to Message component
* feature: New storeCaseRating to store the rating given by the customer to the agent

### DIFF
--- a/packages/botonic-core/index.d.ts
+++ b/packages/botonic-core/index.d.ts
@@ -1,0 +1,6 @@
+export interface Route {
+  path: string;
+  payload?: RegExp| string;
+  text?: string;
+  action: React.ReactNode;
+}
--- a/packages/botonic-core/index.d.ts
+++ b/packages/botonic-core/index.d.ts
@@ -1,10 +1,19 @@
+type StringMatcher = RegExp| string | ((data: string) => boolean);
+
 export interface Route {
-  path: string;
-  payload?: RegExp| string | ((input: string) => boolean);
-  text?: RegExp | string | ((input: string) => boolean);
+  path: StringMatcher;
+  payload?: StringMatcher;
+  text?: StringMatcher;
+  type?: StringMatcher;
+  intent?: StringMatcher;
+  session?: (session: any) => boolean;
   action: React.ReactNode;
 }
 
+// handoff
 export declare function humanHandOff(session: any, queue_name: string, on_finish: {payload?: any, path?: any}): Promise<void>;
-
 export declare function getOpenQueues(session: any): Promise<{queues: string[]}>;
+export declare function storeCaseRating(session: any, rating: number): Promise<any>;
+
+
+export declare function storeCaseRating(session: any, rating: number): Promise<void>;
--- a/packages/botonic-core/index.d.ts
+++ b/packages/botonic-core/index.d.ts
@@ -1,19 +1,39 @@
 type StringMatcher = RegExp| string | ((data: string) => boolean);
 
+export interface Session {
+  is_first_interaction?: boolean,
+  last_session?: any,
+  user: {
+    id: string
+    // login
+    username?: string,
+    // person name
+    name?: string,
+    // whatsapp, telegram,...
+    provider?: string,
+    // The provider's user id
+    provider_id?: string,
+    extra_data?: any
+  },
+  bot: {
+    id: string,
+    name?: string
+  }
+  __locale?: string,
+  __retries?: number
+}
+
 export interface Route {
   path: StringMatcher;
   payload?: StringMatcher;
   text?: StringMatcher;
   type?: StringMatcher;
   intent?: StringMatcher;
-  session?: (session: any) => boolean;
+  session?: (session: Session) => boolean;
   action: React.ReactNode;
 }
 
-// handoff
-export declare function humanHandOff(session: any, queue_name: string, on_finish: {payload?: any, path?: any}): Promise<void>;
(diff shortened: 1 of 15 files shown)


packages/botonic-core/src/handoff.js
```diff
@@ -26,8 +26,9 @@ export async function humanHandOff(session, queue_name, on_finish) {
 
 export async function storeCaseRating(session, rating) {
   let base_url = session._hubtype_api || 'https://api.hubtype.com'
-  const queues_url = `${base_url}/v1/cases/store_case_rating/`
   let chat_id = session.user.id
+  const queues_url = `${base_url}/v1/chats/${chat_id}/store_case_rating/`
```
guinii: Maybe I would change the name of this variables, it is not using any queue nor pointing to a queue_url!

packages/botonic-core/index.d.ts
```diff
@@ -1,10 +1,19 @@
...
 export interface Route {
-  path: string;
-  payload?: RegExp| string | ((input: string) => boolean);
-  text?: RegExp | string | ((input: string) => boolean);
+  path: StringMatcher;
+  payload?: StringMatcher;
+  text?: StringMatcher;
+  type?: StringMatcher;
+  intent?: StringMatcher;
+  session?: (session: any) => boolean;
```
guinii: Can we use the Interface Session instead of the any? Why this parameter is returning a bool?

packages/botonic-core/index.d.ts
```diff
@@ -4,3 +4,7 @@ export interface Route {
   text?: RegExp | string | ((input: string) => boolean);
   action: React.ReactNode;
 }
+
+export declare function humanHandOff(session: any, queue_name: string, on_finish: {payload?: any, path?: any}): Promise<void>;
+
```
guinii: Can we use the Interface Session here too?

packages/botonic-react/index.d.ts
```diff
@@ -0,0 +1,115 @@
...
+
+
+//TODO inherit from MessageProps?
+export interface TextProps{
+  delay?: number;
+  typing?: number;
+}
+export class Text extends React.Component<TextProps, any> {}
+
+export interface ButtonProps{
```
guinii: A button can have a path that it directly redirect to specific Action

packages/botonic-react/index.d.ts
```diff
@@ -0,0 +1,115 @@
...
+  payload?: string;
+  url?: string;
+}
+export class Button extends React.Component<ButtonProps, any> {}
+
+export interface ButtonProps{
```
guinii: This line is the same as line23? Maybe its ReplyProps? If it's ReplyProps it should have an optional variables called path as ButtonProps

packages/botonic-core/package.json
```diff
@@ -19,6 +19,23 @@
...
+    "eslint": "^5.16.0",
+    "eslint-config-prettier": "^4.1.0",
+    "eslint-plugin-no-null": "^1.0.2",
+    "eslint-plugin-prettier": "^3.0.1",
+    "prettier": "^1.17.0"
+  },
+  "homepage": "https://github.com/hubtype/botonic#readme",
+  "engines": {
+    "node": ">=8.10.0"
+  }
```
ericmarcos: Are all this dependencies really needed in core?

packages/botonic-react/index.d.ts
ericmarcos: And a `webview` attribute too.

packages/botonic-react/index.d.ts
```diff
@@ -0,0 +1,137 @@
...
+    provider_id?: string,
+    extra_data?: any
+  },
+  bot: {
+    id: string,
+    name?: string
+  }
+  __locale?: string,
+  __retries?: number
+}
```
ericmarcos: Maybe we can leave the session interface as just an object (any), so we're not rigged to the current data model. This is something that can change a lot and also, everytime the user wants to add properties (which will be very common), he'll go into "casting hell"...

packages/botonic-core/package.json
dpinol: they're just dev dependencies, mostly for babel & linters. Since we don't even really need to build core/react,  I think they're harmless

packages/botonic-react/index.d.ts
dpinol: yes, but it's very useful to have at least the types of the most stable ones. Developers using JS will not be affected, and developers using TS are used casting to any or disabling the warnings (typically via IDE hints)

packages/botonic-core/index.d.ts
dpinol: 👏 
I didn't do it because I had Session in react. I moved it to core and now I use it

packages/botonic-core/index.d.ts
dpinol: same

packages/botonic-react/index.d.ts
dpinol: added webview and path

packages/botonic-react/index.d.ts
dpinol: good spot! fixed

PR #78, opened 2019-06-20
feature(contentful): allow changing MatchType

dpinol:


Khaeshah:
Maybe you should commit the file `package-lock.json`

The file `package.json` has a plugin-contentful version `0.9.0-beta.10`, but package-lock has the version `0.9.0-beta.8`. Maybe this could cause dependency errors

PR #82, opened 2019-06-21
Contentful queue

dpinol:
on behalf of Arnau

### DIFF
--- a/packages/botonic-plugin-contentful/src/contentful/delivery-api.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/delivery-api.ts
@@ -1,13 +1,15 @@
 import { Entry } from 'contentful';
-import * as contentful from 'contentful';
 import { UrlFields } from './url';
 import {
   Callback,
-  ContentCallback,
   CallbackToContentWithKeywords,
+  Content,
+  ContentCallback,
   ModelType
 } from '../cms';
+import { QueueDelivery } from './queue';
 import * as cms from '../cms';
+import * as contentful from 'contentful';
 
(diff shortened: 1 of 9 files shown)


packages/botonic-plugin-contentful/src/contentful/delivery-api.ts
```diff
@@ -88,6 +90,25 @@ export class DeliveryApi {
   static urlFromAsset(assetField: contentful.Asset): string {
     return 'https:' + assetField.fields.file.url;
   }
+
+  async contents(model: ModelType): Promise<Content[]> {
+    let entries = await this.getEntries<ContentWithKeywordsFields>({
```
dpinol: we should throw when type is not content

packages/botonic-plugin-contentful/src/contentful/queue.ts
```diff
@@ -0,0 +1,31 @@
+import * as cms from '../cms';
```
dpinol: missing unit test

packages/botonic-plugin-contentful/src/contentful/schedule.ts
```diff
@@ -19,16 +25,16 @@ export class ScheduleDelivery {
...
   ): void {
     let days = [
-      fields.sundays,
-      fields.mondays,
-      fields.tuesdays,
-      fields.wednesdays,
-      fields.thursdays,
-      fields.fridays,
-      fields.saturdays
+      fields.sundays || undefined,
```
dpinol: was this really necessary?

PR #138, opened 2019-07-26
Webchat style setup

imOscarCrespo:
webchat can receive custom components from bots and apply it

### DIFF
--- a/packages/botonic-react/src/components/message.jsx
+++ b/packages/botonic-react/src/components/message.jsx
@@ -4,6 +4,15 @@ import { isBrowser, isNode } from '@botonic/core'
 import { WebchatContext, RequestContext } from '../contexts'
 import { Button } from './button'
 import { Reply } from './reply'
+import Logo from '../webchat/botonic_react_logo100x100.png'
+import styled from 'styled-components'
+
+const DefaultMessage = styled.div`
+  position: relative;
+  margin: 8px;
+  font-family: Arial, Helvetica, sans-serif;
+  border-radius: 8px;
+`
 
 export const Message = props => {
   const { defaultTyping, defaultDelay } = useContext(RequestContext)
@@ -77,6 +86,22 @@ export const Message = props => {
     return isFromUser() ? webchatState.theme.brandColor : '#F1F0F0'
   }
 
+  const getFontColor = colors => {
+    if (isFromUser()) {
+      if (webchatState.theme.customUserMessages.color) {
+        return webchatState.theme.customUserMessages.color
+      } else {
+        return '#ffffff'
+      }
+    } else {
+      if (webchatState.theme.customBotMessages.color) {
+        return webchatState.theme.customBotMessages.color
+      } else {
+        return '#000'
+      }
+    }
+  }
+
   const renderBrowser = () => {
     let m = webchatState.messagesJSON.find(m => m.id === state.id)
     if (!m || !m.display) return <></>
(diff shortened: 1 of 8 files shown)


packages/botonic-react/src/webchat/menu.jsx
```diff
@@ -0,0 +1,42 @@
...
+import { Button } from '../components/button'
+import { WebchatMessageList } from './messageList'
+import Textarea from 'react-textarea-autosize'
+
+const DefaultMenu = () => {
+  const { webchatState } = useContext(WebchatContext)
+  return (
+    <Textarea
+      name="text"
+      minRows={2}
```
Khaeshah: Is this going to be always between 2 and 4 or in the future the user can define it?

packages/botonic-react/src/components/message.jsx
```diff
@@ -92,52 +117,75 @@ export const Message = props => {
...
           style={{
-            padding: '8px 12px',
-            display: 'flex',
-            flexDirection: 'column',
-            whiteSpace: 'pre-line'
+            left: isFromBot() ? 30 : 0,
+            top: isFromBot() ? 5 : 0,
+            backgroundColor: getBgColor(),
+            color:
+              webchatState.theme.customUserMessages ||
```
dpinol: since the color calculation is complex, maybe we can set it in a variable before the return at line 117?

packages/botonic-react/src/webchat/header.jsx
```diff
@@ -1,58 +1,116 @@
...
+const Subtitle = styled.h1`
+  @import url('https://fonts.googleapis.com/css?family=Noto+Sans+JP');
+  font-size: 11px;
+  line-height: 16px;
+  /* identical to box height */
+  margin-top: -12px;
+  color: #ffffff;
+`
+const Diffuse = styled(Flex)`
+  background: linear-gradient(90deg, #2e203b 0%, ${props => props.colorr} 100%);
```
dpinol: is there a typo in "colorr"?

packages/botonic-react/src/webchat/header.jsx
```diff
@@ -1,58 +1,116 @@
...
+  width: 1;
+  height: 55px;
+  border-radius: 6px 6px 0px 0px;
+`
+
+export const DefaultHeader = props => (
+  <Diffuse colorr={props.color}>
```
dpinol: same with colorr

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -116,9 +129,28 @@ export const Webchat = forwardRef((props, ref) => {
...
   }, [props.theme])
 
+  const openMenu = () => {
+    if (menuIsOpened) {
+      setMenuIsOpened(false)
+    } else {
+      setMenuIsOpened(true)
+    }
+  }
+  const openEmoji = () => {
```
dpinol: if the method also closes the emoji , maybe we could call it toggleEmoji instead of openEmoji?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -116,9 +129,28 @@ export const Webchat = forwardRef((props, ref) => {
     updateTheme({ ...webchatState.theme, ...props.theme })
   }, [props.theme])
 
+  const openMenu = () => {
```
dpinol: if the method also closes the menu , maybe we could call it toggleMenu instead of openMenu?

packages/botonic-react/src/webchat/header.jsx
```diff
@@ -1,58 +1,116 @@
...
-          toggleWebchat(false)
-          event.preventDefault()
-        }}
-      >
-        ✕
-      </div>
-    </div>
-  )
+    </Flex>
+    <Flex width={1} flexDirection="column">
```
dpinol: I think we use single quotes everywhere (see that we have ./packages/botonic-react/.prettierrc). I suggest:
* configure your IDE to highlight prettier warnings
* when you finish your current task, commit or stash all you pending code
* run "prettier --write" 
* create a commit with only the prettified changes

packages/botonic-react/src/components/message.jsx
```diff
@@ -77,6 +86,22 @@ export const Message = props => {
     return isFromUser() ? webchatState.theme.brandColor : '#F1F0F0'
   }
 
+  const getFontColor = colors => {
+    if (isFromUser()) {
+      if (webchatState.theme.customUserMessages.color) {
```
guinii: If `customUserMessages` is not defined, this line and the function will break!
Same in the return and the 'else' part

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -54,38 +62,45 @@ export const Webchat = forwardRef((props, ref) => {
     updateHandoff,
     updateTheme,
     updateDevSettings,
-    toggleWebchat
+    toggleWebchat,
+    openWebviewT,
+    closeWebviewT
   } = props.webchatHooks || useWebchat()
-
+  console.log(webchatState)
```
guinii: remove console.log

packages/botonic-react/src/webchat/webchat.jsx
imOscarCrespo: sure, but i realize it doesn't have to be here

packages/botonic-react/src/components/message.jsx
imOscarCrespo: done

packages/botonic-react/src/webchat/webchat.jsx
imOscarCrespo: done

PR #158, opened 2019-08-07
Change LUIS plugin intents.score and improve types

dpinol:
- changed LUIS plugin format of intents to rename score to confidence
- improve typings of core and react

### DIFF
--- a/packages/botonic-core/index.d.ts
+++ b/packages/botonic-core/index.d.ts
@@ -1,9 +1,18 @@
 export type Locales = { [id: string]: string | string[] | Locales }
 
 export interface Input {
-  type: string // text, postback...
+  // text, postback...
+  type: string
   payload?: string
   data?: string
+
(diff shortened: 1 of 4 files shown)


packages/botonic-core/index.d.ts
```diff
@@ -37,37 +46,61 @@ export interface Route {
...
 ): Promise<any>
 
-type Routes = Route[] | ((_: { input: Input; session: Session }) => Route[])
+export interface Plugin {
+  pre(_: { input: Input; session: Session; lastRoutePath: string })
+  post(_: {
+    input: Input
+    session: Session
+    lastRoutePath: string
+    response: string
```
vanbasten17: with what is associated this response param?

dpinol: good catch! see d37df80 for a fix and description of the semantics

PR #189, opened 2019-08-19
Webchat styles styled components

imOscarCrespo:
In this PR I did some changes about:
- Styled components: I've used the library styled-components for updating the way they were defined, there are still components to do.
- Persistent menu: in our webchat.jsx now we can recieve from a bot a persistent menu. It's a must  for Hp, but we can use it for other clients
- Credit cards and phone numbers: Hp asked us to block when user input was a phone number or credit card.


packages/botonic-react/src/components/carousel.jsx
```diff
@@ -1,8 +1,22 @@
 import React from 'react'
+import styled from 'styled-components'
 
 import { Message } from './message'
 import { isBrowser, isNode } from '@botonic/core'
 
+const Box = styled.div`
```
dpinol: cool, I didn't know this syntax

packages/botonic-react/src/components/carousel.jsx
```diff
@@ -16,33 +30,17 @@ export const Carousel = props => {
...
+      </Box>
     )
   }
   return (
     <Message
       style={{ maxWidth: '90%', padding: 0 }}
       blob={false}
       json={serialize(props)}
-      type='carousel'
+      type="carousel"
```
dpinol: did you get an error with single quotes? @Khaeshah shared some prettier config to use " on jsx

packages/botonic-react/src/components/message.jsx
```diff
@@ -77,6 +86,20 @@ export const Message = props => {
     return isFromUser() ? webchatState.theme.brandColor : '#F1F0F0'
   }
 
+  const getFontColor = props => {
+    let fontColorUser = '#ffffff'
```
dpinol: thanks for trying the refactor! 👍

packages/botonic-react/src/webchat/webchatDev.jsx
```diff
@@ -111,21 +116,21 @@ export const WebchatDev = forwardRef((props, ref) => {
             >
               <label style={{ flex: 'none', marginBottom: 6 }}>
                 <input
-                  type='checkbox'
-                  name='toggleSessionView'
+                  type="checkbox"
```
dpinol: same, jsxSingleQuote

imOscarCrespo:
> why did you remove them?

idk why it is not in this code, but i did a push with the changes

vanbasten17:
Another thing that could be interesting. We can destructure the props in the same line we declare the parameters props => {prop1, prop2, ...}. I think this is useful when we are on a specific class to know what we have available to work with.
```javascript
const Test = ({ prop1, prop2 }) => {
  return <h1>{prop1}</h1>
};

function App() {
  return (
    <div className="App">
      <Test prop1={"hola"} />
    </div>
  );
}
```

PR #197, opened 2019-08-22
add: commits from another branch that was done

imOscarCrespo:
I've created this branch becouse I had a problem with the "persistenMenu" branch. It's a copy of that. In this PR I've customized the persistent menu in case that it exists.


packages/botonic-react/src/components/button.jsx
```diff
@@ -37,6 +37,7 @@ export const Button = props => {
     else if (props.path) sendPayload(`__PATH_PAYLOAD__${props.path}`)
     else if (props.payload) sendPayload(props.payload)
```
dpinol: @ericmarcos if we allowed having both payload and url, would we be able to open an url and send a track when user presses a button?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -154,6 +156,10 @@ export const Webchat = forwardRef((props, ref) => {
     menuIsOpened ? setMenuIsOpened(false) : setMenuIsOpened(true)
   }
 
+  const closeMenu = () => {
+    setMenuIsOpened(false)
+  }
+
   const sendInput = async input => {
     let isRegex = false
     let inputMessage = null
```
dpinol: this method is too long. Maybe we can split it using  a private method for text inputs?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -216,17 +222,19 @@ export const Webchat = forwardRef((props, ref) => {
...
-      updateTyping(false)
-      if (Array.isArray(response)) response.map(r => addMessageComponent(r))
-      else if (response) addMessageComponent(response)
-      if (session) {
-        updateSession(session)
-        let action = session._botonic_action || ''
-        let handoff = action.startsWith('create_case')
-        if (handoff && isDev()) addMessageComponent(<Handoff />)
-        updateHandoff(handoff)
+      if (!isRegex) {
```
dpinol: it looks like isRegex is not defined here

packages/botonic-react/src/webchat/webchat.jsx
imOscarCrespo: updated with the last PR

PR #204, opened 2019-08-26
Persistent menu button url

imOscarCrespo:
One of the requirements of HP was that options of persistent menu open urls directly

### DIFF
--- a/packages/botonic-react/src/components/button.jsx
+++ b/packages/botonic-react/src/components/button.jsx
@@ -64,7 +64,6 @@ export const Button = props => {
   }
 
   const renderNode = () => {
-    console.log('hi')
     if (props.webview) {
       let Webview = props.webview
       let params = ''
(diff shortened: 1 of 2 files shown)


packages/botonic-react/src/components/button.jsx
```diff
@@ -36,8 +36,10 @@ export const Button = props => {
     if (props.webview) openWebview(props.webview, props.params)
     else if (props.path) sendPayload(`__PATH_PAYLOAD__${props.path}`)
     else if (props.payload) sendPayload(props.payload)
-    else if (props.url) window.open(props.url)
-    else if (props.onClick) props.onClick()
+    else if (props.url) {
+      window.open(props.url)
+      props.onClick()
```
dpinol: is props.onClick only defined when props.url is defined?

imOscarCrespo: no, props onclick is the method you want to call from parent

PR #208, opened 2019-08-27
fix: simplify code only deleting 'else'

imOscarCrespo:
I was using more code that I needed, the logic of this is for buttons of the persistent menu that once clicked the menu has to be closed

### DIFF
--- a/packages/botonic-react/src/components/button.jsx
+++ b/packages/botonic-react/src/components/button.jsx
@@ -38,8 +38,8 @@ export const Button = props => {
     else if (props.payload) sendPayload(props.payload)
     else if (props.url) {
       window.open(props.url)
-      props.onClick()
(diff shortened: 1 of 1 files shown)


packages/botonic-react/src/components/button.jsx
```diff
@@ -38,8 +38,8 @@ export const Button = props => {
     else if (props.payload) sendPayload(props.payload)
     else if (props.url) {
       window.open(props.url)
-      props.onClick()
-    } else if (props.onClick) props.onClick()
+    }
+    if (props.onClick) props.onClick()
```
dpinol: another difference is that now props.onClick() is also called when there's path or payload

imOscarCrespo: yes because despite of being a path or payload we can want to call a method

PR #216, opened 2019-08-30
add(react): close webchat button for default headers

imOscarCrespo:
- header.jsx: add a styled component button for closing the webchat when is a default header

- webhcat.jsx: Here we send to child component(header.jsx) the method for closing the webchat.

- webchatDev.jsx: When we want to serve our webchat instead of passing a custom header we want to see the dropdown menu with the options "Show session" & "Keep session on reload"

### DIFF
--- a/packages/botonic-react/src/webchat/header.jsx
+++ b/packages/botonic-react/src/webchat/header.jsx
@@ -27,7 +27,7 @@ const Diffuse = styled(Flex)`
   height: 55px;
   border-radius: 6px 6px 0px 0px;
 `
-const CrossHeader = styled(Flex)`
+const CloseHeader = styled(Flex)`
   position: absolute;
   right: 10px;
   top: 14px;
@@ -63,14 +63,14 @@ export const DefaultHeader = props => {
         <HeaderTitle>{headerTitle}</HeaderTitle>
         <Subtitle>{headerSubtitle}</Subtitle>
       </Flex>
-      <CrossHeader onClick={props.onChange}>X</CrossHeader>}
+      <CloseHeader onClick={props.onChange}>X</CloseHeader>}
     </Diffuse>
   )
 }
 
 export const WebchatHeader = () => {
   const { webchatState } = useContext(WebchatContext)
-  const handleCloseWebchat = () => {
+  const handleCloseWebchat = event => {
     props.onChange(event.target.value)
   }
   if (webchatState.theme.customHeader) {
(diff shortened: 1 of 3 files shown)


packages/botonic-react/src/webchat/header.jsx
```diff
@@ -42,16 +54,24 @@ export const DefaultHeader = props => (
...
     </Flex>
+    <CrossHeader onClick={props.onChange}>X</CrossHeader>}
   </Diffuse>
 )
 
 export const WebchatHeader = props => {
   const { webchatState } = useContext(WebchatContext)
-
+  const handleCloseWebchat = event => {
+    props.onChange(event.target.value)
```
guinii: I think it's not needed to pass argument in this function.

packages/botonic-react/src/webchat/header.jsx
```diff
@@ -42,16 +54,24 @@ export const DefaultHeader = props => (
       <HeaderTitle>Botonic</HeaderTitle>
       <Subtitle>Online</Subtitle>
```
guinii: By default, I think that we should not define a subtitle, because 'Online' can be seen as if you are in office hours it's online, if not they might expect to change it.

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -378,6 +378,10 @@ export const Webchat = forwardRef((props, ref) => {
               height: 36,
               flex: 'none'
             }}
+            value={true}
```
guinii: What's the use of this value???

packages/botonic-react/src/webchat/header.jsx
imOscarCrespo: you are right!

dpinol: > I think it's not needed to pass argument in this function.

@guinii https://developer.mozilla.org/en-US/docs/Web/API/Window/event it's deprecated to use the global event.

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -378,6 +378,9 @@ export const Webchat = forwardRef((props, ref) => {
               height: 36,
               flex: 'none'
             }}
+            onChange={() => {
```
ericmarcos: "onChange" is not the best name here, better "onCloseClick"

packages/botonic-react/src/webchat/header.jsx
```diff
@@ -40,18 +52,25 @@ export const DefaultHeader = props => (
     </Flex>
     <Flex width={1} flexDirection='column'>
       <HeaderTitle>Botonic</HeaderTitle>
-      <Subtitle>Online</Subtitle>
     </Flex>
+    <CrossHeader onClick={props.onChange}>X</CrossHeader>}
```
ericmarcos: If the component is called "CrossHeader" I would expect the X character to be included in the component

PR #217, opened 2019-08-30
refactor(react/style): change size of bot icon and adapt carousel

imOscarCrespo:
In this branch I've adjusted the size of the bot icon because it was too big. I also have deleted the pointerStyles from bot because the logo have it.
Because the reduction of the icon I've had to move a little bit the carousel to the left.

### DIFF
--- a/packages/botonic-react/src/components/message.jsx
+++ b/packages/botonic-react/src/components/message.jsx
@@ -125,13 +125,13 @@ export const Message = props => {
           {isFromBot() && webchatState.theme.botLogoChat ? (
             <webchatState.theme.botLogoChat />
           ) : (
-            isFromBot() && <img style={{ width: 30 }} src={Logo} />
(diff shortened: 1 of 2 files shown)


packages/botonic-react/src/components/message.jsx
```diff
@@ -175,7 +175,7 @@ export const Message = props => {
           {isFromBot() && blob && (
             <div
               style={{
-                ...pointerStyles,
+                //...pointerStyles,
```
dpinol: should we remove the comment?
what is "pointer" here? can we find a better name?

imOscarCrespo: yes we can remove it.

imOscarCrespo: for the name, it doesn't come to my mind no others name for this. It was defined before this PR

PR #218, opened 2019-08-30
add(react/header): new customitzation features on default header

imOscarCrespo:
If we don't want to create a customize header and we only want to change the image header, title or subtitle now we can by checking if user passes from index.html

### DIFF
--- a/packages/botonic-react/src/webchat/header.jsx
+++ b/packages/botonic-react/src/webchat/header.jsx
@@ -9,7 +9,6 @@ const HeaderTitle = styled.h1`
   font-size: 15px;
   line-height: 22px;
   color: #ffffff;
-  margin-top: ${props => (props.headerSubtitle == undefined ? '18px' : 5)};
 `
 
 const Subtitle = styled.h1`
@@ -48,8 +47,8 @@ export const DefaultHeader = props => {
           />
         )}
       </Flex>
-      <Flex width={1} flexDirection='column'>
-        <HeaderTitle headerSubtitle={headerSubtitle}>{headerTitle}</HeaderTitle>
+      <Flex width={1} flexDirection='column' justifyContent='center'>
+        <HeaderTitle>{headerTitle}</HeaderTitle>
         <Subtitle>{headerSubtitle}</Subtitle>
       </Flex>
     </Diffuse>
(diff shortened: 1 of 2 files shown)


packages/botonic-react/src/webchat/header.jsx
```diff
@@ -26,32 +27,47 @@ const Diffuse = styled(Flex)`
...
-    <Flex width={1} flexDirection='column'>
-      <HeaderTitle>Botonic</HeaderTitle>
-      <Subtitle>Online</Subtitle>
-    </Flex>
-  </Diffuse>
-)
+export const DefaultHeader = props => {
+  let HeaderImage = props.webchatState.theme.headerImage
+  let headerTitle = props.webchatState.theme.headerTitle
+  headerTitle ? headerTitle : (headerTitle = 'Botonic')
```
dpinol: is this the same as 
`headerTitle = headerTitle or 'Botonic'` ?

packages/botonic-react/src/webchat/header.jsx
```diff
@@ -26,32 +27,47 @@ const Diffuse = styled(Flex)`
...
+              paddingTop: 15,
+              paddingLeft: 15
+            }}
+            src={Logo}
+          />
+        )}
+      </Flex>
+      <Flex width={1} flexDirection='column'>
+        <HeaderTitle headerSubtitle={headerSubtitle}>{headerTitle}</HeaderTitle>
+        <Subtitle>{headerSubtitle}</Subtitle>
```
dpinol: is it ok that headerSubtitle is passed to both the HeaderTitle and the Subtitle?

imOscarCrespo: headerSubtitle is passed because we need to know if there are or not for the position of the title. 
margin-top: ${props => (props.headerSubtitle == undefined ? '18px' : 5)};

packages/botonic-react/src/webchat/header.jsx
imOscarCrespo: yes!! thank you!!

packages/botonic-react/src/webchat/header.jsx
ericmarcos: This is not the best way to do this, fixed margins are hard to maintain. Usually you want to center vertically the content of the header, so it doesn't matter if you have 1, 2 or more elements, they'll always be displayed well. Then you add a "min-width" to the container element so that its height does not depend on the content.

PR #219, opened 2019-09-02
refactor(webchat/persistentMenu): condition for showing persistent menu

imOscarCrespo:


imOscarCrespo:
> In line 402, the part '{Object.values(props.persistentMenu).map()...' is it working when persistentMenu is empty?



> In line 402, the part '{Object.values(props.persistentMenu).map()...' is it working when persistentMenu is empty?
yes because the variable menuIsOpened

guinii:
What I mean is that if props.persistentMenu is not defined, because persistentMenu is not included, then the `Object.values(props.persistentMenu).map()` I don't know it will break or have a bad behaviour.

PR #224, opened 2019-09-03
refactor(react): Compatible with ie11 + fix webchat behaviour

mireiaespuga:
In this PR, we can find the following changes:

- Customize the width of the messages

- Close quickreplies when opening webviews

- Add polyfills for working with ie11


packages/botonic-react/src/webchat/replies.jsx
```diff
@@ -10,7 +10,7 @@ export const WebchatReplies = props => {
     <div
       style={{
         ...(props.style || {}),
-        overflowX: 'auto',
+        // overflowX: 'auto',
```
dpinol: should we remove the line?

ericmarcos:
We're creating a new config argument `webchatConfig` just for one parameter `messageStyle` when we already have `theme` for styling customization. I'd use `theme.messageStyle` so it removes a lot of boilerplate code and keeps styling config inside `theme`.

PR #247, opened 2019-09-13
Feature/windows cross platform

vanbasten17:
Hi guys!

Taking profit that I've been giving support to Mia, who works with Windows, I have spent some time rewriting parts of botonic-cli (basically new and deploy commands) in order to have cross-platform support. I have tested it in Windows and in MacOS and works as expected!
You will see that when creating the bundle (deploy.ts) (it has been a headache to make it work on Windows), I've used a 'tmp' directory in order to preserve the 'dist' folder inside the 'botonic_bundle.zip'. This was necessary due that the library doesn't include the 'dist' parent folder when compressing the content, provoking our backend to fail on posting the zip (we are decompressing botonic_bundle.zip, obtaining a 'dist' dir).


packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -257,7 +259,10 @@ Uploading...
   }
 
   async deploy() {
```
dpinol: this method is too long. I suggest splitting in 
deploy() {
try {
 buildIfChanged... (no need to wrap in new method)
  new zip() method
 new deployBundle() method
new method displayProviders()
} finally () {
remove tmp and bundle
}

packages/botonic-cli/src/commands/new.ts
```diff
@@ -113,7 +113,7 @@ Creating...
     spinner.succeed()
     await this.botonicApiService.buildIfChanged(false)
     this.botonicApiService.beforeExit()
-    await exec('mv ../.botonic.json .')
+    moveSync(join('..', '.botonic.json'), join(process.cwd(), '.botonic.json'))
```
dpinol: can't we use '.' instead of '.botonic.json' for the target argument?

vanbasten17: i remember trying this when developing the code and wasn't working as expected.

packages/botonic-cli/src/commands/deploy.ts
vanbasten17: changed to:
deploy(){
try{
buildIfChanged <- same logic we previously had here
createBundle() <- this is the responsible of generating the bundle
deployBundle() <- all logic of posting the data goes here
displayDeployResults() <- we already had a displayProviders method, so I have encapsulated it into displayDeployResults method
} finally{
removeSync both tmp and bundle
}
}

vanbasten17: thanks for your feeddback Dani! 👍

packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -350,10 +340,27 @@ Uploading...
       track('Deploy Botonic Provider Error', { error: e })
       console.log(colors.red(`There was an error getting the providers: ${e}`))
     }
+  }
+
+  async deploy() {
```
dpinol: 🆒 isn't it a pleasure to read this kind of code? 😺

vanbasten17: for sure!!

packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -345,9 +340,27 @@ Uploading...
...
+      let build_out = await this.botonicApiService.buildIfChanged(
+        force,
+        npmCommand
+      )
+      if (!build_out) {
+        track('Deploy Botonic Build Error')
+        console.log(colors.red('There was a problem building the bot'))
+        return
+      }
+      await this.createBundle()
```
guinii: It's okai, but I think we should start using this `await Promise.all([someCall(), anotherCall()])` when we have more than one async function.

vanbasten17: mmm.. but in this way some methods can be executed not in the right order, right?
I mean, by calling Promise.all, we allow deployBundle to be run before createBundle, for example, or am I wrong?

PR #252, opened 2019-09-17
React/styled components

imOscarCrespo:


packages/botonic-react/src/components/message.jsx
```diff
@@ -13,6 +13,12 @@ const DefaultMessage = styled.div`
   font-family: Arial, Helvetica, sans-serif;
   border-radius: 8px;
 `
+const TextMessage = styled.div`
```
guinii: Maybe I would change the name, it's not mandatory that this is gonna have a Text in it.

packages/botonic-react/src/webchat/replies.jsx
```diff
@@ -1,28 +1,35 @@
...
 const alignOptions = /^left$|^center$|^right$/
 const wrapOptions = /^wrap$|^no-wrap$/
 
 export const WebchatReplies = props => {
   const { webchatState } = useContext(WebchatContext)
   return (
-    <div
+    <StyledReplies
       style={{
         ...(props.style || {}),
```
guinii: We need to put always the default style options (textAlign and whiteSpace in this case) and then the `...(props.style || {})` in all the components, if not, the props will be overwrite and it can't be customized this values.
Check in all the components that this behaviour is not happening.

PR #253, opened 2019-09-17
feat(react): add customMessage wrapper component and deserializer

ericmarcos:
This feature allows Botonic developers to define custom message types so they can extend our predefined collection of message types (beyond Text, Carousel, Image, etc).

In order to create a custom message type, you should create a new component like:
```
import React from 'react'
import { WebchatContext } from '@botonic/react'
import { customMessage } from '@botonic/react/src/components/customMessage'

class TextCopy extends React.Component {
  static contextType = WebchatContext
  render() {
    return (
      <>
        <button onClick={() =>  console.log('copy to clipboard!')}>copy!</button>
        <button onClick={() => this.context.sendText('Hi')}>send Hi!</button>
        {this.props.children}
      </>
    )
  }
}

  name: 'copyToClipboard',
  component: TextCopy,
  serializer: props => ({ text: String(props.children) }),
  deserializer: data => <TextCopy>{data.text}</TextCopy>
})
```

The idea is that the user defines a normal, simple React component like `TextCopy`, then exports a wrapped version of this component using the `customMessage` utility. The developer should specify a name for the message type and how it should be serialized/deserialized.
Additionally, the developer can use the `WebchatContext` in order to trigger bot actions from that component.

Then the developer can use it inside an action, the same way he'd use any other message:
```
import React from 'react'
import { Text, Reply } from '@botonic/react'
import TextCopy from '../components/copyToClipboard'

  render() {
    return (
      <>
        <TextCopy>Copy me!</TextCopy>
        <Text>This is a normal text/Text>
      </>
    )
  }
}
```


ericmarcos:
> Maybe get the CustomMessageTypes from BotonicProject like:
> `import { webchat } from 'BotonicProject'`
> 
> `webchat.theme.customMessageTypes`

I've been thinking about it and I prefer to keep passing arguments. The `BotonicProject` module is a handy black magic trick that Webpack enables us. However, I don't want to abuse it, it will make things harder to maintain in the future. I'd like to use it only when there's no other option available, that is in the entry.js file. Remember: with great power comes great responsibility.
Also, reading the theme from `BotonicProject` would make it impossible to pass custom arguments using the webchat runtimeOptions.

ericmarcos:
I simplified the `customMessage` wrapper so that there's no need to pass serializer/deserializer. Also exported `customMessage` from @botonic/react index.
So now a developer would define a custom message type like this:

```
import React from 'react'
import { WebchatContext } from '@botonic/react'
import { customMessage } from '@botonic/react/src/components/customMessage'

class TextCopy extends React.Component {
  static contextType = WebchatContext
  render() {
    return (
      <>
        <button onClick={() =>  console.log('copy to clipboard!')}>copy!</button>
        <button onClick={() => this.context.sendText('Hi')}>send Hi!</button>
        {this.props.children}
      </>
    )
  }
}

  name: 'copyToClipboard',
  component: TextCopy
})
```

PR #257, opened 2019-09-19
Chore/nlu updates

vanbasten17:
Hi team!!
I have just made some fixes related to NLU packages:
* Fixed one bug when retrieving the entities (an entities key was missing)
* Updating both `nlu` and `plugin-nlu` deps to v.1.2.9 for `tfjs` and `tfjs-node`
* Updated nlu template from cli templates
* Updated documentation for botonic-plugin-nlu
* Versioning and published packages botonic/nlu and botonic/plugin-nlu with new deps and the fix for the bug


docs/5.plugins/3.plugin-nlu.mdx
```diff
@@ -31,28 +31,46 @@ The first step is to install the module with the following command in your bot's
...
 
-**src/nlu/intents/eng**
+**src/nlu/utterances/eng**
 * BookRestaurant.txt
 * GetDirections.txt
 * Greetings.txt
 
-So in this example, we have added three different intents which our bot will know how to understand. Obviously, 
-but the more example utterances you have, the better results you will obtain.
+So in this example, we have added three different intents which our bot will know how to understand. Obviously, the more example utterances you have, the better results you will obtain.
```
dpinol: but it's the plugin which reads the files and not the bot, right?
In any case, why not just saying "in this example, we have added three different intents which our bot will know how to understand." You already described the format of the files.

docs/5.plugins/3.plugin-nlu.mdx
```diff
@@ -31,28 +31,46 @@ The first step is to install the module with the following command in your bot's
...
+
+### **Defining your entities**
+You can also define your entities in the same utterances as follows:
```
dpinol: "in the same utterances" -> "within the same utterances" ?

docs/5.plugins/3.plugin-nlu.mdx
```diff
@@ -31,28 +31,46 @@ The first step is to install the module with the following command in your bot's
...
+
+"Book a table for today's lunch at `[Eggy's Diner](Place)` for 3 people"
+
+With this notation we are telling Botonic NLU that `Eggy's Diner` is the value for an entity of type `Place`.
+`Place` is a generic entity as well as `Person`, `Organization` or `Date`, this is how you can access them with the input object:
```
dpinol: If this is a closed list, maybe we should list all types somewhere. Otherwise, I would document how to declare new types

docs/5.plugins/3.plugin-nlu.mdx
```diff
@@ -31,28 +31,46 @@ The first step is to install the module with the following command in your bot's
...
+* `Place`: `input.entities.places`
+* `Person`: `input.entities.people`
+* `Organization`: `input.entities.organizations`
+* `Date`: `input.entities.dates`
+
+Sometimes, you will need to capture more specific entites which you can define as this:
```
dpinol: entites => entities

"which you can define as this" sounds like this is a different syntax than before, but it looks the same to me.

by "specific entities" do you mean custom ones?

docs/5.plugins/3.plugin-nlu.mdx
dpinol: instead of "generic entity" should we say "entity type"?

docs/5.plugins/3.plugin-nlu.mdx
```diff
@@ -31,28 +31,46 @@ The first step is to install the module with the following command in your bot's
...
+* `Place`: `input.entities.places`
+* `Person`: `input.entities.people`
+* `Organization`: `input.entities.organizations`
+* `Date`: `input.entities.dates`
+
+Sometimes, you will need to capture more specific entites which you can define as this:  
+
+"Show me the way to Rand's birthday party by `[car](Vehicle)`"  
+
+which you can access in `input.entities.tags`, that is a list containing all the tagged words by the algorithm.
```
dpinol: not sure what a tag is. A custom entity tag? You can make explicit where it will contain car or vehice.

docs/5.plugins/3.plugin-nlu.mdx
vanbasten17: --> In this example, we have added three different intents which our bot will know how to understand.

docs/5.plugins/3.plugin-nlu.mdx
vanbasten17: corrected

docs/5.plugins/3.plugin-nlu.mdx
vanbasten17: --> The entity recognition system is based on [spencermountain/compromise](https://github.com/spencermountain/compromise), so this is the [full list](https://observablehq.com/@spencermountain/compromise-tags) of 
entity tags which are supported by default.

docs/5.plugins/3.plugin-nlu.mdx
vanbasten17: fixed, changed specific by custom as well, as this is what they are

docs/5.plugins/3.plugin-nlu.mdx
vanbasten17: replaced by `is a named entity type` as this is the technical definition.
ref: `https://en.wikipedia.org/wiki/Named-entity_recognition`

docs/5.plugins/3.plugin-nlu.mdx
vanbasten17: --> which you can access in `input.entities.tags`. This list will contain all the custom tagged entity types with its values.

vanbasten17:
forever grateful for your feedback Dani 🤘

docs/5.plugins/3.plugin-nlu.mdx
```diff
@@ -25,10 +25,10 @@ Alternatively, if you want to take a closer look of how to make it work from scr
 
 ### **Installing the plugin**
 
-The first step is to install the module with the following command in your bot's project:
+The first step is to install the following modules with the following command in your bot's project:
```
dpinol: not an expert in JS, but I think that a module is a file, and npm installs packages not modules

vanbasten17: sure,fixed, and yes you are an expert! 😜

PR #260, opened 2019-09-23
feat(react/webchat-emojipicker): emojiPicker added in user input box

imOscarCrespo:
I've add a new feature to have an emojipicker in our footer box. To call this menu it's necessary to declare in the index.js of our bot : " emojiPicker: true". I also refactored some styles to have a better responsive properties.


package.json
```diff
@@ -19,5 +19,8 @@
   },
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "dependencies": {
```
dpinol: dependency should be added instead to react package

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -154,6 +155,9 @@ export const Webchat = forwardRef((props, ref) => {
   const handleMenu = () => {
     menuIsOpened ? setMenuIsOpened(false) : setMenuIsOpened(true)
   }
+  const handleEmoji = () => {
+    emojiIsOpened ? setemojiIsOpened(false) : setemojiIsOpened(true)
```
dpinol: setemojiIsOpened => setEmojiIsOpened

where is setemojiIsOpened defined?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -460,15 +462,32 @@ export const Webchat = forwardRef((props, ref) => {
...
                   border: 'none',
                   resize: 'none',
                   overflow: 'auto',
                   outline: 'none',
-                  marginTop: '13px'
+                  flex: '1 1 auto',
+                  padding: '14px 10px 0px 0px'
                 }}
               />
+              {props.emojiPicker && (
```
dpinol: this method is too large.
Please create in this PR a method like renderEmojoPicker with this code, and in another PR try to split this huge return in smaller chunks. 
thanks

packages/botonic-react/src/webchat/webchat.jsx
imOscarCrespo: here : const [emojiIsOpened, setemojiIsOpened] = useState(false)
It's a react hook state.

packages/botonic-react/src/webchat/webchat.jsx
imOscarCrespo: what method? this is not a method. It's only a condition for seeing if user has declared emojiPicker in his index.js file

packages/botonic-react/src/webchat/webchat.jsx
dpinol: thanks, please rename it to setEmojiIsOpened

packages/botonic-react/src/webchat/webchat.jsx
dpinol: I meant that the function it belongs to is too long. Please Move lines 475 to 490 to a different method in the is PR. In another PR, please do the same to simplify further the Webchat return (from line 336 till the end of the file).
Ideally it should look st like:
``` return (
    <WebchatContext.Provider
      value={{
        sendText,
        sendPayload,
        openWebview,
        resolveCase,
        webchatState,
        addMessage,
        toggleWebchat,
        updateMessage,
        updateReplies,
        staticAssetsUrl
      }}
    >
   {renderTriggerButton(webChatState)}
   {webchatState.isWebchatOpen && (
        <div
          style=divStyle(webchatState)
        >
         header(toggleWebchat)
         messageList(webchatState)
         replies(webchatState)
         emojiIsOpened && emojiPicker()
         persistenMenu(...)
        ....
```

PR #274, opened 2019-09-26
refactor: webchat return cleaned declaring components outside

imOscarCrespo:


dpinol:
this PR is not finished right?
https://github.com/hubtype/botonic/pull/260#discussion_r327083340 suggested to simplify even more the return methos

PR #279, opened 2019-10-01
Fix/migrate dialogflow to v2

vanbasten17:
Hi all!

I've just migrated the plugin for Dialogflow to V2. Now it's working with the new api version of Dialogflow.
The endpoint that we used was `/query`. From now on the endpoint is `/detectIntent`[ref](https://cloud.google.com/dialogflow/docs/reference/rest/v2beta1/projects.agent.sessions/detectIntent). In order to avoid using a Node.JS SDK for doing the authorization of the calls, I've used [JWT](https://developers.google.com/identity/protocols/OAuth2ServiceAccount#jwt-auth).
If you want to try the plugin, you should go to [Dialogflow console](https://dialogflow.cloud.google.com/), and click on the settings wheel for an agent (top-left). Then you need to click on your `Service Account` under `Google Projects` label. Once clicked, you should see the option `Create Key` which generates a .json. You need to copy this json inside the bot's project and use it in `plugins.js` as this:
```javascript
import dialogflowCredentials from './newagent-1cdbd-f226af5569ac.json'
export const plugins = [
  {
    id: 'dialogflow',
    resolve: require('@botonic/plugin-dialogflow'),
    options: dialogflowCredentials
  }
]
```
After that, it should work! I've tested that the plugin is working by declaring the intent inside routes, and also using the plugin inside an action.

### DIFF
--- a/packages/botonic-plugin-dialogflow/src/index.js
+++ b/packages/botonic-plugin-dialogflow/src/index.js
@@ -7,10 +7,14 @@ export default class BotonicPluginDialogflow {
     this.projectID = creds.project_id
     this.sessionID = uuid.v4()
     this.creds = creds
-    return (async () => {
-      this.token = await this.generateToken(creds)
-      return this
-    })()
+    this.token = null
+  }
+
+  async getToken() {
+    if (!this.token) {
+      this.token = await this.generateToken(this.creds)
+    }
+    return this.token
   }
 
   async pre({ input, session, lastRoutePath }) {
@@ -61,7 +65,7 @@ export default class BotonicPluginDialogflow {
       method: 'post',
       url: `https://dialogflow.googleapis.com/v2/projects/${this.projectID}/agent/sessions/${this.sessionID}:detectIntent`,
       headers: {
-        Authorization: `Bearer ${this.token}`,
+        Authorization: `Bearer ${await this.getToken()}`,
         'Content-Type': 'application/json'
       },
       data: {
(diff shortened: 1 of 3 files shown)


dpinol:
> Hi all!
> 
> I've just migrated the plugin for Dialogflow to V2. Now it's working with the new api version of Dialogflow.
> The endpoint that we used was `/query`. From now on the endpoint is `/detectIntent`[ref](https://cloud.google.com/dialogflow/docs/reference/rest/v2beta1/projects.agent.sessions/detectIntent). In order to avoid using a Node.JS SDK for doing the authorization of the calls, I've used [JWT](https://developers.google.com/identity/protocols/OAuth2ServiceAccount#jwt-auth).
> If you want to try the plugin, you should go to [Dialogflow console](https://dialogflow.cloud.google.com/), and click on the settings wheel for an agent (top-left). Then you need to click on your `Service Account` under `Google Projects` label. Once clicked, you should see the option `Create Key` which generates a .json. You need to copy this json inside the bot's project and use it in `plugins.js` as this:
> 
> ```js
> import dialogflowCredentials from './newagent-1cdbd-f226af5569ac.json'
> export const plugins = [
>   {
>     id: 'dialogflow',
>     resolve: require('@botonic/plugin-dialogflow'),
>     options: dialogflowCredentials
>   }
> ]
> ```
> 

mayebe we should add this to our doc?

packages/botonic-plugin-dialogflow/src/index.js
```diff
@@ -1,66 +1,102 @@
...
+        }
+      }
+    })
+  }
+
   async dialogflowToInput({ input, session, lastRoutePath }) {
-    let df_session_id = session.user.id
-    let intent = null
     let confidence = 0
+    let intent = null
```
dpinol: if these default values are never used, I would just assign them to their final values. eg.
`let entities = queryResult.parameters`

PR #281, opened 2019-10-03
Release/0.9.0

ericmarcos:
Release branch for version 0.9.0 is ready. This branch includes several "last minute" fixes, refactors and style improvements, including:
* Webchat SDK:
  * Custom Message Types (#253)
  * Pusher error handling
  * Get unread messages (sent while offline) (#275)
  * EmojiPicker (#260)
  * Theme attributes names consolidation
  * Many little styles fixes

* NLU:
  * Refactor config file so that user don't need to define hyperparameters for every language, rename some attributes and refactor some internal functions.

PR #282, opened 2019-10-03
Release/0.9.0 revert styled components

ericmarcos:
Picking up from #281 but removing styled-components commits.
We should investigate further what's going on, but some styles defined with styled-components don't show up in production build (message, buttons) but some others do (header). As it's not critical, I've switched to inline styles to make sure it works so we can move forward with the release.


packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -371,115 +528,30 @@ export const Webchat = forwardRef((props, ref) => {
...
-                style={{
-                  ...props.theme.webviewStyle
-                }}
-                webview={webchatState.webview}
-              />
-            </RequestContext.Provider>
+          ) : (
+            <>
+              {webchatMessageList()}
+              {webchatState.replies &&
```
guinii: The replies, is going to be rendered fixed at the end of the message list. It should be as a new message that goes away when scrolling

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -320,13 +375,115 @@ export const Webchat = forwardRef((props, ref) => {
         ...webchatState.theme.triggerButtonStyle
       }}
     >
-      <img
+      {triggerImage && (
+        <img
+          style={{
+            height: 50
```
guinii: Maybe put some style option in order to define the size of the triggerImage. I don't know if it's better this option or force the user to create a CustomTriggerButton

packages/botonic-react/src/components/persistentMenu.jsx
```diff
@@ -4,6 +4,7 @@ export const PersistentMenu = props => {
     <div
       style={{
         position: 'absolute',
+        zIndex: 1,
```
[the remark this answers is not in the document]
ericmarcos: z-index is relative to the container. In this case we want the persistent menu to be in front of the other elements of our webchat.
If someone has visibility problems with the webchat on their website, the only solution is that they setup the zindex of the #root div manually.

packages/botonic-react/src/webchat/header.jsx
```diff
@@ -1,69 +1,76 @@
 import React, { useContext } from 'react'
 import { WebchatContext } from '../contexts'
+import { staticAsset } from '../utils'
 import styled from 'styled-components'
```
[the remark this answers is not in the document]
ericmarcos: yes.. they work ok in header elements. Not sure why.
I'll leave it like this as it's working

packages/botonic-react/src/webchat/webchat.jsx
ericmarcos: I'll leave it like this for now. I'd like to add a Messenger-like behaviour: replies are fixed at the end but they disappear if you scroll up more than X%

PR #297, opened 2019-10-08
Feat/new handoff options

vanbasten17:
Hi all! This PR is related with `https://github.com/metis-ai/hubtype-backend/pull/76`.
In this PR I have included new features to @botonic/core, which are:
* The method humanHandoff is redefined with new parameters in order to support passing an `agentEmail`, or `extraInfo: {caseInfo: "", note: ""}`.
* The new method getAvailableAgentsForQueue is added in order to have a way for dealing with the bot response if an agent is not available, for example.
I have also renamed some vars we have with snake_case to camelCase.

Usage:

```javascript
let agents = await getAvailableAgentsForQueue(session, 'CustomerSupport')

await humanHandOff({
      session,
      queueName: 'CustomerSupport',
      onFinish: { path: 'thanks-for-contacting' },
      agentEmail: 'agent-1@hubtype.com',
      extraInfo: {
        caseInfo: 'This is some case ::::info...',
        note: 'This is a NOTE:::_;$'
      }
    })
```


packages/botonic-core/index.d.ts
```diff
@@ -56,8 +56,10 @@ type Routes = Route[] | ((_: { input: Input; session: Session }) => Route[])
 // Desk
 export declare function humanHandOff(
   session: Session,
-  queue_name: string,
-  on_finish: { payload?: any; path?: any }
+  queueName?: string = '',
+  onFinish: { payload?: any; path?: any },
+  agentEmail?: string = '',
+  extraInfo: { caseInfo?: any; note?: any }
```
guinii: `any` type is always going to be a `string` right?

packages/botonic-core/src/handoff.js
```diff
@@ -1,40 +1,66 @@
...
   })
   return resp.data
 }
 
-export async function humanHandOff(session, queue_name, on_finish) {
-  let params = `create_case:${queue_name}`
-  if (on_finish) {
-    if (on_finish.path) params += `:__PATH_PAYLOAD__${on_finish.path}`
-    else if (on_finish.payload) params += `:${on_finish.payload}`
+export async function humanHandOff({
```
guinii: This function in the file `botonic-core/index.d.ts` is declared to return a `Promise<void>`, maybe we should change that declaration? Here we are not returning any `Promise`, right?

dpinol: @guinii async functions always return a promise

vanbasten17:
**UPDATE, NOW CAN BE MERGED**
* I have renamed the method `getAvailableAgentsForQueue` to `getAvailableAgents()`
* I have encoded the data with special characters internally for case_info and notes in order to avoid using tricks like `__NOTE__this_was_the_user_info__END_NOTE__`.
* I have ensured that is backward-compatible with previous @botonic/react versions by changing the arguments passed through an object to positional arguments. I have tested it and works as expected.
New usage:

```javascript
let agents = await getAvailableAgents(session,'QUEUE_ID')

await humanHandOff(
      session,
      'QUEUE_ID',
      { path: 'thanks-for-contacting' },
      'agent-1@hubtype.com',
      {
        caseInfo: 'This is some case ::::info...',
        note: 'This is a NOTE:::_;$'
      }
    )
```

PR #303, opened 2019-10-09
feat: add styledComponents to master

imOscarCrespo:
the other styled component branch is outdated so i 've created this branch from master and I 've done all the work here.

**UPDATED** (@vanbasten17 ): I achieved to make it work as expected in production with version `4.4.1`. I tried newests versions like 5.0.1 but it seems that are pretty new that doesn't work very well yet.
Taking profit of it I have:
* Fixed an issue which made Noto Sans JP typography not showing up.
* Tried to put in `constants.js` as many variables as possible which were used in lots of parts of the code.
* Converted the remainding inline-style components to styled components.
* Moved components which are only related to webchat into `webchat/components` and the ones used in Botonic `Audio, Button, Carousel, ...` left in the major level `components`.


imOscarCrespo:
We are waiting for a new release of styled-component library to fix the problems we have in production with some components. Meanwhile there are others components that work correctly

vanbasten17:
is this still alive? maybe we should rebase it with master to have it up-to-date

PR #316, opened 2019-10-10
chore(nlu/plugin-nlu): upgrade deps to tfjs 2.0

vanbasten17:
Just updated tensorflow js to the latest version (2.0 or 1.2.11). I've tested it in my machine without problems.

### DIFF
--- a/packages/botonic-nlu/package.json
+++ b/packages/botonic-nlu/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/nlu",
-  "version": "0.1.0-alpha.2",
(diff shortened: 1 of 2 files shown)


dpinol:
from 1.2.9 to 1.2.11 they upgraded to tf2?
no change in the api?

vanbasten17:
no Dani, whereas they have updated the package itself, the API methods remain the same. In fact, the version of its API methods is on its [1.2.7](https://js.tensorflow.org/api/1.2.7/) version which is the one we currently have in our code.

PR #317, opened 2019-10-10
Fix/update location component

vanbasten17:
In this PR I've applied the following changes:
* The documentation for this component was wrong, I have updated it.
I have improved also the following things for `botonic serve`:
* We had an url for Google Maps which was not working as expected, with this one it seems to work properly.
* I have applied little styles to the current text being displayed.
* I have added the prop `text` in order to redefine the name of the copy which appears when the component is displayed, as we don't have a way to redefine it from an upper level. As before we only use serve for testing the bots, we didn't need this, but now that we have the webchat I think that it makes sense.

### DIFF
--- a/packages/botonic-react/src/components/location.jsx
+++ b/packages/botonic-react/src/components/location.jsx
@@ -8,10 +8,9 @@ const serialize = locationProps => {
 }
 
 export const Location = props => {
+  let lat = parseFloat(props.lat)
+  let long = parseFloat(props.long)
   const renderBrowser = () => {
-    let lat = parseFloat(props.lat)
-    let long = parseFloat(props.long)
-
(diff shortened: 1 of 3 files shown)


packages/botonic-react/src/components/location.jsx
```diff
@@ -31,8 +33,8 @@ export const Location = props => {
   const renderNode = () => {
     return (
       <message type='location'>
-        <lat>{props.lat}</lat>
-        <long>{props.long}</long>
+        <lat>{lat}</lat>
```
guinii: Is this variable accessible here? I don't think so!

vanbasten17: i think so, renderNode is inside the scope where props is defined

PR #330, opened 2019-10-17
Watson Assistant Plugin

guinii:


packages/botonic-plugin-watson/src/index.js
```diff
@@ -0,0 +1,38 @@
...
+      })
+      intent = res.result.intents[0].intent
+      confidence = res.result.intents[0].intent.confidence
+      intents = res.result.intents
+      entities = res.result.entities
+    } catch (error) {
+      console.log(error)
+    }
+    Object.assign(input, { intent, confidence, intents, entities })
+    return { input, session, lastRoutePath }
```
dpinol: please remove the return, as it's not used

packages/botonic-plugin-watson/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@botonic/plugin-watson",
+  "version": "0.9.0",
+  "main": "src/index.js",
+  "dependencies": {
+    "ibm-watson": "latest"
```
dpinol: I would fix at least the minor version to minimize disruption when they upgrade

packages/botonic-plugin-watson/package.json
```diff
@@ -0,0 +1,11 @@
...
+  },
+  "devDependencies": {
+    "@babel/runtime": "^7.5.5"
```
dpinol: I would remove it, as react already has it

PR #347, opened 2019-10-25
Site/initial

MarcosCA:


website/site/src/components/common/BotDemo.js
```diff
@@ -0,0 +1,53 @@
+import React from "react"
```
dpinol: our coding standard defines using single quotes.
I suggest taking temporarilly .prettierc from https://github.com/hubtype/botonic/pull/353 and fixing it in a  commit with no more changes

dpinol:
@MarcosCA can you merge this?

PR #361, opened 2019-10-30
Add example folder with hubtype demo

guinii:


examples/hubtype_demo/package.json
```diff
@@ -0,0 +1,99 @@
+{
+  "scripts": {
+    "build": "webpack --env.target=all --mode=production",
+    "build_development": "webpack --env.target=node --mode=development",
+    "build_with_tests": "tsc -b tests/tsconfig.json",
+    "lint": "node_modules/.bin/eslint --fix --quiet 'src/**/*.ts*' 'tests/**/*.ts*'",
```
dpinol: add --cache flag for faster performance

dpinol: do 
`npm i -D eslint_d`
and replace node_modules/.bin/eslint node_modules/.bin/eslint_d for even faster performance :-)
https://www.npmjs.com/package/eslint_d

examples/hubtype_demo/src/config.ts
```diff
@@ -0,0 +1,9 @@
+export class Staging {
+  spaceId = ''
+  accessToken = ''
+}
+
+/*export class Production {
```
dpinol: I would remove this and rename Staging to ContentfulConfig

examples/hubtype_demo/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+cd bin
+./import_contentful.sh
+```
+
+Now, in Contentful you should have new Content model objects.
+
+### Connect our bot project with Contentful
+
+Open the file `src/config.ts`. Here we will define all our configurations. - The `spaceId` is the same value that we set before in the `bin/credentials` file. - The `accessToken` is a new value, in order to get it, go to Contentful in `Settings-ApiKeys-Content delivery/preview tokens` and press the `Add API key` in the right.
```
dpinol: I think it's more standard to use pipes for config paths:
Settings | ApiKeys | Content delivery/preview tokens

examples/hubtype_demo/bin/import_contentful.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/zsh
+BIN_DIR=${0:a:h}
+
```
dpinol: add 
`cd $BIN_DIR`
in case the script is executed from top folder

examples/hubtype_demo/README.md
```diff
@@ -0,0 +1,49 @@
+## Set up
+
+npm i -g contentful-cli
+npm i package.json
+
+### Import space to Contentful:
+
+Create an account into Contentful
```
dpinol: I would add.
`See https://docs.botonic.io/plugins/plugin-contentful for more details`

examples/hubtype_demo/src/actions/wait.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+export default class extends React.Component {
+  render() {
+    return <></>
+  }
+}
```
vanbasten17: what is doing this action, can be removed?

examples/hubtype_demo/src/locales/en.ts
```diff
@@ -0,0 +1,2 @@
+export default {
+};
```
vanbasten17: can we define the copies in locales? in this way would be easier to maintain if we have to do modifications in this bot.

examples/hubtype_demo/src/routes.ts
```diff
@@ -0,0 +1,56 @@
...
+    _.input.data == 'reset' ||
+    _.input.data == 'Reset' ||
+    _.input.data == 'hola' ||
+    _.input.data == 'Hola' ||
+    _.input.data == 'buenas' ||
+    _.input.data == 'Buenas' ||
+    _.input.data == 'hello' ||
+    _.input.data == 'Hello' ||
+    _.input.data == 'hi' ||
+    _.input.data == 'Hi' ||
```
vanbasten17: here we could store these keywords inside a variable holding the regex itself.
`let introRegex = /reset|hola|buenas|hello|hi/i`, with `i` flag we can ignore case.
`_input.data.match(introRegex)`.
I think that in this way would be tidier.

examples/hubtype_demo/src/routes_bank.js
```diff
@@ -0,0 +1,104 @@
...
+    action: GeneralTransfer,
+    childRoutes: [
+      {
+        path: 'feedback',
+        text: /\b(?!reset|hola|buenas|hello|hi|webview|webviews)\b\S+/i,
+        action: Feedback,
+        childRoutes: [
+          {
+            path: 'wait',
+            text: /\b(?!reset|hola|buenas|hello|hi|webview|webviews)\b\S+/i,
```
vanbasten17: this regex is repeated several times in this file, use a var holding the regex instead?

examples/hubtype_demo/src/actions/wait.js
guinii: It's a requirement that Marc wanted, after a handoff, de bot don't reply, just ignore the return, I remove the file, and put this behaviour inside `routes.js`

examples/hubtype_demo/src/locales/en.ts
guinii: Good  point, the idea is to have all the text inside Contentful. I create a new task for doing it

examples/hubtype_demo/src/routes_bank.js
```diff
@@ -32,13 +34,13 @@ export const BankRoutes = [
     childRoutes: [
       {
         path: 'feedback',
-        text: /\b(?!reset|hola|buenas|hello|hi|webview|webviews)\b\S+/i,
+        text: restartFlowRegex,
```
dpinol: 👏

examples/hubtype_demo/src/locales/en.ts
dpinol: @guinii @vanbasten17 remember that this is a demo. Let's not overengineer it.

PR #376, opened 2019-11-06
Contentful/keep stems tokens

dpinol:
Refactor Keyword and NormalizedUtterance so that we keep both the token and the stem.
This allows:
* We now match similar words to words which in the stemming blacklist
* Now we can match keywords which only contain stopwords when input contains other (no stopwords) words

### DIFF
--- a/packages/botonic-plugin-contentful/src/nlp/locales.ts
+++ b/packages/botonic-plugin-contentful/src/nlp/locales.ts
@@ -1,6 +1,12 @@
 import { stemmerFor } from './stemmer'
 
+// TODO convert to string enum?
 export type Locale = string
+export const SPANISH = 'es'
+export const CATALAN = 'ca'
+export const ENGLIGH = 'en'
+export const PORTUGUESE = 'pt'
+export const POLISH = 'pl'
 
 export function checkLocale(locale?: Locale): Locale {
   if (!locale) {
--- a/packages/botonic-plugin-contentful/src/nlp/normalizer.ts
+++ b/packages/botonic-plugin-contentful/src/nlp/normalizer.ts
@@ -30,23 +30,42 @@ export class StemmingBlackList {
   }
 }
 
-export class NormalizedUtterance {
-  joinedStems: string
-
+export class Word {
   /**
-   *
-   * @param raw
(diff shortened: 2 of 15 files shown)


packages/botonic-plugin-contentful/src/nlp/locales.ts
```diff
@@ -1,6 +1,12 @@
 import { stemmerFor } from './stemmer'
 
+// TODO convert to string enum?
 export type Locale = string
+export const SPANISH = 'es'
+export const CATALAN = 'ca'
+export const ENGLIGH = 'en'
+export const PORTUGUESE = 'pt'
+export const POLISH = 'pl'
```
vanbasten17: I'm not sure if enums exists as is in Javascript, but maybe you could just create the object languages with `freeze` in this way:
```
export const locales = Object.freeze({
SPANISH: 'es',
CATALAN:'ca',
....
})
```
when accessing them would look like `locales.SPANISH` which looks pretty good.

packages/botonic-plugin-contentful/src/nlp/normalizer.ts
```diff
@@ -86,25 +105,25 @@ export class Normalizer {
     const stemmer = stemmerFor(locale)
     // tokenizer will replace i18n characters
     const tokens = this.tokenizer(locale).tokenize(txt, true)
-    let stems: string[] = []
+    let words: Word[] = []
     const stopWords = this.stopWordsPerLocale[locale]
+    let numStopWords = 0
     for (const token of tokens) {
       const black = this.getBlackListStem(locale, token)
```
vanbasten17: `if isBlacklist`...

packages/botonic-plugin-contentful/src/nlp/locales.ts
dpinol: in TS there's enum. in the future, I'll give it a try

packages/botonic-plugin-contentful/src/nlp/normalizer.ts
dpinol: actually it needs to be like this, because getBlackListStem returns the stem from the blacklist structure.
Eg. getBlackListStem("pedidos") will return "pedido)

PR #381, opened 2019-11-11
feat: getting plugins config from lambda

vanbasten17:
Hi all! This PR is related to: https://github.com/metis-ai/hubtype-backend/pull/87
In this PR I added the method `getConfig` to be called from the lambda function in order to retrieve the plugins that the user has integrated in his bot.

### DIFF
--- a/packages/botonic-core/src/core-bot.js
+++ b/packages/botonic-core/src/core-bot.js
@@ -18,7 +18,7 @@ export class CoreBot {
     defaultRoutes,
   }) {
     this.renderer = renderer
-    this.plugins = plugins ? loadPlugins(plugins) : []
+    this.plugins = loadPlugins(plugins)
     this.theme = theme || {}
     this.defaultTyping =
       typeof defaultTyping !== 'undefined' ? defaultTyping : 0.6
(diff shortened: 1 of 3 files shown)


packages/botonic-core/src/core-bot.js
```diff
@@ -15,10 +15,10 @@ export class CoreBot {
     appId,
     defaultTyping,
     defaultDelay,
-    defaultRoutes
+    defaultRoutes,
   }) {
     this.renderer = renderer
-    this.plugins = loadPlugins(plugins)
+    this.plugins = plugins ? loadPlugins(plugins) : []
```
guinii: Maybe this logic can be done inside `loadPlugins` in file `src/plugins` because we are checking there if `plugins` is not defined. We need to return only `[]` instead of `null`

packages/botonic-core/src/core-bot.js
```diff
@@ -44,6 +44,10 @@ export class CoreBot {
     session.__locale = locale
   }
 
+  getPlugins() {
```
guinii: Why is this method necessary? `plugins` is a public value of the CoreBot.

packages/botonic-core/src/core-bot.js
vanbasten17: yes, i'm agree, i have refactored it as you suggest 👍

packages/botonic-core/src/core-bot.js
vanbasten17: yes, i thought that maybe would be clearer with a getter, but as it's public, makes sense we can retrieve it directly, changed 👍

PR #377, opened 2019-11-06
Release/0.10.0

vanbasten17:
**Current changelog**

- Mobile responsive
- Added new customizable scrollbars (enabled by default in Windows)
- Export staticAsset in @botonic/react's index to make it available from bot's project
- Enabled attachments
- Messages bubbles (bot/user) pseudo-elements are now modified according to botMessageStyle or userMessageStyle 
- Customizable text input area
- Customizable typography (fontFamily) attribute
- Fixed typing indicator not showing up when message had delay/typing
- Now typing delay working in production (added new export in bot's `index.js`: `export const config = { defaultDelay: 1, defaultTyping: 2 }`)
- Removed console errors generated by @rebass/grid dep.
- Customizable styles for Button and Reply components
- New handoff options
- Carousels with one element don't break
- PersistentMenu, adding `{closeLabel: 'Your_copy'}` inside persistentMenu, will change the label

PR #382, opened 2019-11-11
Fix/handoff json

dpinol:
refactor #297 broke on_finish handoffs for payloads. This PR tries to fix it. 

Since botonic handoff call now has many arguments, now we can pass them in json format.


vanbasten17:
Amazing work Dani!! 💪 Thanks for fixing the broken part! 🙇🏻‍♂️

vanbasten17:
I have tested it locally a now seems to work properly! 👍 I had to add the export HandoffBuilder in `@botonic/core/index.js` to get it work

dpinol:
Applied to #377

PR #387, opened 2019-11-14
revert(core): on_finish handoff path again within payload

dpinol:
Revert part of  #382 

on_finish handoff path again within payload with __PATH_PAYLOAD__ prefix

### DIFF
--- a/packages/botonic-core/src/handoff.test.js
+++ b/packages/botonic-core/src/handoff.test.js
@@ -5,17 +5,17 @@ describe('handOff', () => {
     [
       `create_case:{
         "queue":"q1",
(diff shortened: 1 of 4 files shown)


packages/botonic-core/src/handoff.test.js
```diff
@@ -31,22 +31,22 @@ describe('handOff', () => {
           agent_email: 'email1',
           case_info: '{}{:::: m"ho menjo tot}',
           note: '{}{:::: m"ho menjo tot2}',
```
vanbasten17: maybe we should change these literals in tests, we have forgotten removing them

PR #390, opened 2019-11-15
feat(react): decorator for msgToBotonic

dpinol:
See usage at https://github.com/hubtype/botonic/compare/release/0.10.0...react/render-decorator?expand=1#diff-0c4c312cd83fa0d4c66ec3045b76f2d3R47

I only implemented it for Text. If the design is approved, it's trivial to extend it for other types of messages


ericmarcos:
This can be achieved via [custom message types](https://docs.botonic.io/tutorials-and-examples/custom-webchat#custom-message-styles)

Issue #409, opened 2019-12-06
Cannot receive url parameters


bndby:
**Describe the bug**

Try to pass URL parameters (https://docs.botonic.io/tutorials-and-examples/guided-flow#working-with-url-parameters), but receive an empty params

**To Reproduce**

index.js
```
...
<Button key={site.id} path={`menu?id=${site.id}`}>
{site.title}
</Button>
...
```

menu.js
```
...
render() {
    //
    console.log('Menu', this.context.params);
...
```

result of console.log():
```
params: {
e: ""
__proto__: Object
}
```

**What is the expected behaviour?**
Url parameters in params

**About the environment**
node 12
botonic 0.10.0

 Did this work in previous versions of botonic? 
work fine on 0.9.0

**Additional context**

vanbasten17:
Hi @bndby, thanks for your feedback! 
This bug will be fixed in the next [Botonic release](https://github.com/hubtype/botonic/pull/422).
Closing this issue.

PR #422, opened 2019-12-19
Release/0.10.1

vanbasten17:
- Bump all versions for @botonic/core, @botonic/react and @botonic/cli from `0.10.0` to `0.10.1`.
- Bump all templates @botonic/react deps from `0.10.0` to `~0.10.1` in order to update always to latest patch version automatically.
- Add `config` with defaultTyping and defaultDelay by default for every template.

**Fixes**
- Allow calling dynamic routes (routes as a function) in nested routes
- Fix issue https://github.com/hubtype/botonic/issues/409

**Feats**
- Getting configuration for bots plugins

### DIFF
--- a/packages/botonic-cli/templates/blank/package.json
+++ b/packages/botonic-cli/templates/blank/package.json
@@ -16,7 +16,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.6.2",
-    "@botonic/react": "0.10.0"
+    "@botonic/react": "~0.10.1"
   },
   "devDependencies": {
     "@babel/core": "^7.6.2",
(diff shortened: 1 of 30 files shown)


packages/botonic-cli/templates/blank/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.6.2",
-    "@botonic/react": "0.10.0"
+    "@botonic/react": "~0.10.1"
```
dpinol: I think we can change it to ^0.10.1 to avoid changing it so often. @ericmarcos what do you think?

vanbasten17: i talked with him and we agreed to leave it as '~' to upgrade only to next patches

dpinol:
@vanbasten17 can you try creating a test for the fixed bug?

dpinol:
can you merge it? the test can be added in a different PR

PR #470, opened 2020-01-21
feat: allow to deploy bot with email and password throw cli

guinii:


packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -34,26 +34,40 @@ Uploading...
...
     track('Deployed Botonic CLI')
 
     force = flags.force ? flags.force : false
     npmCommand = flags.command
-    let botName = flags.botName ? flags.botName : false
-
-    if (!this.botonicApiService.oauth) await this.signupFlow()
-    else if (botName) {
-      this.deployBotFromFlag(botName)
+    this.botName = flags.botName ? flags.botName : undefined
```
dpinol: would this work?
```
this.botName = flags.botName || undefined
```

packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -34,26 +34,40 @@ Uploading...
...
+    let email = flags.email ? flags.email : undefined
+    let password = flags.password ? flags.password : undefined
```
vanbasten17: same here as Dani suggest above. I'm wondering if we could update @botonic/cli to latest typescript versions to use the newest features/syntax as for example https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

dpinol: @vanbasten17 good point, actually cli already uses 3.7!
actually I'm not use why @guinii converts the empty values to undefined

guinii: Both of you are right! I change the declaration! Thanks guys

PR #513, opened 2020-02-05
Feat/create GitHub action test

MarcosCA:


.github/workflows/botonic-core.tests.yml
```diff
@@ -0,0 +1,28 @@
...
+        node-version: '12'
+    - name: Setting up cache (Step 3 of 6)
+      uses: actions/cache@v1
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+    - name: Install botonic-core (Step 4 of 6)
+      run: (cd ./packages/botonic-core && npm install)
```
dpinol: Initially I thought that we should verify that we can build from package.json creating the lock.json file with "npm i -D"
,but I have mixed feelings. Let's keep it simple and leave it as it is :+1:

.github/workflows/botonic-core.tests.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Build botonic-core (Step 5 of 6)
+      run: (cd ./packages/botonic-core && npm run build)
```
dpinol: all workflows are very similar. 
And we have still need to add other plugins (watson, nlu, segment,...)
In a separate PR I'd investigate if we can reuse a common workflow to avoid repetitions. Maybe with
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsuses ?

dpinol:
please rebase it

PR #515, opened 2020-02-05
Feat/animations flag and send button

vanbasten17:
Just added a pair of configurable properties, `userInput.sendButton.enable` and `animations.enable`, both inside the `theme` object of webchat, as shown below:
```
userInput: {
      sendButton:{
        enable:true, // set it to false to disable send button (enabled by default)
        custom: CustomSendButton // React Component
      }
    },
```
```
    animations: {
      enable: true // set it to false to disable animations (enabled by default)
    },
```
**breaking change (for enabling/disabling) emojiPicker**: 
`userInput.emojiPicker` --> `userInput.emojiPicker.enable`, 
and instead of (alternative api):
`emojiPicker` --> `enableEmojiPicker`

Screenshot 2020-02-05 at 21 12 28

### DIFF
--- a/packages/botonic-react/src/utils.js
+++ b/packages/botonic-react/src/utils.js
@@ -61,3 +61,6 @@ export const _getThemeProperty = theme => (
     }
   }
 }
+
+export const ConditionalWrapper = ({ condition, wrapper, children }) =>
+  condition ? wrapper(children) : children
--- a/packages/botonic-react/src/webchat/webchat.jsx
+++ b/packages/botonic-react/src/webchat/webchat.jsx
@@ -506,6 +506,7 @@ export const Webchat = forwardRef((props, ref) => {
     'userInput.sendButton.enable',
     true
   )
+  const CustomSendButton = getThemeProperty('userInput.sendButton.custom')
   const inputUserArea = () => {
     return (
       userInputEnabled && (
@@ -588,7 +589,7 @@ export const Webchat = forwardRef((props, ref) => {
                 />
               </ConditionalWrapper>
             )}
-            {sendButtonEnabled && (
+            {(sendButtonEnabled || CustomSendButton) && (
               <ConditionalWrapper
                 condition={animationsEnabled}
                 wrapper={children => (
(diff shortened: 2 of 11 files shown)


packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -206,12 +212,19 @@ export const Webchat = forwardRef((props, ref) => {
       />
     </div>
   )
-
-  const persistentMenuOptions =
-    getThemeProperty('userInput.persistentMenu') || props.persistentMenu
+  const animationsEnabled = getThemeProperty('animations.enable', true)
+  const persistentMenuOptions = getThemeProperty(
```
guinii: Here we lose `props.persistentMenu`?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -225,12 +238,11 @@ export const Webchat = forwardRef((props, ref) => {
...
         <img src={staticAsset(LogoMenu)} />
       </div>
-    </motion.div>
+    </ConditionalWrapper>
   )
 
   const checkBlockInput = input => {
-    let blockInputs =
-      getThemeProperty('userInput.blockInputs') || props.blockInputs
+    let blockInputs = getThemeProperty('userInput.blockInputs', undefined)
```
guinii: Here also `props.blockInputs`?

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: already in theme an getting in with getThemeProperty 👍

packages/botonic-react/src/utils.js
```diff
@@ -61,3 +61,6 @@ export const _getThemeProperty = theme => (
     }
   }
 }
+
+export const ConditionalWrapper = ({ condition, wrapper, children }) =>
```
dpinol: this is so cool :100:

vanbasten17: yes it's so cool! Reference: https://blog.hackages.io/conditionally-wrap-an-element-in-react-a8b9a47fab2

PR #516, opened 2020-02-06
Core/fix routes call

dpinol:
When routes is a function, from route redirection and in documentation it 1 was calling like this:

`routes(session,input)
`
instead of

`routes({session,input})
`

@guinii when this is published, we should adapt projects which implemented the routes function with the wrong syntax


MarcosCA:
There are conlficts now

dpinol:
@guinii @vanbasten17 PR merged, please mention API breaking change in release notes (no need to make a big deal, as most bots won't be affected) and inform affected customers

PR #519, opened 2020-02-06
Contentful/content id field

dpinol:
So far we used an id to retrieve CMS contents, but the returned object did not have an id type.
In this PR we add it to the CommonFields object, which all TopContents have

### DIFF
--- a/packages/botonic-core/index.d.ts
+++ b/packages/botonic-core/index.d.ts
@@ -18,23 +18,24 @@ export interface Input {
   entities?: any
 }
 
-type StringMatcher = RegExp | string | ((data: string) => boolean)
+
+export interface SessionUser {
+  id: string
+  // login
+  username?: string
+  // person name
+  name?: string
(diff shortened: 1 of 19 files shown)


packages/botonic-core/index.d.ts
```diff
@@ -105,20 +118,29 @@ export declare function getAvailableAgents(
...
- * See the @botonic/react's index.d.ts for this implementations particular type.
+ * See Response at @botonic/react's index.d.ts for the React type
  * */
-export type Response = any
+export interface BotResponse extends BotRequest {
+  response: any
+}
+
+export type PluginPreRequest = BotRequest
+export type PluginPostRequest = BotRequest & {
```
vanbasten17: Use maybe BotResponse instead of BotRequest & {response: any}?

dpinol: yes, I'll do it in another PR

PR #530, opened 2020-02-10
Chore/react lint fixes

dpinol:
Depends on #526 
- eslint now enabled for  jsx files on botonic/react
- fixes for eslint react warnings
- fix for https://mathiasbynens.github.io/rel-noopener/ for location links

PR #535, opened 2020-02-11
fix: add tests with multibutton carousel

vanbasten17:
Depends on #530
* Support carousel with multibutton


dpinol:
merged through #547

PR #536, opened 2020-02-11
Chore/react request context

dpinol:
Depends on #535 
Now RequestContext for test has RequestContext default values


dpinol:
merged through #547

PR #541, opened 2020-02-12
Fix: Update hooks.js - Fix outer page scrolling bug

YoniSchirris:
Previously, the file used "scrollIntoView()" on the latest message element. This produced buggy behavior: When the chat was on a page that itself was scrollable, the entire page would scroll to get the bottom of the chat into view.

Currently, we 'simply' scroll the scrollable element of the chatbot all the way down. No matter where the bot is on the parent page, it will not influence the location of the parent page.

TODO: To get the exact same smooth scrolling behaviour, we need to add 
.simplebar-content-wrapper{scroll-behavior: smooth}
I currently can't find your CSS files.

### DIFF
--- a/packages/botonic-react/src/webchat/hooks.js
+++ b/packages/botonic-react/src/webchat/hooks.js
@@ -178,10 +178,10 @@ export function useWebchat() {
 export function useTyping({ webchatState, updateTyping, updateMessage }) {
   useEffect(() => {
     let delayTimeout, typingTimeout
-    let end = document.getElementById('messages-end')
-    if (end) {
-      end.scrollIntoView({ behavior: 'smooth' })
-      setTimeout(() => end.scrollIntoView({ behavior: 'smooth' }), 100)
+    let frame = document.querySelectorAll('.simplebar-content-wrapper')[0];
+    if (frame) {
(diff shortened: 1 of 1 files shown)


YoniSchirris:
Note that it **will** have the smooth behavior on most browsers with CSS: https://caniuse.com/#search=scroll-behavior

dpinol:
@YoniSchirris could you rebase the branch?

YoniSchirris:
@dpinol What do you mean exactly? I thought @marcrabat would look into 
1. Adjusting class name / id to prevent conflicts
2. adding the smooth scroll CSS to this adjusted class name / ID

YoniSchirris:
@dpinol What do you mean exactly? I thought @marcrabat would look into 
1. Adjusting class name / id to prevent conflicts
2. adding the smooth scroll CSS to this adjusted class name / ID

vanbasten17:
Hi @YoniSchirris! First of all thanks for your contribution, we are so happy that developers like you contribute to make this project better and as a result, bigger. 💯💯💯 
We have just created a new PR against our master branch solving this issue inspired by your proposal and will be available in the next release after being approved by our contributors.

PR #543, opened 2020-02-12
fix(react): error when recovering messages and add text tests

vanbasten17:
By chance I've discovered this bug we had when after setting "keep session on reload".
The component I was rendering in question is this:
```
<>
        {[1, 2].map((e, i) => {
          return (
            <Text key={i}>
              This is the message {i + 1}
              <Button payload={'lele'}>
                Text {i}.{i + 1}
              </Button>
              <Button payload={'lele'}>
                Text {i}.{i + 2}
              </Button>
            </Text>
          )
        })}
</>
```
So when it is rendered appears as expected:
Screenshot 2020-02-12 at 18 58 20 

But after refreshing the page with "Keep session on reload" checked, it wasn't loading the previous message properly:

Screenshot 2020-02-12 at 18 58 49

This is due to the fact that we weren't taking into account numeric generated indices when serializing the component:
Screenshot 2020-02-12 at 18 54 54

This PR should solve this issue. Also I have added some simple tests for Text components.


dpinol:
@vanbasten17 rebase and merge

packages/botonic-react/package.json
```diff
@@ -49,9 +49,14 @@
     "@types/react": "16.9.23",
     "babel-dts-generator": "^0.6.3",
     "babel-plugin-add-module-exports": "^1.0.2",
+    "eslint": "^6.8.0",
```
dpinol: @dpinol I'm gonna remove this in a different PR

vanbasten17: okay, as commented now all depends on the configuration of the parent. 👍

PR #547, opened 2020-02-13
Chore/unify dev dependencies and improve index.d.ts

dpinol:
Depends on #542 
* fix eslint for jsx
* unify dev dependencies in package.json
* unify .eslintrc
* organize index.d.ts in different folders
* change React.Component types to React.FunctionComponent


packages/botonic-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/react",
-  "version": "0.10.1",
+  "version": "0.11.0-alpha.0",
```
vanbasten17: @dpinol we already published a version 0.11.0-alpha.0 https://www.npmjs.com/package/@botonic/react/v/0.11.0-alpha.0 which was needed by Zurich. Maybe rename it as 0.11.0-alpha.1 better.

packages/botonic-react/package.json
```diff
@@ -17,7 +17,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "@botonic/core": "^0.10.1",
+    "@botonic/core": "0.11.0-alpha.0",
```
vanbasten17: in case of botonic/core, we didn't published it as it wasn't needed, but let's leave it as well as 0.11.0-alpha.1 to be consistent

packages/botonic-core/src/handoff.js
```diff
@@ -140,9 +140,9 @@ export async function getAvailableAgentsByQueue(session, queueId) {
 }
 
 export async function getAvailableAgents(session) {
-  let baseUrl = session._hubtype_api || 'https://api.hubtype.com'
-  let botId = session.bot.id
-  let resp = await axios({
+  const baseUrl = session._hubtype_api || 'https://api.hubtype.com'
+  const botId = session.bot.id
+  const resp = await axios({
```
vanbasten17: "let"ophobia? 😆 is it very wrong to use `let` in all the pieces of code of this commit?

vanbasten17:
Nice job as always!! Remember to bump to 0.11.0-alpha.1 🚀🚀🚀

packages/botonic-react/package.json
dpinol: good point, actually the last commit of the PR already upgrades to alpha.1
It's nice that npm (I guess when publishing) automatically upgrades from alpha/beta.X to alpha/beta.X+1

packages/botonic-core/src/handoff.js
dpinol: :-D
I'm not too strong on this one
some people are not willing to pay the 2 extra keystrokes, and I respect that
But immutability eases your mind and side-effects are a bitch, so I left it since it's in the eslint recommended list

vanbasten17: hahaha nice!! 🎉

PR #593, opened 2020-02-28
Contentful/refactors

dpinol:
Several refactors. Some of them introduce API incompatibilities wrt 0.10 version
* Now Content also have id and name fields
* ModelType renamed to TopContentType. Added new MesseContentType and others
* Method CMS.contents renamed to toContents. Added CMS.contents for all content types

### DIFF
--- a/packages/botonic-plugin-contentful/package-lock.json
+++ b/packages/botonic-plugin-contentful/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/plugin-contentful",
-  "version": "0.10.4",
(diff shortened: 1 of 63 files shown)


dpinol:
> Remind everybody using bots with latests versions to have changed properly lines like the ones we found previously in volkswagen:
> `return schedule.contains(now)` -> `return schedule.schedule.contains(now)`
> or
> `resource => (copies[resource['name']] = resource['text']` --> `resource => (copies[resource['common']['name']] = resource['text'])`
> Gas!! rocketrocketrocket

You're right. Fortunately with TS it's so easy to find these changes

PR #603, opened 2020-02-28
feat(core): add handoff options for shadowing, cancelHandoff and deleteUser

ericmarcos:


packages/botonic-core/src/handoff.js
```diff
@@ -151,3 +161,12 @@ export async function getAvailableAgents(session) {
   })
   return resp.data
 }
+
+export function cancelHandoff(session, typification = null) {
+  let params = { typification }
```
dpinol: please run "pre-commit install" to catch these things

packages/botonic-core/src/handoff.js
```diff
@@ -151,3 +161,12 @@ export async function getAvailableAgents(session) {
...
+  session._botonic_action = `discard_case:${JSON.stringify(params)}`
```
dpinol: when typification is not passed this will "generate discard_case:null". Does backend support this?

packages/botonic-core/src/handoff.js
```diff
@@ -151,3 +161,12 @@ export async function getAvailableAgents(session) {
...
+}
+
+export function deleteUser(session) {
```
dpinol: please add this to botonic-core/index.d.ts:
`export function deleteUser(session: Session): void`

packages/botonic-core/src/handoff.js
```diff
@@ -49,14 +49,20 @@ export class HandOffBuilder {
     return this
   }
 
+  withShadowing(shadowing = true) {
```
dpinol: please add this to HandOffBuilder in botonic-core/index.d.ts:
`withShadowing(shadowing = true): HandOffBuilder`

packages/botonic-core/src/handoff.js
```diff
@@ -151,3 +161,12 @@ export async function getAvailableAgents(session) {
   })
   return resp.data
 }
+
+export function cancelHandoff(session, typification = null) {
```
dpinol: once Caroline mentioned that [typification](https://en.wikipedia.org/wiki/Typification) was not a correct word for this. I suggest [classification](https://www.collinsdictionary.com/es/diccionario/espanol-ingles/tipificaci%C3%B3n)

ericmarcos: I agree, but I prefer to do this refactor in botonic and hubtype at the same time. Leaving it like this for now.

packages/botonic-core/index.d.ts
```diff
@@ -118,6 +124,13 @@ export declare function getAvailableAgents(
   session: Session
 ): Promise<{ agents: HubtypeAgentsInfo[] }>
 
+export declare function cancelHandoff(
+  session: Session,
+  typification: string
```
dpinol: should be "typification?: string"

packages/botonic-core/index.d.ts
```diff
@@ -80,6 +85,7 @@ export class HandOffBuilder {
   withAgentEmail(email: string): HandOffBuilder
   withNoteURL(note: string): HandOffBuilder
   withCaseInfoURL(caseInfo: string): HandOffBuilder
+  withShadowing(shadowing: boolean): HandOffBuilder
```
dpinol: withShadowing(shadowing **= true**): HandOffBuilder

ericmarcos: if I do this typescript complains: `A parameter initializer is only allowed in a function or constructor implementation.`

dpinol: that's true, my bad. It should be:
`withShadowing(shadowing?:boolean): HandOffBuilder`

dpinol:
ignore contentful test errors. they'll be fixed in next PR

PR #615, opened 2020-03-07
Contentful/csv export was not exporting keywords

dpinol:
Depends on #614


dpinol:
please ignore CI errors. Fixed on #616

PR #641, opened 2020-03-16
Chore(dynamo/contentful)eslint on .eslintrc*.js

dpinol:
.eslintrc*.js were not being formatted by prettier.
Split in 2 commits because an extra tsconfig file is required to avoid an eslint error when specifying the eslintsrc files to test in the eslint command with wildcards https://github.com/typescript-eslint/typescript-eslint/issues/967#issuecomment-531817014

### DIFF
--- a/packages/botonic-core/package.json
+++ b/packages/botonic-core/package.json
@@ -7,7 +7,7 @@
     "test": "../../node_modules/.bin/jest --coverage",
     "prepare": "node ../../preinstall.js",
     "lint": "npm run lint_ci -- --fix",
-    "lint_ci": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.js*' '**/*.d.ts'",
+    "lint_ci": "../../node_modules/.bin/eslint_d --cache --quiet '.*.js' '*.js' 'src/**/*.js*' '**/*.d.ts'",
     "build": "rm -rf lib && babel src --out-dir lib --source-maps --copy-files"
   },
   "repository": {
--- a/packages/botonic-plugin-dynamodb/package.json
+++ b/packages/botonic-plugin-dynamodb/package.json
@@ -5,7 +5,7 @@
     "test": "../../node_modules/.bin/jest",
     "lint": "npm run lint_core -- --fix",
     "lint_ci": "npm run lint_core -- -c .eslintrc_slow.js",
-    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.ts*' 'tests/**/*.ts*'",
+    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet '.*.js' 'src/**/*.ts*' 'tests/**/*.ts*'",
     "prepare": "node ../../preinstall.js",
     "prepublishOnly": "npm run build && npm test && npm run lint",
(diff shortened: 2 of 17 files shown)


packages/botonic-core/package.json
```diff
@@ -7,7 +7,7 @@
     "test": "../../node_modules/.bin/jest --coverage",
     "prepare": "node ../../preinstall.js",
     "lint": "npm run lint_ci -- --fix",
-    "lint_ci": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.js*'",
+    "lint_ci": "../../node_modules/.bin/eslint_d --cache --quiet '.*.js' '*.js' 'src/**/*.js*'",
```
marcpdw: Is this duplication of the kind of files intended? Probably we can remove one `'*.js'`, right?

packages/botonic-react/package.json
```diff
@@ -9,7 +9,7 @@
     "build": "rm -rf lib && babel src --out-dir lib --source-maps --copy-files",
     "lint": "npm run lint_core -- --fix",
     "lint_ci": "npm run lint_core",
-    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.js*' '**/*.d.ts' 'tests/**/*.js' 'tests/**/*.jsx'"
+    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet '.*.js' '*.js' 'src/**/*.js*' '**/*.d.ts' 'tests/**/*.js' 'tests/**/*.jsx'"
```
marcpdw: Is it correct to be `.*.js` instead of `*.js`?

packages/botonic-core/package.json
dpinol: nope, try "ls *.js" from bash and you'll see that the hidden files do not appear.
globs in package.json work likewise

packages/botonic-react/package.json
dpinol: we need both

packages/botonic-plugin-dynamodb/package.json
```diff
@@ -5,7 +5,7 @@
     "test": "../../node_modules/.bin/jest",
     "lint": "npm run lint_core -- --fix",
     "lint_ci": "npm run lint_core -- -c .eslintrc_slow.js",
-    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.ts*' 'tests/**/*.ts*'",
+    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet '.*.js' 'src/**/*.ts*' 'tests/**/*.ts*'",
```
marcpdw: One last observation... I guess that here there is no `*.js` because all the files are `*.ts`, right?
Now I think I see the differences among the several linter files

dpinol:
> LGTM

@marcpdw I didn't know this acronym :-) https://www.urbandictionary.com/define.php?term=LGTM

packages/botonic-plugin-dynamodb/package.json
dpinol: yes, in some places there are babel.config.js or similar. I think that if you put a glob which finds nothing it complains, so I have to tailor it to each project

PR #649, opened 2020-03-24
fix(contentful): TokenizerCa to splits "donar-se" but not "covid-19"

dpinol:
### Description
TokenizerCa should split "donar-se" but not "covid-19"

### Context
For Catalan, we use our own tokenizer since node-nlp does not stem correctly "adonar-se'm"


packages/botonic-plugin-contentful/tests/nlp/stemmer.test.ts
```diff
@@ -31,3 +32,25 @@ test.each<any>([
     expect(stemmed).toEqual(expected)
   }
 )
+
+describe('Stemmers', () => {
+  test.each<any>(SUPPORTED_LOCALES)(
+    'TEST numbers in words are kept for locale %s',
+    (locale: Locale) => {
+      // TODO remove when node-nlp PR is merged
+      if (locale == PORTUGUESE) {
```
marcpdw: nit: I like to use the strict comparator, `===` instead of `==` when possible

dpinol: I agree. done, thanks

PR #650, opened 2020-03-24
Contentful/test timeout and error improvements

dpinol:
### Description
* Improve message in exceptions raised by contentful plugin
* Extend timeouts
* Wait for contentful.com content is updated before checking the new value
* Exponential backoff class 

### Context
Contentful tests which modify contents on contentful.com sometimes failed
because we we're checking the updated value before their CDN updated the new
version

### DIFF
--- a/packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
@@ -75,9 +75,10 @@ export class ManageContentful implements ManageCms {
       )
     }
     if (!context.allowOverwrites) {
-      if (entry.fields[field.cmsName][context.locale]) {
+      const value = entry.fields[field.cmsName][context.locale]
+      if (value) {
         throw new CmsException(
-          `Cannot overwrite field ${field.cmsName} of entry ${entry.sys.id} because ManageContext.allowOverwrites is enabled`
+          `Cannot overwrite field $'{field.cmsName}' of entry '${entry.sys.id}' (has value '${value}) because ManageContext.allowOverwrites is false`
         )
       }
     }
--- a/packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
@@ -78,7 +78,7 @@ export class ManageContentful implements ManageCms {
       const value = entry.fields[field.cmsName][context.locale]
       if (value) {
         throw new CmsException(
-          `Cannot overwrite field $'{field.cmsName}' of entry '${entry.sys.id}' (has value '${value}) because ManageContext.allowOverwrites is false`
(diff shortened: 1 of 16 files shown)


packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
```diff
@@ -75,9 +75,10 @@ export class ManageContentful implements ManageCms {
       )
     }
     if (!context.allowOverwrites) {
-      if (entry.fields[field.cmsName][context.locale]) {
+      const value = entry.fields[field.cmsName][context.locale]
+      if (value) {
         throw new CmsException(
-          `Cannot overwrite field ${field.cmsName} of entry ${entry.sys.id} because ManageContext.allowOverwrites is enabled`
+          `Cannot overwrite field $'{field.cmsName}' of entry '${entry.sys.id}' (has value '${value}) because ManageContext.allowOverwrites is false`
```
marcpdw: Shouldn't the quote be around the dollar sign here? `$'{field.cmsName}' `

packages/botonic-plugin-contentful/src/util/backoff.ts
```diff
@@ -0,0 +1,34 @@
...
+    this.times++
+  }
+}
+
+export async function repeatWithBackoff<T>(
+  func: () => Promise<T>,
+  backoff = new ExponentialBackoff(),
+  logger: (msg: string) => void = console.log
+): Promise<T> {
+  for (;;) {
```
marcpdw: Why do you need the `for` here if you're returning an `await`?

packages/botonic-plugin-contentful/tests/tools/translators/csv-import.test.ts
```diff
@@ -21,116 +22,117 @@ function ctxt(ctx: Partial<ManageContext>): ManageContext {
...
-      Field: ContentFieldType.TEXT,
-      from: 'from',
-      to: 'válor2',
-    } as Record)
-  }, 10000)
+test('TEST: CsvImport read text & carousel', async () => {
+  const mockFieldImporter = mock<StringFieldImporter>()
+  const readLines: Record[] = []
+  when(mockFieldImporter.consume(anything())).thenCall((record: Record) => {
+    console.log(record.Code, record.Field, record.to)
```
marcpdw: Did you leave this here intentionally?

packages/botonic-plugin-contentful/src/util/backoff.ts
dpinol: in case it fails, we want to repeat. backof will be the responsible to break with exception if it times out

packages/botonic-plugin-contentful/tests/tools/translators/csv-import.test.ts
dpinol: nope. removed. thanks!

packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
dpinol: :+1: good eyes!

PR #652, opened 2020-03-24
feat(contentful): now buttons can refer queues or any other top content

dpinol:


packages/botonic-plugin-contentful/tests/contentful/contents/text.test.ts
```diff
@@ -1,19 +1,21 @@
 import { expectImgUrlIs } from './image.test'
 import { testContentful, testContext } from '../contentful.helper'
 import * as cms from '../../../src'
+import { ENGLISH } from '../../../src'
```
guinii: In the future, refactor it because we are import * from 'src', and then another constant from 'src' too. It looks rare!

PR #654, opened 2020-03-27
Feat/handoff improvements

vanbasten17:
* Being able to make the handoff via agent id.
* Fixed also handoff on finish not working in development mode (we were accessing the last element of an splitted string (which was coming from a stringified JSON, so it was taking the last part of the stringified JSON `"}`)
Usage:

```
   let handOffBuilder = new HandOffBuilder(session);
   handOffBuilder.withAgentId("AGENT_ID");
   handOffBuilder.withOnFinishPath("thanks-for-contacting");
   await handOffBuilder.handOff();
```

### DIFF
--- a/packages/botonic-core/src/handoff.js
+++ b/packages/botonic-core/src/handoff.js
@@ -39,6 +39,11 @@ export class HandOffBuilder {
     return this
   }
 
+  withAgentId(agentId) {
+    this._agent_id = agentId
+    return this
+  }
+
   withNote(note) {
     this._note = note
     return this
@@ -60,6 +65,7 @@ export class HandOffBuilder {
       this._queue,
       this._onFinish,
       this._email,
+      this._agent_id,
       this._caseInfo,
       this._note,
       this._shadowing
@@ -92,6 +98,7 @@ async function _humanHandOff(
   queueNameOrId = '',
   onFinish,
   agentEmail = '',
+  agentId = '',
   caseInfo = '',
   note = '',
   shadowing = false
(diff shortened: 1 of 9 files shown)


packages/botonic-react/src/utils.js
```diff
@@ -77,3 +77,10 @@ export const scrollToBottom = (timeout = 100) => {
     setTimeout(() => (frame.scrollTop = frame.scrollHeight), timeout)
   }
 }
+
+export const getParsedAction = botonicAction => {
+  if (botonicAction.split('create_case:').length > 1) {
```
dpinol: in this case imho creating a temporary variable for "botonicAction.split('create_case:'" has value due 1) dry
2) for debugging, it wil be easy to view the value of the splitted string

dpinol: early return practice says to returns the "bad path" asap, and end up with the "happy path" (ie. reverse if condition)

packages/botonic-react/src/utils.js
```diff
@@ -79,8 +79,7 @@ export const scrollToBottom = (timeout = 100) => {
 }
 
 export const getParsedAction = botonicAction => {
-  if (botonicAction.split('create_case:').length > 1) {
-    return JSON.parse(botonicAction.split('create_case:')[1])
-  }
-  return undefined
+  const splittedAction = botonicAction.split('create_case:')
+  if (!(splittedAction.length > 1)) return undefined
```
dpinol: if (splittedAction.length <= 1)

packages/botonic-core/src/handoff.js
```diff
@@ -60,6 +65,7 @@ export class HandOffBuilder {
       this._queue,
       this._onFinish,
       this._email,
+      this._agent_id,
```
dpinol: _agentId. we only use snakecase on the msg sent to backend

PR #664, opened 2020-03-30
Feat/more adaptable webchat

AinaVendrell:
New features:
- timestamps in messages
- style timestamps
- custom persistenMenu button
- custom persistentMenu 
- dark background when persistenMenu is open (true, false)
- hover on the triggerButton (color and opacity) 
- messages right padding (px)
- messages max-width (px)


packages/botonic-react/src/components/message.jsx
```diff
@@ -8,12 +8,14 @@ import { Button } from './button'
...
 import { WEBCHAT, COLORS } from '../constants'
 import Fade from 'react-reveal/Fade'
+import moment from 'moment'
 
 const MessageContainer = styled.div`
   display: flex;
   justify-content: ${props => (props.isfromuser ? 'flex-end' : 'flex-start')};
   position: relative;
   padding-left: 5px;
+  padding-right: ${props => props.userRightPadding};
```
vanbasten17: maybe it would be better to be able to redefine the container, not only the right padding

packages/botonic-react/src/constants.js
```diff
@@ -51,10 +51,15 @@ export const WEBCHAT = {
     botMessageBlobTick: 'message.bot.blobTick',
     botMessageStyle: 'message.bot.style',
     botMessageBackground: 'message.bot.style.background',
+    botBlobWidth: 'message.bot.blobWidth',
```
vanbasten17: `botMessageBlobWidth` to be more consistent

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -466,10 +475,33 @@ export const Webchat = forwardRef((props, ref) => {
...
+        'triggerButton.hoverOpacity',
+        getThemeProperty('triggerButton.style.opacity', 1)
+      )
+    : getThemeProperty('triggerButton.style.opacity', 1)
+  const TriggerButtonHoverColor = hover
+    ? getThemeProperty(
+        'triggerButton.hoverColor',
+        getThemeProperty('triggerButton.style.background', COLORS.SOLID_WHITE)
+      )
+    : getThemeProperty('triggerButton.style.background', COLORS.SOLID_WHITE)
```
vanbasten17: have you tried to use the `CustomTriggerButton` here with on hover styles instead of adding properties to the "default" trigger button?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -466,10 +475,33 @@ export const Webchat = forwardRef((props, ref) => {
...
     <CustomTriggerButton />
   ) : (
-    <StyledTriggerButton style={{ ...triggerButtonStyle }}>
+    <StyledTriggerButton
+      onMouseEnter={() => setHover(true)}
+      onMouseLeave={() => setHover(false)}
+      style={{
+        ...triggerButtonStyle,
+        opacity: TriggerButtonHoverOpacity,
+        background: TriggerButtonHoverColor,
```
vanbasten17: Maybe these could be removed if you can make it work with `CustomTriggerButton`

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -528,9 +564,13 @@ export const Webchat = forwardRef((props, ref) => {
             />
           )}
           {persistentMenuOptions && (
-            <ConditionalAnimation>
-              <PersistentMenu onClick={handleMenu} />
-            </ConditionalAnimation>
+            <FeaturesWrapper>
```
vanbasten17: what's `FeaturesWrapper`, where is defined?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -665,10 +706,34 @@ export const Webchat = forwardRef((props, ref) => {
...
+                        backgroundColor: '#000000',
+                        position: 'absolute',
+                        width: '100%',
+                        height: '100%',
+                        opacity: 0.3,
+                        zIndex: 1,
+                        right: 0,
+                        bottom: 0,
+                      }}
+                    />
```
vanbasten17: please, convert it into a styled-component and try to use the colors defined in constants

packages/botonic-react/src/webchat/webchat.jsx
AinaVendrell: I have used the one that is already defined, just to center the menu button. If not, when there is a lot of text in the input container the menu button remains on the top whereas the other buttons are centered. Should I create a new div instead of using FeaturesWrapper?

dpinol:
@AinaVendrell you need to rebase due to the conflicts (which is something normal). please investigate what commands below due and execute them:
- git fetch
- git rebase origin/master
- solve the conflict in package.json from your ide
- git add package.json
- git rebase --continue
- git push -f

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -266,6 +277,15 @@ export const Webchat = forwardRef((props, ref) => {
     props.persistentMenu
   )
 
+  const CustomPersistentMenu = getThemeProperty(
+    'userInput.menu.custom',
+    undefined
+  )
+  const darkBackgroundMenu = getThemeProperty(
```
dpinol: should this start with capitals?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -665,10 +693,22 @@ export const Webchat = forwardRef((props, ref) => {
                 Object.keys(webchatState.replies).length > 0 &&
                 webchatReplies()}
               {persistentMenuIsOpened && (
-                <OpenedPersistentMenu
-                  onClick={closeMenu}
-                  options={persistentMenuOptions}
-                />
+                <div>
+                  {darkBackgroundMenu && (
```
dpinol: should it be darkBackgroundMenu?

packages/botonic-react/src/webchat/webchat.jsx
AinaVendrell: It's a variable that can be true or false, should I use capitals in this case?

packages/botonic-react/src/webchat/webchat.jsx
AinaVendrell: I don't understand the problem

packages/botonic-react/package.json
```diff
@@ -27,6 +27,7 @@
     "axios": "^0.19.2",
     "emoji-picker-react": "^2.1.1",
     "framer-motion": "^1.8.3",
+    "moment": "^2.24.0",
```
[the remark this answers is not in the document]
vanbasten17: yes, it's okay for me! 👍

packages/botonic-react/src/webchat/webchat.jsx
dpinol: the convention says that boolean names should start or contain "is" "has" (eg. darkBackgroundMenu). However, if in other parts of the code we don't use this convention for the names of the properties or the names of the variables, we should leave it as it is. @marcrabat what do you think?

packages/botonic-react/src/webchat/webchat.jsx
dpinol: see my previous comment

packages/botonic-react/src/components/message.jsx
```diff
@@ -72,13 +96,23 @@ export const Message = props => {
...
+  const getTimestampLocale = getThemeProperty(`message.timestamps.locale`, 'en')
+  moment.locale(getTimestampLocale)
+
+  const getTimestampFormat = getThemeProperty(
+    `message.timestamps.format`,
+    false
+  )
+
   if (isBrowser()) {
     useEffect(() => {
```
dpinol, line 109: I suggest moving the whole useEffect code to its own private function

dpinol:
@AinaVendrell I added the complexity-rule specially for webchat :-)
You'll need to split the code to reduce the complexity below 18. 
Please run "npm i -D" to ensure you have latest development dependencies and run "pre-commit install", as "git commit" shouldn't have let you make this commit.
See at https://github.com/hubtype/botonic/pull/664/checks?check_run_id=586645233
``` 

/home/runner/work/botonic/botonic/packages/botonic-react/src/webchat/webchat.jsx
##[error]  131:35  error  Arrow function has a complexity of 19. Maximum allowed is 18  complexity
```

dpinol:
@AinaVendrell please fix the complexity issue and I'll push the team so that this is reviewed.

AinaVendrell:
@dpinol I’ve been checking the complexity issue and I saw that the `useEffect` functions didn’t add complexity to de webchat function. 
The complexity was comming from the return of the `webchat` function with complexity 10, the constant `const webviewRequestContext` with complexity 4, the constant `const triggerButton` with complexity 2 and from an `if(isMobile)` with complexity 1, there were 2 other complexities that I haven’t found.
Finally, to solve the complexity issue I have replaced the `const triggerButton` with a function.

dpinol:
> @dpinol I’ve been checking the complexity issue and I saw that the `useEffect` functions didn’t add complexity to de webchat function.
> The complexity was comming from the return of the `webchat` function with complexity 10, the constant `const webviewRequestContext` with complexity 4, the constant `const triggerButton` with complexity 2 and from an `if(isMobile)` with complexity 1, there were 2 other complexities that I haven’t found.
> Finally, to solve the complexity issue I have replaced the `const triggerButton` with a function.

:100: thank you for the info!

PR #665, opened 2020-03-30
fix: fix 'covid-19' tokenization in Portuguese & fix pre-commit

dpinol:
1) node-nlp did not correctly tokenize "covid-19"
Upgrade it so that it has incorporated this patch https://github.com/axa-group/nlp.js/pull/419

2) pre-commit script was not aborting when "npm run lint" returned non 0


dpinol:
@guinii @mireiaespuga merging to avoid having too many open PRs on contentful

PR #667, opened 2020-04-01
fix: BotonicMsgConverter was not using replaceEmptyStringsWith option value

dpinol:
* also added unit test
* minor improvements in text builders

### DIFF
--- a/packages/botonic-plugin-contentful/src/render/botonic-converter.ts
+++ b/packages/botonic-plugin-contentful/src/render/botonic-converter.ts
@@ -149,6 +149,6 @@ export class BotonicMsgConverter {
     if (this.options.replaceEmptyStringsWith == undefined) {
       return str
     }
-    return str || ' '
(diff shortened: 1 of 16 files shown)


packages/botonic-plugin-contentful/src/render/botonic-converter.ts
```diff
@@ -149,6 +149,6 @@ export class BotonicMsgConverter {
     if (this.options.replaceEmptyStringsWith == undefined) {
       return str
     }
-    return str || ' '
+    return str || this.options.replaceEmptyStringsWith
```
vanbasten17, line 152: could be useful to use `??` here also?

dpinol: str will never be undefined, and when it's "" I want this.options.replaceEmptyStringsWith, so it wouldn't work

PR #668, opened 2020-04-01
Chore/downgrade compromise

vanbasten17:
* `compromise` has introduced some breaking changes in newest versions that break botonic-nlu. Working version should be `11.13.2` while we don't make the proper changes.
* Add ignore `compromise` in `.dependabot/config.yml`

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,7 +1,13 @@
+# When modifying this config.yml file,
+# please, validate format: https://dependabot.com/docs/config-file/validator/
 version: 1
 update_configs:
   # Keep package.json (& lockfiles) up to date as soon as
   # new versions are published to the npm registry
-  - package_manager: "javascript"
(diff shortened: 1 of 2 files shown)


.dependabot/config.yml
```diff
@@ -1,7 +1,13 @@
...
   # Keep package.json (& lockfiles) up to date as soon as
   # new versions are published to the npm registry
-  - package_manager: "javascript"
-    directory: "/"
-    update_schedule: "weekly"
+  # Ignore updates: https://dependabot.com/docs/config-file/#ignored_updates
+  - package_manager: 'javascript'
+    directory: '/'
+    update_schedule: 'weekly'
+    ignored_updates:
```
dpinol, line 11: having the ignore_updates here might disable the ones that dependabot stored on their website when we refected some dependabot PRs. However, I think it's better to have it here :+1:

PR #681, opened 2020-04-06
Chore/limit complexity

dpinol:
### Description
Limit cyclomatic complexity to 18 through eslint and split functions more complex than that. Core router function has been left at 35, as it needs a major refactor.

### Context
Most coding gurus (eg. [Martin Fowler](https://martinfowler.com/bliki/FunctionLength.html) & [Uncle Bob](https://dev.to/danialmalik/a-brief-guide-to-clean-code-functions-104h) ) agree that writing small and simple functions is key to write good code.

The standard way to measure functions complexity is through its cyclomatic complexity (see [wikipedia](https://en.wikipedia.org/w/index.php?title=Cyclomatic_complexity). According to  [here](https://www.aivosto.com/project/help/pm-complexity.html), 20 is considered to be "alarming" (algorithm creator originally proposed a limit of 10).

In the future we might limit function length.


packages/botonic-core/src/router.js
```diff
@@ -130,9 +131,8 @@ export class Router {
       if (pathParams.length > 1) {
         return pathParams[1]
       }
-    } catch (e) {
-      return undefined
-    }
+    } catch (e) {}
```
marcpdw, line 134: Shouldn't this `catch` say something? At least on a log?

MarcosCA: Yes, I agree

dpinol: @marcpdw @MarcosCA thanks, I did a modification so that it should never fail, added log and unit tests :-) merging to avoid so many branches, but as usual open to feedback

PR #684, opened 2020-04-07
fix: make sure there are no residual tmp files

vanbasten17:
I managed to reproduce the following error by creating manually a `tmp` file in my bot's project directory.

```shell
? Do you want to create a new Bot? Yes
? Bot name: borar65
✔ Building...
[    ] Creating bundle...Deploy Error { Error: EEXIST: file already exists, mkdir 'tmp'
    at Object.mkdirSync (fs.js:757:3)
    at Run.createBundle (/usr/local/lib/node_modules/@botonic/cli/lib/commands/deploy.js:222:12)
    at Run.deploy (/usr/local/lib/node_modules/@botonic/cli/lib/commands/deploy.js:302:24)
    at process._tickCallback (internal/process/next_tick.js:68:7) errno: -17, syscall: 'mkdir', code: 'EEXIST', path: 'tmp' }
[==  ] Creating bundle...^C
```

Taking a deeper look on it, it seems that sometimes we have the residual `tmp` directory if a deploy failed previously at certain point or we abort the command in the middle of the process. 
I managed to solve it being sure that we always check and delete this directory whichever it was the situation. Now seems that it's not stuck in this part of deploy process at least.

**Updated**:
I keep it testing more with the new fix and forcing the error to occur. Now sometimes the error is exactly the opposite, directory `tmp` doesn't exist. As I read, fs doesn't remove the content recursively. I tried to pass also the flag `recursive: true` in options (it only works under experimental mode in Node v12).  Now I'm using `rimraf` instead which seems to work better for these situations. [Ref](https://stackoverflow.com/questions/18052762/remove-directory-which-is-not-empty)

### DIFF
--- a/packages/botonic-cli/src/commands/deploy.ts
+++ b/packages/botonic-cli/src/commands/deploy.ts
@@ -58,7 +58,6 @@ Uploading...
   async run() {
     const { args, flags } = this.parse(Run)
     track('Deployed Botonic CLI')
-
     force = flags.force || false
     npmCommand = flags.command
     this.botName = flags.botName
@@ -283,6 +282,7 @@ Uploading...
       text: 'Creating bundle...',
       spinner: 'bouncingBar',
     }).start()
+    if (fs.existsSync('tmp')) fs.rmdirSync('tmp')
     fs.mkdirSync(join('tmp'))
     copySync('dist', join('tmp', 'dist'))
     const zipRes = await zip('tmp', join(BOTONIC_BUNDLE_FILE))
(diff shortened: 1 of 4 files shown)


packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -148,7 +147,7 @@ Uploading...
       }
       // Show the current bot in credentials at top of the list
       const first_id = this.botonicApiService.bot.id
-      bots.sort(function (x, y) {
+      bots.sort(function(x, y) {
```
dpinol, line 150: @marcrabat please ensure you have pre-commit installed and execute "npm i -D" to update to prettier 2

vanbasten17: fixed 👍

PR #686, opened 2020-04-08
chore(react): eslint plugin for react hooks

dpinol:
- Disabled the rule "hook is called conditionally" where the condition was constant. Despite https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level , when the condition is fixed I don't think there should be an issue.
- Plugin added to root because IDEs may require plugins to be in some folder as eslint

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4853,6 +4853,12 @@
         }
       }
     },
+    "eslint-plugin-react-hooks": {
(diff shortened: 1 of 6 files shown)


dpinol:
> Why is it necessary to disable the linter in those cases? What is it complaining about?

I extended PR description

PR #687, opened 2020-04-08
Contentful/Several error reporting improvements

dpinol:
See commit messages

### DIFF
--- a/packages/botonic-plugin-contentful/src/cms/contents.ts
+++ b/packages/botonic-plugin-contentful/src/cms/contents.ts
@@ -69,7 +69,7 @@ export abstract class TopContent extends Content {
     return new TopContentId(this.contentType, this.id)
   }
 
-  cloneWithFollowUp(newFollowUp: FollowUp): TopContent {
+  cloneWithFollowUp(newFollowUp: FollowUp): this {
     const clone = shallowClone(this)
     ;(clone as any).common = shallowClone(clone.common)
     ;(clone as any).common.followUp = newFollowUp
@@ -219,13 +219,13 @@ export class Text extends MessageContent {
     return Content.validateContents(this.buttons)
   }
 
-  cloneWithButtons(buttons: Button[]): Text {
+  cloneWithButtons(buttons: Button[]): this {
     const clone = shallowClone(this)
     ;(clone as any).buttons = buttons
-    return clone as Text
+    return clone
   }
 
-  cloneWithText(newText: string): Text {
+  cloneWithText(newText: string): this {
     const clone = shallowClone(this)
     ;(clone as any).text = newText
     return clone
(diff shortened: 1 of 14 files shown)


packages/botonic-plugin-contentful/src/cms/contents.ts
```diff
@@ -69,7 +69,7 @@ export abstract class TopContent extends Content {
     return new TopContentId(this.contentType, this.id)
   }
 
-  cloneWithFollowUp(newFollowUp: FollowUp): TopContent {
+  cloneWithFollowUp(newFollowUp: FollowUp): this {
```
marcpdw, line 72: Why wasn't the return type correct before? `TopContent` is the right thing here, no?

packages/botonic-plugin-contentful/src/contentful/contents/button.ts
```diff
@@ -32,7 +32,14 @@ export class ButtonDelivery {
   ): Promise<cms.Button> {
     // we could pass the entry to fromId to avoid fetching it again, but it makes
     // the code more complex when the reference is a button
-    return this.fromId(entry.sys.id, context)
+    try {
+      return await this.fromId(entry.sys.id, context)
```
marcpdw, line 36: If the method is defined as `async`, is it necessary to `await` for the result? Is it double wrapping it into Promises?

packages/botonic-plugin-contentful/src/contentful/traverser.ts
```diff
@@ -200,7 +200,7 @@ export class I18nEntryTraverser {
     if (
       !val.fields ||
       val.fields.payload ||
-      val.sys.contentType.sys.id == ButtonDelivery.BUTTON_CONTENT_TYPE
+      val.sys.contentType?.sys.id == ButtonDelivery.BUTTON_CONTENT_TYPE
```
marcpdw, line 203: I normally consider the `===` as good practise

packages/botonic-plugin-contentful/src/cms/contents.ts
dpinol: it's explained here https://www.typescriptlang.org/docs/handbook/advanced-types.html#polymorphic-this-types

packages/botonic-plugin-contentful/src/contentful/contents/button.ts
dpinol: I only do the await when I use the result or then when want to treat the exceptions with a try/catch (through I'm not 100% about this one)

ericmarcos: @dpinol is right here. `return await` makes sense when you want to catch the rejection and return other stuff (and it only makes sense in that situation).
See https://jakearchibald.com/2017/await-vs-return-vs-return-await/

PR #688, opened 2020-04-08
feat(core): get agent vacations ranges

vanbasten17:
**UPDATED: Renamed `ranges` to `vacation_ranges` to be more explicit**

* Added new method to read agent vacations ranges.

Usage:

```
import { getAgentVacationRanges } from "@botonic/core";

const res = await getAgentVacationRanges(session, {
        agentId: "SOME_AGENT_ID",
      });
for (let range of res.ranges){
....
}
```
Then in the bot we will receive the ranges with start_date, end_date in timestamp format (can be managed easily with Date or libraries like moment)

Screenshot 2020-04-08 at 17 52 37

### DIFF
--- a/packages/botonic-core/src/handoff.js
+++ b/packages/botonic-core/src/handoff.js
@@ -1,7 +1,9 @@
 import axios from 'axios'
 
+const HUBTYPE_API_URL = 'https://api.hubtype.com'
+
 export async function getOpenQueues(session) {
-  const baseUrl = session._hubtype_api || 'https://api.hubtype.com'
+  const baseUrl = session._hubtype_api || HUBTYPE_API_URL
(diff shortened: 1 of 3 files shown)


packages/botonic-core/src/handoff.js
```diff
@@ -172,6 +172,23 @@ export async function getAvailableAgents(session) {
   return resp.data
 }
 
+export async function getAgentVacationsRanges(
+  session,
+  { agentId, agentEmail }
+) {
+  const baseUrl = session._hubtype_api || 'https://api.hubtype.com'
```
marcpdw, line 179: nit: Ideally the URL string should be a constant or an env var

vanbasten17: yes, good catch, put it in a constant! 👍

PR #689, opened 2020-04-09
fix: fix numbering issues in prod build

vanbasten17:
Fix numbering issues in docz prod build. (Already deployed, **doesn't need revision**, put it separated the changes in this branch).
It's very likely we are moving to a better documentation tool like docusaurus nearly.

### DIFF
--- a/docs/4.templates/2.template-custom-webchat.mdx
+++ b/docs/4.templates/2.template-custom-webchat.mdx
@@ -1,7 +1,7 @@
 ---
 name: Custom Webchat
 route: /templates/template-custom-webchat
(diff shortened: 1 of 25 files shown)


docs/4.templates/2.template-custom-webchat.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 name: Custom Webchat
 route: /templates/template-custom-webchat
-menu: Templates
+menu: 6. Templates
```
dpinol, line 4: should we rename the "4.templates" path to "6.templates"?

same for others

PR #710, opened 2020-04-20
Contentful/ chore and util improvements 1

dpinol:
* new content cloners
* type contentful entry fields as undefineable as they should be
* ValueObject (Stringable & Equatable) implemented in Callback & ContentId
* followUp fields is now readonly to ensure immutability

### DIFF
--- a/packages/botonic-plugin-contentful/src/cms/contents.ts
+++ b/packages/botonic-plugin-contentful/src/cms/contents.ts
@@ -149,6 +149,12 @@ export class Button extends Content {
     }
     return undefined
   }
+
+  cloneWithText(newText: string): this {
+    const clone = shallowClone(this)
+    ;(clone as any).text = newText
+    return clone
+  }
 }
 
 export class StartUp extends MessageContent {
@@ -177,6 +183,12 @@ export class Carousel extends MessageContent {
   validate(): string | undefined {
     return Content.validateContents(this.elements)
   }
+
+  cloneWithElements(elements: Element[]): this {
+    const clone = shallowClone(this)
+    ;(clone as any).elements = elements
+    return clone
+  }
 }
 
 /** Part of a carousel */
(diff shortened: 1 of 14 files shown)


packages/botonic-plugin-contentful/src/cms/contents.ts
```diff
@@ -69,7 +69,7 @@ export abstract class TopContent extends Content {
     return new TopContentId(this.contentType, this.id)
   }
 
-  cloneWithFollowUp(newFollowUp: FollowUp): this {
+  cloneWithFollowUp(newFollowUp: FollowUp | undefined): this {
```
marcpdw, line 72: I think I'd go for `cloneWithFollowUp(newFollowUp?: FollowUp): this { ... }` instead

packages/botonic-plugin-contentful/src/contentful/contents/follow-up.ts
```diff
@@ -77,18 +77,23 @@ export class FollowUpDelivery {
...
             entry.fields.followup.sys.id,
             context
           )
-
-      common.followUp = await this.fromEntry(followUp, context)
-      return common
+      return (await this.fromEntry(followUp, context)) as cms.FollowUp
     }
-    return common
+    return undefined
```
marcpdw, line 97: Why do we prefer to return `undefined` rather than `null`?

packages/botonic-plugin-contentful/src/contentful/contents/startup.ts
```diff
@@ -41,15 +41,15 @@ export class StartUpDelivery extends DeliveryWithFollowUp {
     return new cms.StartUp(
       await this.getFollowUp().commonFields(entry, context),
       img,
-      fields.text,
+      fields.text ?? '',
```
marcpdw, line 44: I've never used the double question mark. When does it apply?

packages/botonic-plugin-contentful/src/contentful/contents/follow-up.ts
dpinol: I decided to use undefined everywhere rather than null at the beginning of the contentful project based on my ignorance on js. In the future I may decide to change the policy

PR #721, opened 2020-04-28
React/multichannel config

dpinol:
## Description
Allow customizing some visualization aspects of Multichannel render for whatsapp

## Context

A customer wants some content types with a,b,c bullet points

Combine usage of React Component props and contexts

## Testing

The pull request...

- [x] has unit tests


packages/botonic-plugin-contentful/src/cms/contents.ts
```diff
@@ -177,6 +177,12 @@ export class StartUp extends MessageContent {
     ;(clone as any).text = newText
     return clone
   }
+
+  cloneWithButtons(buttons: Button[]): this {
+    const clone = shallowClone(this)
+    ;(clone as any).buttons = buttons
```
marcpdw, line 183: the semicolon at the beginning?

vanbasten17: @marcpdw I think this is a "trick" for the compiler to do an inline assignment when using "whatever" `as any`.

dpinol: prettier/eslint inserts it automatically. 
I think it's necessary when we don't use final semicolons and a line starts with a bracket

PR #727, opened 2020-04-30
Fix/fix last messages updates

vanbasten17:

## Description
* Fixed a naming mismatch between expected class attribute and argument received.
* Unify names to avoid confusion.

## Context (Updated)
In init method we were setting `this.lastMessageUpdate` instead of `this.lastMessageUpdateDate`. Then, when we were calling it in `'X-BOTONIC-LAST-MESSAGE-UPDATE-DATE': this.lastMessageUpdate` it was undefined because of this naming mismatch. I have renamed all these occurrences to the same name to avoid conflicts.
We finally decided not to use this header as we have found a better solution at this moment, but it will be interesting to have this implemented as can be implemented in the near future.

Renamed properties

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests
- [X] doesn't need tests because... fixes a clear bug.

### DIFF
--- a/packages/botonic-core/src/hubtype-service.js
+++ b/packages/botonic-core/src/hubtype-service.js
@@ -14,10 +14,11 @@ export class HubtypeService {
     if (user.id && (lastMessageId || lastMessageUpdateDate)) this.init()
   }
 
-  init(user, lastMessageId, lastMessageUpdate) {
+  init(user, lastMessageId, lastMessageUpdateDate) {
     if (user) this.user = user
     if (lastMessageId) this.lastMessageId = lastMessageId
-    if (lastMessageUpdate) this.lastMessageUpdate = lastMessageUpdate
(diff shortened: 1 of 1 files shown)


packages/botonic-core/src/hubtype-service.js
```diff
@@ -27,7 +28,7 @@ export class HubtypeService {
         headers: {
           'X-BOTONIC-USER-ID': this.user.id,
```
dpinol, line 29: since I had a hard time finding these headers in backend:
 
I suggest:
*adding a comment here like
// See https://stackoverflow.com/a/22856867/145289 on why the headers use dashes here but underscore in backend
* asking Marcos why in backend he checked both "HTTP-X-BOTONIC" and "X-BOTONIC" https://github.com/metis-ai/hubtype-backend/blob/develop/organization/views/provider_account.py#L525

PR #740, opened 2020-05-04
Contentful/prefer token match

dpinol:
## Description
When finding a user input in the keywords of CMS contents, we were matching only the stems and not the tokens in order to be tolerant to changes in gender, number, verbs conjugations...
However, it's better to provide that matched substring from the full token. In this way, in case this keyword has to compete with another keyword, we want to compare the length of the matching substring.

## Context

Only taking the stems into account was specially problematic when eg. comparing a keyword from a FAQwith a keyword from a chitchat text. We want to give priority to the FAQ one, but not if the one from the chitchat has a much longer match.

## Testing

The pull request...

- [x ] has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/nlp/keywords.ts
+++ b/packages/botonic-plugin-contentful/src/nlp/keywords.ts
@@ -49,6 +49,10 @@ export class Keyword {
     }
     return this.words.map(w => new Keyword(w.token, [w], w.isStopWord))
   }
+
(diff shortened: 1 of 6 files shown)


dpinol:
> Looking good!! Just a question, is it right to have twice `joinedTokens` class method in two different classes?

maybe there's a cleaner refactor, but normally duplication of "composition" is ok, as long as you don't duplicate the algorithm (which is within Word class)

PR #742, opened 2020-05-05
Refactor/reduce-reducer-complexity

vanbasten17:

⚠️  _I suggest you going commit by commit._

## Description
* Extract action names into constants.
* Add new file messages-reducer with actions related with operating with messages.
* Add missing test for updateMessage dispatch.
* Add helper function in test-utils for rendering `useWebchat` hook.
* Set complexity allowed by linter to 18 again.

## Context
* As the number of operations in our webchat is growing, some tests were failing due to a large number of statements in the `webchat-reducer`. To solve this, I have refactored a little bit this part. The main in this PR was to pass the tests with complexity set to 18 again.

As we have talked with @dpinol, right now an easy alternative is to have the `messagesReducer` splitted inside the `default` statement for of the `switch-case`.
I have investigated a little bit and maybe we should consider to split the reducer in these reducers (my proposal):
* **MessagesReducer**: Operations with messages
* **DevWebchatReducer**: Variables needed for serve mode 
* **WebchatAppStateReducer**: Operations related with what is the current state of the application itself: showing/not showing typing indicator, showing/not showing webview...  

The syntax would be something like:
```
const [webchatReducer, webchatInitialState] = combineReducers({
    reducer1: [reducer1, { ...reducer1InitialState }],
    reducer2: [reducer2, { ...reducer2InitialState }],
  })
  const [webchatState, webchatDispatch] = useReducer(
    webchatReducer,
    webchatInitialState
  )
```

This library could help us to do this: https://www.npmjs.com/package/react-combine-reducers

## Testing

The pull request...

- [X] has unit tests
- [ ] has integration tests


packages/.eslintrc.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
     'prettier/prettier': 'error',
     'filenames/match-regex': ['error', '^[a-z-.]+$', true],
 
-    complexity: ['error', { max: 19 }],
+    complexity: ['error', { max: 18 }],
```
dpinol, line 26: please try reduce it as much as possible in case the reducer was our most complex function

packages/botonic-react/src/webchat/hooks.js
```diff
@@ -2,6 +2,25 @@ import { useEffect, useReducer, useRef, useState } from 'react'
 import { COLORS, WEBCHAT } from '../constants'
 import { scrollToBottom } from '../utils'
 import { webchatReducer } from './webchat-reducer'
+import {
```
dpinol, line 5: I suggest "import * as actions from ..."

dpinol: no, leave it so that many reduce functions are one-liners

dpinol:
:100: excellent work!

packages/.eslintrc.js
marcpdw: Silly question... what does this `max` value mean?

vanbasten17: As far as I know, this is the maximum number of "paths of execution" allowed by the linter within a block statement, so every statement (if, else if, else, switch/case) will contribute with 1 to this complexity counter.

vanbasten17: in our current reducer we had 18 cases + 1 default, so this was the reason which made the linter to complain.

PR #743, opened 2020-05-05
Contentful/search candidate

dpinol:
## Description
SearchResult's had a score field, but was not set by SearchByKeywords.  Now it's calculated checking the length of the matched substring.

## Context
The score is very important in order to compare several SearchResults 

Some fields from SearchResult moved to SearchCandidate so that both classes are actually immutable (before we were modifying a readonly field :cry:  ) 

## Testing

The pull request...

- [x] has unit tests


packages/botonic-plugin-contentful/src/search/search-by-keywords.ts
```diff
@@ -35,11 +38,8 @@ export class SearchByKeywords {
     )
     const results = kws.findCandidatesWithKeywordsAt(inputText)
     return results.map(res => {
-      const candidate = res.candidate as SearchResult
-      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-      // @ts-ignore
-      candidate.match = res.match
-      return candidate
+      const score = res.match.length / inputText.raw.length
```
marcpdw, line 41: Can `inputText.raw.length` be `0`?

dpinol: :+1: very good point. Actually I think we get strange results somewhere else if we get empty texts.I added a check

PR #744, opened 2020-05-06
fix: handle keyboard issues for Safari Iphones

vanbasten17:

## Description
Added a general solution to try to solve iPhone keyboard issues.
* Added function to detect if device is an iPhone.
* Added handlers to act when input is focused/blurred and do a visual hack to avoid iPhone "intelligent" keyboards to apply visual side effects.
* Refactored into `keyboard-resizer.js` and using `window.visualViewport` as the preferable way to retrieve the window height when keyboard is shown.
* Set overscroll-behavior to contain to have "independent" scrolling bars.  
Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior
* Try to limit scroll boundaries and prevent scrolling whole window.

## Context
Some people has reported having issues with virtual keyboard of iPhones with the webchat. 
I found this post https://blog.opendigerati.com/the-eccentric-ways-of-ios-safari-with-the-keyboard-b5aa3f34228d where it more specifically about this well-known issue.

* Adding a little timeout to calculate how much the keyboard fills the screen (by checking the offset given by window.innerHeight) and set a new webchat height based on that when input is focused.
Attached video with before/after behavior.
[iphoneKeyboardComparison.zip](https://github.com/hubtype/botonic/files/4586409/iphoneKeyboardComparison.zip)

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests
- [X] doesn't need tests because... it's a display/visual aspect

### DIFF
--- a/packages/botonic-react/src/utils.js
+++ b/packages/botonic-react/src/utils.js
@@ -68,13 +68,18 @@ export const _getThemeProperty = theme => (
 export const ConditionalWrapper = ({ condition, wrapper, children }) =>
   condition ? wrapper(children) : children
 
-export const scrollToBottom = (timeout = 100) => {
-  const botonicScrollableArea = document.getElementById(
+export const getScrollableArea = () => {
+  const botonicScrollableContent = document.getElementById(
     'botonic-scrollable-content'
   )
-  const frame =
-    botonicScrollableArea &&
-    botonicScrollableArea.querySelectorAll('.simplebar-content-wrapper')[0]
+  const scrollableArea =
+    botonicScrollableContent &&
+    botonicScrollableContent.querySelectorAll('.simplebar-content-wrapper')[0]
+  return scrollableArea
+}
+
+export const scrollToBottom = (timeout = 200) => {
+  const frame = getScrollableArea()
   if (frame) {
     frame.scrollTop = frame.scrollHeight
     setTimeout(() => (frame.scrollTop = frame.scrollHeight), timeout)
@@ -93,36 +98,9 @@ export function renderComponent({ renderBrowser, renderNode }) {
   throw new Error('Unexpected process type. Not recognized as browser nor node')
 }
 
-export const isIphone = () => {
-  // Detecting iOS solution, relying on navigator.platform: https://stackoverflow.com/a/9039885
-  return !!navigator.platform && /iPhone/.test(navigator.platform)
-}
+export const getWebchatElement = () =>
+  document.getElementById('botonic-webchat')
 
-const getWebchatElement = () => document.getElementById('botonic-webchat')
-
-const setWebchatElementHeight = newHeight => {
+export const setWebchatElementHeight = newHeight => {
   getWebchatElement().style.height = newHeight
-}
-
-export const handleIphoneOnFocus = () => {
-  /*
-    Based on Tip #4 from https://blog.opendigerati.com/the-eccentric-ways-of-ios-safari-with-the-keyboard-b5aa3f34228d,
-    taking window.innerHeight as the amount of pixels the virtual keyboard adds
-  */
-  const waitUntilKeyboardIsShown = 200
-  const calculateNewWebchatElementHeight = () => {
-    const keyboardOffset = window.innerHeight
-    const webchatHeight = getWebchatElement().clientHeight
-    let newWebchatPercentualHeight = keyboardOffset / webchatHeight
-    const toTwoDecimal = toRound => Number(Math.round(toRound + 'e2') + 'e-2')
-    newWebchatPercentualHeight = toTwoDecimal(newWebchatPercentualHeight) * 100
-    return newWebchatPercentualHeight
-  }
-  setTimeout(() => {
-    setWebchatElementHeight(`${calculateNewWebchatElementHeight()}%`)
-    scrollToBottom()
-  }, waitUntilKeyboardIsShown)
-}
-export const handleIphoneOnBlur = () => {
-  setWebchatElementHeight('100%')
 }
--- a/packages/botonic-react/src/utils.js
+++ b/packages/botonic-react/src/utils.js
@@ -92,3 +92,37 @@ export function renderComponent({ renderBrowser, renderNode }) {
   else if (isNode()) return renderNode()
   throw new Error('Unexpected process type. Not recognized as browser nor node')
 }
+
+export const isIphone = () => {
+  // Detecting iOS solution, relying on navigator.platform: https://stackoverflow.com/a/9039885
+  return !!navigator.platform && /iPhone/.test(navigator.platform)
+}
+
+const getWebchatElement = () => document.getElementById('botonic-webchat')
+
+const setWebchatElementHeight = newHeight => {
+  getWebchatElement().style.height = newHeight
+}
+
+export const handleIphoneOnFocus = () => {
+  /*
+    Based on Tip #4 from https://blog.opendigerati.com/the-eccentric-ways-of-ios-safari-with-the-keyboard-b5aa3f34228d,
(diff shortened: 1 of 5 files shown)


MarcosCA:
nice trick

packages/botonic-react/src/utils.js
```diff
@@ -92,3 +92,37 @@ export function renderComponent({ renderBrowser, renderNode }) {
   else if (isNode()) return renderNode()
   throw new Error('Unexpected process type. Not recognized as browser nor node')
 }
+
+export const isIphone = () => {
+  // Detecting iOS solution, relying on navigator.platform: https://stackoverflow.com/a/9039885
+  return !!navigator.platform && /iPhone/.test(navigator.platform)
```
dpinol, line 98: here it shouldn't matter, but remember that regex are probably slower than string.includes

packages/botonic-react/src/utils.js
```diff
@@ -92,3 +92,37 @@ export function renderComponent({ renderBrowser, renderNode }) {
...
+  /*
+    Based on Tip #4 from https://blog.opendigerati.com/the-eccentric-ways-of-ios-safari-with-the-keyboard-b5aa3f34228d,
+    taking window.innerHeight as the amount of pixels the virtual keyboard adds
+  */
+  const waitUntilKeyboardIsShown = 200
+  const calculateNewWebchatElementHeight = () => {
+    const keyboardOffset = window.innerHeight
+    const webchatHeight = getWebchatElement().clientHeight
+    let newWebchatPercentualHeight = keyboardOffset / webchatHeight
+    const toTwoDecimal = toRound => Number(Math.round(toRound + 'e2') + 'e-2')
```
dpinol, line 117: is there any reason for not doing Math.round(toRound*100)/100?
It's more standard and probably faster

packages/botonic-react/src/utils.js
```diff
@@ -92,3 +92,37 @@ export function renderComponent({ renderBrowser, renderNode }) {
...
+  return !!navigator.platform && /iPhone/.test(navigator.platform)
+}
+
+const getWebchatElement = () => document.getElementById('botonic-webchat')
+
+const setWebchatElementHeight = newHeight => {
+  getWebchatElement().style.height = newHeight
+}
+
+export const handleIphoneOnFocus = () => {
```
dpinol, line 107: we should NEVER have "utils.js". At least it should be called string_utils.js , file_utils.js...
I suggest moving all iphone functions to iphone_utils.js 

still, utils are also not so extensible, I think it's not too hard to have st like device_config.js
```
class DeviceConfig {
 private isIPhone() {
}
 onFocus() {...}
 onBlur() {...}
 fontsize() {...}
}

const deviceConfig = new DeviceConfig()
```
In the future we may have a subclass IphoneConfig, androidConfig ,....
In this way:
* we won't have to modify the usage of the functions, only the implementation
* we can make DeviceConfig grow instead of patching the TextArea
in summary, **think big, but be lazy** :-)

packages/botonic-react/src/utils.js
vanbasten17: yes, sure, I have changed this 👍

packages/botonic-react/src/utils.js
vanbasten17: 👍

packages/botonic-react/src/utils.js
vanbasten17: I have moved it into a `iphone-utils.hs` at this moment. I agree that we can solve this in lots of different better ways, but first I would like to check the solution is acceptable for our needs. Also, this is a very specific solution for iPhones/Safari browsers. If we can try it with @guinii and works as expected, I can invest some time trying to do it in the way you suggest 💪

dpinol: Approved, thanks for moving it! you can merge it :+1: 

"this is a very specific solution for iPhones/Safari browsers"
This is **precisely** the reason why you cannot mix a specific solution with the rest of the application or mix it with utils. Hacks and tunings should always be separated, because they tend to get more complex. If the 3 iphone utils need to go together, in the future they should be eg. in a HackForIphoneKeyboard class with a link to the post link you added in the PR

packages/botonic-react/src/device-config.js
```diff
@@ -0,0 +1,88 @@
...
+    return !!navigator.platform && navigator.platform.includes('iPhone')
+  }
+
+  onFocus() {
+    if (this.isIphone) {
+      /*
+        Based on Tip #4 from https://blog.opendigerati.com/the-eccentric-ways-of-ios-safari-with-the-keyboard-b5aa3f34228d,
+        taking window.innerHeight as the amount of pixels the virtual keyboard adds
+      */
+      const waitUntilKeyboardIsShown = 400
```
dpinol, line 24: :100: I love how the names of these const document what the function does

packages/botonic-react/src/device-config.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+    timerId = setTimeout(() => {
+      fn(...args)
+      timerId = null
+    }, delay)
+  }
+}
+
+const stopAtScrollLimit = element => {
+  if (element.scrollTop === 0) element.scrollTop = 1
```
dpinol, line 85: are 0,1 and -1 like flags, or do the mean actual pixels?

vanbasten17: are the actual pixels 👍

packages/botonic-react/src/device-config.js
```diff
@@ -0,0 +1,88 @@
+import {
+  getWebchatElement,
+  setWebchatElementHeight,
+  scrollToBottom,
+  getScrollableArea,
+} from './utils'
+
+export class DeviceConfig {
```
dpinol, line 8: if this does only keyboardresize, rename it accordingly, otherwise, split in different classes

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -598,8 +598,7 @@ export const Webchat = forwardRef((props, ref) => {
               onKeyDown={e => onKeyDown(e)}
               style={{
                 display: 'flex',
-                // Disabling auto-zoom on input (iPhone devices): https://stackoverflow.com/a/25614477
-                fontSize: isIphone() ? 'initial' : 14,
+                fontSize: keyboardResizer.fontSize(),
```
dpinol, line 601: I would pass the "14" as argument to fontSize, since 14 is a general config and not something particular to the keyboardResizer

packages/botonic-react/src/utils.js
```diff
@@ -92,3 +97,10 @@ export function renderComponent({ renderBrowser, renderNode }) {
   else if (isNode()) return renderNode()
   throw new Error('Unexpected process type. Not recognized as browser nor node')
 }
+
+export const getWebchatElement = () =>
+  document.getElementById('botonic-webchat')
```
ericmarcos, line 102: Beware that this will cause problems if somebody has 2 instances of our webchat in the same page. A solution could be to generate a unique id for each webchat instance.

I don't think this will be a common case, but I leave the note here to keep that in mind for the future. Maybe we can add a comment in the documentation.

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: yes, sure, would be better! added 👍

packages/botonic-react/src/utils.js
vanbasten17: You're right! Maybe we will need to do the same for the other "important" id `botonic-scrollable-content`. I'm adding a pair of todos there 👍

PR #754, opened 2020-05-11
feat(react): add borderRadius for persistentMenu

AinaVendrell:

## Description

Adds a bottom border radius for the persistent menu.

## Context
When the persistent menu is open, the bottom radius of the webchat is always zero even if the webchat radius is not.

Pass the webchat border radius to the persistent menu and assigns it to the  bottom radius of the close button.

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests
- [X] doesn't need tests because... is a style improvement.

### DIFF
--- a/packages/botonic-react/src/webchat/components/persistent-menu.jsx
+++ b/packages/botonic-react/src/webchat/components/persistent-menu.jsx
@@ -12,7 +12,7 @@ const ButtonsContainer = styled.div`
   text-align: center;
 `
 
-export const OpenedPersistentMenu = ({ onClick, options }) => {
+export const OpenedPersistentMenu = ({ onClick, options, borderRadius }) => {
   let closeLabel = 'Cancel'
   try {
     closeLabel = options.filter(opt => opt.closeLabel !== undefined)[0]
@@ -35,7 +35,9 @@ export const OpenedPersistentMenu = ({ onClick, options }) => {
           )
         )
       })}
-      <Button onClick={onClick}>{closeLabel}</Button>
+      <Button onClick={onClick} bottom={borderRadius}>
+        {closeLabel}
+      </Button>
     </ButtonsContainer>
   )
 }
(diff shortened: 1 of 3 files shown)


packages/botonic-react/src/webchat/components/persistent-menu.jsx
```diff
@@ -12,7 +12,7 @@ const ButtonsContainer = styled.div`
   text-align: center;
 `
 
-export const OpenedPersistentMenu = ({ onClick, options }) => {
+export const OpenedPersistentMenu = ({ onClick, options, borderRadius }) => {
```
AinaVendrell, line 15: should I change it to props?

AinaVendrell, line 15: should I change `{ onClick, options, borderRadius }` to `props`?

ericmarcos: I prefer the long form here, so it's clear what props are accepted by this component

marcpdw: In my opinion it's preferably this way. You avoid the first step of... `const { borderRadius } = this.props`.

packages/botonic-react/src/webchat/components/persistent-menu.jsx
```diff
@@ -35,7 +35,9 @@ export const OpenedPersistentMenu = ({ onClick, options }) => {
           )
         )
       })}
-      <Button onClick={onClick}>{closeLabel}</Button>
+      <Button onClick={onClick} bottom={borderRadius}>
```
marcpdw, line 38: Could you explain why a radius is applied to a position (`bottom`) property?

AinaVendrell: Inside the Button the bottom radius are defined as:
` border-bottom-right-radius: ${props => props.bottom || '0px'};`
` border-bottom-left-radius: ${props => props.bottom || '0px'};`
So I have used the same name, is it more clear if I use `bottomRadius` instead?

marcpdw: I think so, yes. I got confused because there's the `bottom` css property. I thought you were setting that position on the button.

PR #756, opened 2020-05-12
Fix/upload media crashes

vanbasten17:

# Pending
* Test all in staging env
* React tests will fail until we publish the core package (tests are passing for me locally)

## Description
* Add new event to be listened in pusher: `update_message_info`.
* Add method to update a message, given the new info (with the url)
* Add `message-utils` (first aprox., before moving these kind of things into a class)
* Refactored `sendInput` to be clearer (early returns and function to explicitly generate components to be added)
* Refactored constants with #755 const input types and provider.
* Add a limit in order to not allow media uploads greater than 10MB.
* Move attachment, setAttachments logic into reducer
* Added unit tests for some attachments related functions

## Context
We were having issues in Webchat when some files where uploaded. The fact was that we were storing the binary files for them in the local storage (provoking an overflow of the storage).

Listening to `update_message_info` event and updating a particular message with the new url. The media is displayed with a temporary generated URL in the same domain (see https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL), whereas the data is sent as form data via axios as it was done before.

## Testing

The pull request...

- [X] has unit tests
- [ ] has integration tests


packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -344,23 +344,28 @@ export const Webchat = forwardRef((props, ref) => {
           {input.data}
         </Text>
       )
+    } else {
+      const mediaProps = {
```
dpinol, line 352: complexity warning :-)
try to move to a function please

packages/botonic-react/src/message-utils.js
```diff
@@ -22,10 +22,19 @@ export const MEDIA_TYPES = [
...
 export const toBase64 = file =>
   new Promise((resolve, reject) => {
     const reader = new FileReader()
     reader.readAsDataURL(file)
     reader.onload = () => resolve(reader.result)
     reader.onerror = error => reject(error)
   })
+
+export const isAllowedSize = fileSize => {
+  const MAX_MB_SIZE_ALLOWED = 10
```
dpinol, line 38: in my personal convention I always put the units last (MAX_SIZE_ALLOWED_MB).
pedantic: word ALLOWED is not required here if you already have MAX in the name and you're within a isAllowedSize method

packages/botonic-react/tests/webchat/usewebchat.test.js
```diff
@@ -183,4 +183,14 @@ describe('TEST: useWebchat ', () => {
     })
     expect(result.current.webchatState.messagesJSON).toEqual([])
   })
+  it('setCurrentAttachment: sets current attachment to given object', () => {
```
dpinol, line 186: I prefer empty line between test methods

packages/botonic-react/src/message-utils.js
```diff
@@ -38,3 +40,14 @@ export const isAllowedSize = fileSize => {
   if (fileSize > maxAllowedBytesSize) return false
   return true
 }
+
+export const getAttachmentType = fileType => {
```
dpinol, line 44: let's try to be consistent with naming.
is getMediaType a better name?

packages/botonic-react/src/message-utils.js
```diff
@@ -38,3 +40,14 @@ export const isAllowedSize = fileSize => {
...
+  return Object.entries(MIME_WHITELIST)
+    .filter(([_, formatsForType]) => formatsForType.includes(fileType))
+    .map(([type, _]) => type)[0]
+}
+
+export const getAcceptedFormats = () =>
```
dpinol, line 50: st like  getFullMimeWhiteList() to be consistent with naming?

packages/botonic-react/tests/webchat/attachments.test.js
```diff
@@ -0,0 +1,58 @@
+import {
+  isAllowedSize,
+  getAttachmentType,
+  getAcceptedFormats,
+} from '../../src/message-utils'
+
+describe('TEST: attachments ', () => {
+  const createFile = ({ fakeData, fileName, mimeType, sizeInMB }) => {
```
dpinol, line 8: :100:

packages/botonic-react/tests/webchat/attachments.test.js
```diff
@@ -0,0 +1,58 @@
...
+      Object.defineProperty(file, 'size', {
+        value: sizeInMB * 1024 * 1024,
+        writable: false,
+      })
+    }
+    return file
+  }
+
+  it('Gets correctly the attachment type', () => {
+    const file = createFile({
```
dpinol, line 21: don't create the file if you don't use it

packages/botonic-react/tests/webchat/attachments.test.js
```diff
@@ -0,0 +1,58 @@
...
+  })
+
+  it('Accepts allowed files with size', () => {
+    const file = createFile({
+      fakeData: 'video_data',
+      fileName: 'video_filename.mp4',
+      mimeType: 'video/mp4',
+      sizeInMB: 10,
+    })
+    const sut = isAllowedSize(file.size)
```
dpinol, line 37: don't create the file if you don't use it

packages/botonic-react/src/message-utils.js
```diff
@@ -0,0 +1,53 @@
...
+  return true
+}
+
+export const getAttachmentType = fileType => {
+  return Object.entries(MIME_WHITELIST)
+    .filter(([_, formatsForType]) => formatsForType.includes(fileType))
+    .map(([type, _]) => type)[0]
+}
+
+export const getAcceptedFormats = () =>
```
dpinol, line 50: imho this method is very specific for a particular usage. it does not deserve to be in utils. 2 options:
a) convert it to return a single list of all the mimes so that the user only needs to do getFullWhiteList().join(",)
b) move it as private method to where it's used.

packages/botonic-react/tests/webchat/attachments.test.js
vanbasten17: my idea was to also cover the same way as the File will be called in a browser, is really that wrong to have it like this?

PR #759, opened 2020-05-14
Feat/blob tick styles

AinaVendrell:

## Description

- Add a border for blobTick that has the same color as the border message.
- The position of the blobTick can be styled
- The blockInputs don't have blobTick

## Context

- When using blockInputs the blobTick doesn't change its color. 
- The blobTick doesn't have a border so when the background of the message is white it looks weird. 
- The blobTick position can not be changed

- To add a border to the blobTick we need to create two triangles and overlap them that is why the color of the blobTick depends on the size
https://developpaper.com/realization-code-of-css-drawing-triangle-border-method/

- The border of the blobTick will just appear if the message border color is defined using `borderColor`, because there are many different ways to define a border using the property `border` and it’s difficult to extract the color from there.

- There is a blobTickContainer that can be styled to change the blobTick position.

## To documentate / Usage examples

- The border of the blobTick will just appear if the message border color is defined using `borderColor`.

- `blobTickStyle` can be used to set the position of the blobTick 

```
message: {
 bot: {
  style: {
   border: '1px solid black',
   borderColor: 'black',
  },
  blobTickStyle: {
   alignItems: 'flex-end',
  },
 },
},
```

- It is also useful to define a top or bottom padding inside the blobTickStyle to set a more exact position.

```
blobTickStyle: {
 alignItems: 'flex-end',
 paddingBottom: '30px',
},
```

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests
- [X] doesn't need tests because...is an style improvement


packages/botonic-react/src/components/message.jsx
```diff
@@ -154,31 +169,66 @@ export const Message = props => {
...
-      ? getThemeProperty('brand.color', COLORS.BOTONIC_BLUE)
-      : COLORS.SEASHELL_WHITE
+    if (isFromUser()) {
+      const color = getThemeProperty('message.user.style.background', undefined)
+      return color
+        ? color
+        : getThemeProperty('brand.color', COLORS.BOTONIC_BLUE)
+    }
+    const color = getThemeProperty('message.bot.style.background', undefined)
+    return color ? color : COLORS.SEASHELL_WHITE
```
marcpdw, line 182: To simplify this return and the one a few lines above, you could go for 
```
return color || COLORS.SEASHELL_WHITE
```

packages/botonic-react/src/components/message.jsx
```diff
@@ -217,45 +267,15 @@ export const Message = props => {
...
-                    getBgColor()
-                  ),
-                  left: 0,
-                  borderLeft: 0,
-                  marginLeft: -pointerSize + 1,
-                }}
-              />
-            )}
+            {blob && hasBlobTick() && getBlobTick(6)}
+            {blob && hasBlobTick() && getBlobTick(5)}
```
marcpdw, line 278: If `hasBlobTick()` returns a `boolean`, you did the right thing here with the refactor of the method 👍

packages/botonic-react/src/components/message.jsx
```diff
@@ -154,31 +169,66 @@ export const Message = props => {
...
+      }
+      const containerBotStyle = {
+        ...getThemeProperty('message.bot.blobTickStyle'),
+        left: 0,
+        marginLeft: -pointerSize,
+      }
+      const blobTickBotStyle = {
+        borderRightColor: color,
+        borderLeft: 0,
+      }
```
marcpdw, line 221: It looks good the way you put it together. You defined first the `user` styles and then the `bot` styles. There's a tiny improvement if you want, not a must.
You have `4` variables here and only `2` of them will be used at the same time. Probably I'd go for

```
const containerStyle = {
  ...getThemeProperty('message.user.blobTickStyle'),
}
const blobTickStyle = {}

if (isFromUser()) {
  containerStyle.left: 0
  containerStyle.marginLeft: -pointerSize
  blobTickStyle:borderLeftColor: color
} else {
  containerStyle.right: 0
  containerStyle.marginRight: -pointerSize
  blobTickStyle:borderRightColor: color
}

... 

return (
  <BlobTickContainer
    style={containerStyle}
  >
    <BlobTick
      pointerSize={pointerSize}
      style={blobTickStyle}
    />
  </BlobTickContainer>
)
```

Again, up to you!

vanbasten17: this would be a nice improvement! good catch 👍

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -313,9 +313,12 @@ export const Webchat = forwardRef((props, ref) => {
           <Text
             id={input.id}
             from='user'
+            blob={false}
             style={{
               backgroundColor: COLORS.SCORPION_GRAY,
               borderColor: COLORS.SCORPION_GRAY,
+              padding: '10px 15px',
+              maxWidth: '54%',
```
vanbasten17, line 321: why are these two new properties needed?

AinaVendrell: Because the message is no longer inside the blob, I could create a new prop inside the message to remove just the blobTick, but since blockInputs are not used very much I thought this could work.

packages/botonic-react/src/constants.js
```diff
@@ -49,12 +49,16 @@ export const WEBCHAT = {
...
+    botMessageBlobTickSize: 'message.bot.blobTickStyle',
     botMessageStyle: 'message.bot.style',
     botMessageBackground: 'message.bot.style.background',
+    botMessageBorderColor: 'message.bot.style.borderColor',
     botMessageBlobWidth: 'message.bot.blobWidth',
     userMessageStyle: 'message.user.style',
     userMessageBlobTick: 'message.user.blobTick',
+    userMessageBlobTickSize: 'message.user.blobTickStyle',
     userMessageBackground: 'message.user.style.background',
+    userMessageBorderColor: 'message.user.style.borderColor',
```
vanbasten17, line 61: `botMessageBlobTickSize` and `userMessageBlobTickSize` are not consistent with the naming here. Can you please rename it as `botMessageBlobTickStyle` and  `userMessageBlobTickStyle`

packages/botonic-react/src/components/message.jsx
```diff
@@ -154,31 +169,56 @@ export const Message = props => {
...
     if (!blob) return COLORS.TRANSPARENT
-    return isFromUser()
-      ? getThemeProperty('brand.color', COLORS.BOTONIC_BLUE)
-      : COLORS.SEASHELL_WHITE
+    if (isFromUser()) {
+      const color = getThemeProperty('message.user.style.background', undefined)
+      return color || getThemeProperty('brand.color', COLORS.BOTONIC_BLUE)
+    }
+    const color = getThemeProperty('message.bot.style.background', undefined)
+    return color || COLORS.SEASHELL_WHITE
```
vanbasten17, line 180: suggestion: you can use the second parameter of getThemeProperty as a fallback to the default properties. If the property is not found in the theme, then the second parameter will be used.

packages/botonic-react/src/components/message.jsx
```diff
@@ -154,31 +169,56 @@ export const Message = props => {
...
-      marginTop: -pointerSize,
+
+    const getBlobTick = pointerSize => {
+      const color =
+        pointerSize == 5
+          ? getBgColor()
+          : getThemeProperty(
+              `message.${from}.style.borderColor`,
+              COLORS.TRANSPARENT
+            )
```
vanbasten17, line 201: This logic it's confusing to me. It's not easy to deduce how this works (a color which depends on the `pointerSize` sounds strange). I suggest you to make this part more readable or at least adding a comment.

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: Consider to put here the same styles as the other messages (the ones from user/bot, I think it's `8px 12px`) in order to reduce the different variations of blobs (will be easy to maintain for us or to be refactored in a unified component in a future :) ). I think the width should not be necessary as if this component is flex, it should resize automatically depending on its inner content.

PR #767, opened 2020-05-20
fix(react): handle parent page scrolling issues on iOS

vanbasten17:

## Description
* Handling specific situations depending on the device --> DeviceAdapter
* KeyboardResizer renamed to WebchatResizer (only used for iOS)
* ScrollbarController --> It enables/disable scroll/touch events depending if webchat is scrollable or not, so if we scroll inside webchat when we have no messages, the parent page is not scrolled.

## Context
Webchat in iOS was causing a non-desired scroll in the parent page when trying to scroll in the messages list.

* Detecting when webchat is scrollable: 
`scrollableArea.visible.clientHeight - scrollableArea.full.clientHeight < 0`
* If webchat is scrollable, then enable mousewheel / touchmove (mobile) events in Webchat. Disable them otherwise. (`onmousewheel` and `onmousetouch` events).
* Added visual improvement for iOS Webchat.

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests
- [X] doesn't need tests because... UX/Visual improvements


packages/botonic-react/src/webchat/devices/scrollbar-controller.js
```diff
@@ -0,0 +1,103 @@
...
+  return function (...args) {
+    if (timerId) {
+      clearTimeout(timerId)
+    }
+    timerId = setTimeout(() => {
+      fn(...args)
+      timerId = null
+    }, delay)
+  }
+}
```
marcpdw, line 15: Do you include `lodash` or `underscore` on botonic? They should include this feature, but I might be wrong. If not, that's okay

vanbasten17: We don't use lodash neither underscore right now in Botonic, but maybe we should consider it! What do you think about?

vanbasten17:
yes, I agree it's cleaner with early returns @marcpdw, refactored 👍

PR #770, opened 2020-05-22
pre-release-fixes

vanbasten17:

I will add in this PR fixes I found before 0.12.0 release

## Description
* `markdown` prop should not be stored for all messages (only text messages).
* Set the webchat scrollbar to bottom when webchat is open (behavior: 'auto' for "jumping" directly to position, behavior:'smooth' to do it smoothly. 
Removed the css property as with `scrollTo` can be done programatically (not supported natively by Safari, see https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollTo) if it is necessary, then a polyfill should be used: https://www.npmjs.com/package/scroll-behavior-polyfill
* Added `cleanOnceBeforeBuildPatterns: ['dist']` to avoid race conditions when removing `dist` folder.
* Sending payload text button by default.

## Context
In this PR will try to fix little things I see during testings the templates and that have an affordable solution before the release.

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-react/src/components/message.jsx
+++ b/packages/botonic-react/src/components/message.jsx
@@ -95,9 +95,9 @@ export const Message = props => {
     style,
     imageStyle,
     timestamps = true,
-    markdown = true,
     ...otherProps
   } = props
+  const markdown = props.markdown
   const {
     webchatState,
     addMessage,
--- a/packages/botonic-cli/templates/blank/webpack.config.js
+++ b/packages/botonic-cli/templates/blank/webpack.config.js
@@ -228,7 +228,7 @@ const botonicServerConfig = {
   resolve: resolveConfig,
   devtool: 'source-map',
   plugins: [
-    new CleanWebpackPlugin({}),
+    new CleanWebpackPlugin({ cleanOnceBeforeBuildPatterns: ['dist'] }),
     imageminPlugin,
     new webpack.EnvironmentPlugin({
       HUBTYPE_API_URL: null,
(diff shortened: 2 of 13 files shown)


packages/botonic-react/src/components/message.jsx
```diff
@@ -95,9 +95,9 @@ export const Message = props => {
     style,
     imageStyle,
     timestamps = true,
-    markdown = true,
     ...otherProps
   } = props
+  const markdown = props.markdown
```
dpinol, line 100: pardon my ignore of botonic-react, I don't understand why this change affects only texts

vanbasten17: Maybe the description now is clearer. The fact here is that from within the `Text` component, we create the message as this. Inside textProps we have {markdown:true} set by default:
```
<Message json={serialize(textProps)} {...textProps} type={INPUT.TEXT}>
        {textProps.children}
</Message>
```
When destructuring the message in Message component, we don't want to have this prop for all messages, but only for the Components that have this prop (in our case Text)

packages/botonic-cli/templates/blank/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "scripts": {
-    "build": "webpack --env.target=all --mode=production",
+    "build": "rimraf dist/; webpack --env.target=all --mode=production",
```
dpinol, line 3: why not rm -rf dist?

vanbasten17: is not cross-platform

PR #782, opened 2020-05-27
fix(react): parse only links for enduser inputs

vanbasten17:

## Description
* Extract function `withLinksTarget` to be used by markdown and link renderers.
* Configure a linksRenderer (note you can initialize MarkdownIt with the 'zero' preset and enable only the required features) to linkify enduser inputs
* Add unit tests
* Tested locally in "prod" webchat mode.

## Context
As discussed in the demo, Markdown should not be applied for enduser inputs except for URLs patterns (we still want the enduser URL inputs to be clickable by default).

Before rendering the message, check if this is from `user` and apply `renderLinks` to transforms raw url text into corresponding anchor tags if necessary.

Screenshot 2020-05-27 at 11 23 38

Doesn't need documentation, we only need to specify that the only thing that will be parsed from user inputs will be the links.

## Testing

The pull request...

- [X] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-react/src/components/markdown.js
+++ b/packages/botonic-react/src/components/markdown.js
@@ -6,29 +6,47 @@ export const ESCAPED_LINE_BREAK = '&lt;br&gt;'
 const ESCAPED_LINE_BREAK_REGEX = new RegExp(ESCAPED_LINE_BREAK, 'g')
 const isLineBreakElement = element => element.type === 'br'
 
+const withLinksTarget = (renderer, target = '_blank') => {
+  // Support opening links in new tabs: https://github.com/markdown-it/markdown-it/blob/master/docs/architecture.md#renderer
+  const newRenderer =
+    renderer.renderer.rules.link_open ||
+    function (tokens, idx, options, env, self) {
+      return self.renderToken(tokens, idx, options)
+    }
+  renderer.renderer.rules.link_open = function (
+    tokens,
+    idx,
+    options,
+    env,
+    self
+  ) {
+    const aIndex = tokens[idx].attrIndex('target')
+    if (aIndex < 0) tokens[idx].attrPush(['target', target])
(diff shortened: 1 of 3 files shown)


packages/botonic-react/src/components/markdown.js
```diff
@@ -6,29 +6,47 @@ export const ESCAPED_LINE_BREAK = '&lt;br&gt;'
...
-  const md = new MarkdownIt({
-    html: true,
-    linkify: true,
-    typographer: true,
-  })
+const withLinksTarget = (renderer, target = '_blank') => {
   // Support opening links in new tabs: https://github.com/markdown-it/markdown-it/blob/master/docs/architecture.md#renderer
-  const renderer =
-    md.renderer.rules.link_open ||
+  const newRenderer =
```
dpinol, line 11: is there a better name than newRenderer? "new" is not so expressive. 
does mdRenderer or withoutMdRenderer make sense?

vanbasten17: this is only in the local scope of `withLinksTarget` to differentiate it from the one received in the first argument. I think that the diff lines make it confusing maybe.
```
const withLinksTarget = (renderer, target = '_blank') => {
  // Support opening links in new tabs: https://github.com/markdown-it/markdown-it/blob/master/docs/architecture.md#renderer
  const newRenderer =
    renderer.renderer.rules.link_open ||
    function (tokens, idx, options, env, self) {
      return self.renderToken(tokens, idx, options)
    }
  renderer.renderer.rules.link_open = function (
    tokens,
    idx,
    options,
    env,
    self
  ) {
    const aIndex = tokens[idx].attrIndex('target')
    if (aIndex < 0) tokens[idx].attrPush(['target', target])
    else tokens[idx].attrs[aIndex][1] = target
    return newRenderer(tokens, idx, options, env, self)
  }
}
```

PR #785, opened 2020-05-29
feat(react): cover component

AinaVendrell:

## Description

- Cover component that can be defined using the `coverComponent` variable and that will appear in the first interaction with the bot

- The WebchatContext has the function `updateUser` that will update the user and also the session user

## Context
This component can be useful to get information from the users before they start using the chat

- The cover component will be fully designed by the developer using the `coverComponent` variable.

- It has a property to close the component named `closeComponent`

- I have created a function for the persistentMenu because there was a complexity issue in the webchat function.

If this component is defined, it will appear at the first interaction with the bot.
To use it, we have to define the variable `coverComponent` at the webchat `index` as:
```
export const webchat = {
  coverComponent: CustomCover,
}
```

Where `CustomCover` will be the component that the developer has created. To close this component there is a property called `closeComponent()`. Moreover, if we want to store the information of the user we can use the `updateUser()` function.

### Example: 
Considering that we have defined our component `CustomCover` that has a form to know the name and email of the user and has a function called `close`, we can define `close` as:
```
close() {
  this.context.updateUser({
    name: this.state.name,
    email: this.state.email,
  })
  this.context.sendText('START', 'PAYLOAD')
  this.props.closeComponent()
}
```

When the user sends the form:
- The information will be stored in the user information (`session.user` and `user`) 
- The component will close 
- A message from the user saying ‘START’ will appear and the action that has a `payload` called ‘PAYLOAD’ will be called

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests


packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -340,6 +345,40 @@ export const Webchat = forwardRef((props, ref) => {
     togglePersistentMenu(false)
   }
 
+  const persistentMenu = () => {
```
dpinol, line 348: :clap: much clearer this way

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -166,6 +170,7 @@ export const Webchat = forwardRef((props, ref) => {
   const { initialSession, initialDevSettings, onStateChange } = props
   const [botonicState, saveState, deleteState] = useLocalStorage('botonicState')
   const deviceAdapter = new DeviceAdapter()
+  const [coverComponentIsOpened, setCoverComponentIsOpened] = useState(false)
```
vanbasten17, line 173: I think it would be better to have this two methods in our reducers files, just like we did for persistentMenu or emojis

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -121,7 +122,10 @@ const DarkBackgroundMenu = styled.div`
   z-index: 1;
   right: 0;
   bottom: 0;
-  border-radius: ${props => props.borderRadius || '25px'};
+  border-radius: 10px;
+  position: absolute;
+  width: 100%;
+  height: 100%;
```
vanbasten17, line 128: make sure this looks good for the related previous implementations you did with this component 👁️

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -685,6 +730,18 @@ export const Webchat = forwardRef((props, ref) => {
     }
   }
 
+  const updateAllUserReferences = user => {
+    if (user.name) {
+      const name = `${user.name} - ${webchatState.user.name}`
```
vanbasten17, line 735: would be useful to add a commentary to know why it's done like this

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -743,32 +801,17 @@ export const Webchat = forwardRef((props, ref) => {
...
-                  ) : (
-                    <OpenedPersistentMenu
-                      onClick={closeMenu}
-                      options={persistentMenuOptions}
-                      borderRadius={
-                        webchatState.theme.style.borderRadius || '10px'
-                      }
-                    />
-                  )}
+                  {persistentMenu()}
```
vanbasten17, line 809: cool refactor here!! 😉 👏

PR #793, opened 2020-06-03
feat(core/react): dynamically resolve webchat visibility

vanbasten17:

## Description
* Modified webchat `render` logic to render it depending on visibility.

## Context
Coordinate with backend in order to determine whether webchat should be visible or not.

## Approach Taken

**API Description**:
Added `visibility` endpoint to determine if webchat should be visible or not.

Webchat visibility is determined by the following webchat settings:
* Checkbox is set to visible
* Requester url is in whitelisted domains
* Queue linked to webchat is open

Webchat will still be visible in the following cases:
* No settings defined (webchat which don't have these values set, previous versions)
* Whitelisted urls list is empty, `[]`

**Checks won't be done proactively.**

**Default behavior**: an api call to backend will be done to determine (depending on Hubtype Desk settings) if webchat should be shown or not.

**Flexible behavior**:

In generated **index.html**, we can add `visibility` set to `true` if we want to make it always visible (also we can add a function returning `true/false`):
```
Botonic.render(document.getElementById("root"), {
          appId: "MY APP ID",
          visibility: true,
});
```
```
Botonic.render(document.getElementById("root"), {
          appId: "MY APP ID",
          visibility: function (){
                 const result = // do your api calls or other stuff
                 return Boolean(result)
          },
});
```

We can also declare above logic in **src/webchat/index.js** with the same identifier `visibility`, just like the ones we have for `onInit`, `onOpen`, ...

**Webchat SDK API**
It can also be called with the the public api `Botonic.getVisibility()`, or from `app` references,  which will return a promise resolving to `true/false` in case webchat should be visible (it will check both backend response and `visibility` passed option).
## Testing
The pull request...

- [ ] has unit tests
- [ ] has integration tests
- [X] doesn't need tests because...

### DIFF
--- a/packages/botonic-core/src/hubtype-service.js
+++ b/packages/botonic-core/src/hubtype-service.js
@@ -86,4 +86,10 @@ export class HubtypeService {
       }
     )
   }
+
+  static async getWebchatVisibility({ appId }) {
+    return axios.get(
+      `${HUBTYPE_API_URL}/v1/provider_accounts/${appId}/get_webchat_visibility/`
+    )
+  }
 }
--- a/packages/botonic-core/src/hubtype-service.js
+++ b/packages/botonic-core/src/hubtype-service.js
@@ -89,7 +89,7 @@ export class HubtypeService {
 
   static async getWebchatVisibility({ appId }) {
     return axios.get(
-      `${HUBTYPE_API_URL}/v1/provider_accounts/${appId}/get_webchat_visibility/`
+      `${HUBTYPE_API_URL}/v1/provider_accounts/${appId}/visibility/`
     )
   }
 }
--- a/packages/botonic-react/src/webchat-app.jsx
+++ b/packages/botonic-react/src/webchat-app.jsx
@@ -131,6 +131,13 @@ export class WebchatApp {
     this.webchatRef.current.clearMessages()
   }
 
+  async getVisibility() {
+    return this.resolveWebchatVisibility({
+      appId: this.appId,
+      visibility: this.visibility,
+    })
+  }
+
   getLastMessageUpdate() {
     return this.webchatRef.current.getLastMessageUpdate()
   }
--- a/packages/botonic-react/src/webchat-app.jsx
+++ b/packages/botonic-react/src/webchat-app.jsx
@@ -201,8 +201,7 @@ export class WebchatApp {
       const { status } = await HubtypeService.getWebchatVisibility({
         appId,
       })
-      if (status === 200) return true
-      return false
+      return status === 200
     } catch (e) {
       return false
     }
--- a/packages/botonic-react/src/webchat-app.jsx
+++ b/packages/botonic-react/src/webchat-app.jsx
@@ -212,10 +212,9 @@ export class WebchatApp {
     return false
   }
 
-  render(dest, optionsAtRuntime = {}) {
-    ;(async () => {
-      const isVisible = await this.resolveWebchatVisibility(optionsAtRuntime)
-      if (isVisible) render(this.getComponent(optionsAtRuntime), dest)
-    })()
+  async render(dest, optionsAtRuntime = {}) {
+    const isVisible = await this.resolveWebchatVisibility(optionsAtRuntime)
+    if (isVisible) render(this.getComponent(optionsAtRuntime), dest)
+    return null
   }
 }
--- a/packages/botonic-react/src/webchat-app.jsx
+++ b/packages/botonic-react/src/webchat-app.jsx
@@ -19,6 +19,7 @@ export class WebchatApp {
     onClose,
     onMessage,
     appId,
+    visibility,
   }) {
     this.theme = theme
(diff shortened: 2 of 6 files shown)


packages/botonic-react/src/webchat-app.jsx
```diff
@@ -185,7 +189,33 @@ export class WebchatApp {
...
+    if (typeof visibility === 'function' && visibility()) return true
+    return false
+  }
+
   render(dest, optionsAtRuntime = {}) {
-    render(this.getComponent(optionsAtRuntime), dest)
+    ;(async () => {
+      const isVisible = await this.resolveWebchatVisibility(optionsAtRuntime)
+      if (isVisible) render(this.getComponent(optionsAtRuntime), dest)
+    })()
```
ericmarcos, line 219: Why can't we define the whole render function as `async`?

packages/botonic-core/src/hubtype-service.js
```diff
@@ -86,4 +86,10 @@ export class HubtypeService {
       }
     )
   }
+
+  static async getWebchatVisibility({ appId }) {
+    return axios.get(
+      `${HUBTYPE_API_URL}/v1/provider_accounts/${appId}/get_webchat_visibility/`
```
ericmarcos, line 92: No need to put the html verb in the path: `/get_webchat_visibility/` => `/visibility/`

vanbasten17: sure, will be cleaner, both applied 👍

packages/botonic-core/src/hubtype-service.js
```diff
@@ -86,4 +86,10 @@ export class HubtypeService {
       }
     )
   }
+
+  static async getWebchatVisibility({ appId }) {
+    return axios.get(
+      `${HUBTYPE_API_URL}/v1/provider_accounts/${appId}/visibility/`
```
dpinol, line 92: Since it requires a new API, I suggest documenting the backend PR on the PR description https://github.com/metis-ai/hubtype-backend/pull/257

packages/botonic-react/src/webchat-app.jsx
```diff
@@ -185,7 +189,31 @@ export class WebchatApp {
...
+      if (await this.isWebchatVisible({ appId })) return true
+    }
+    if (visibility === true) return true
+    if (typeof visibility === 'function' && visibility()) return true
+    return false
+  }
+
+  async render(dest, optionsAtRuntime = {}) {
+    const isVisible = await this.resolveWebchatVisibility(optionsAtRuntime)
+    if (isVisible) render(this.getComponent(optionsAtRuntime), dest)
```
dpinol, line 217: in gneeral I prefer returning an explicit null, not sure if it's a common react practice

packages/botonic-core/src/hubtype-service.js
vanbasten17: added 👍

vanbasten17:
@dpinol, right now checks won't be done in a proactive way, only when webchat is rendered, if this answers your first question. So for example, if the html page is opened and few moments later should not be visible, webchat will continue being visible

packages/botonic-react/src/webchat-app.jsx
```diff
@@ -185,7 +196,32 @@ export class WebchatApp {
...
+
+  async resolveWebchatVisibility(optionsAtRuntime) {
+    let { appId, visibility } = optionsAtRuntime
+    visibility = visibility || this.visibility
+    if (visibility === undefined) {
+      if (await this.isWebchatVisible({ appId })) return true
+    }
+    if (visibility === true) return true
+    if (typeof visibility === 'function' && visibility()) return true
+    return false
```
marcpdw, line 219: Take another look at this boolean conditions. Most of them can be simplified. Also, remember that you can always do `Boolean(value)`. Example: `Boolean(undefined)` is `false`

ericmarcos: I don't see how this can be simplified without affecting readability. `visibility` here can take several values:
* `undefined`: do the default behaviour (check visibility from Hubtype API). It has to be exactly `undefined`, not `false` or other falsy values.
* A function: in this case we need to call that function and set visibility according to the boolean return value of that.
* `true`: visibility on
* "anythinbg else": visibility off

packages/botonic-react/src/webchat-app.jsx
```diff
@@ -185,7 +196,32 @@ export class WebchatApp {
...
 
-  render(dest, optionsAtRuntime = {}) {
-    render(this.getComponent(optionsAtRuntime), dest)
+  async isWebchatVisible({ appId }) {
+    try {
+      const { status } = await HubtypeService.getWebchatVisibility({
+        appId,
+      })
+      if (status === 200) return true
+      return false
```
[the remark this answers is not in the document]
vanbasten17: absolutely, pure inertia of early returns logic 😅

PR #805, opened 2020-06-09
feat: working approach of dynamic webchat settings updates

vanbasten17:

## Description
* Created a new component `WebchatSettings` to update webchat properties dynamically (works for serve and prod versions)
* Make use of `lodash.merge` (it has 0 deps) for deep merging webchat theme properties.
* Make sure themeUpdates are applied on webchat reloads.
* Add missing `enableUserInput` prop, also for `WebchatSettings`.
* Added unit tests

## Context
This feature is necessary for some internal projects.

Examples of usage (in botonic actions):

Action 1 - Will change the webchat color to black and hide text input. Also `blockInputs` will be these ones from now on and attachments will be enabled.

```
import React, { useContext } from "react";
import { Text, Reply, WebchatSettings } from "@botonic/react";
import { RequestContext } from "@botonic/react";

  static contextType = RequestContext;
  render() {
    return (
      <>
        <Text>
          Please, type "start" to start the tutorial.
          <Reply payload="reply1">Reply 1</Reply>
          <Reply payload="reply2">Reply 2</Reply>
        </Text>
        <WebchatSettings
          theme={{
            brand: { color: "black" },
            userInput: { enable: { false } },
          }}
          blockInputs={[
            { match: [/ugly/i, /bastard/i], message: "don't say me this" },
          ]}
          enableAttachments={true}
        />
      </>
    );
  }
}
```

Action 2 - In the next action, we give a response to the previous action and we enable again the text input. Also we activated the emojiPicker and added a persistentMenu dynamically.
```
import React from "react";
import { Text, WebchatSettings } from "@botonic/react";

  render() {
    return (
      <>
        <Text>Ok dude 😉</Text>
        <WebchatSettings
          theme={{ brand: { color: "orange" }, userInput: { enable: true } }}
          enableEmojiPicker={true}
          persistentMenu={[
            { label: "option1", payload: "opt1" },
            { label: "option2", payload: "opt2" },
          ]}
        />
      </>
    );
  }
}
```
We could also have a simple action which is only responsible to change the color as `WebchatSettings` will render nothing but the visual changes:

```
  render() {
    return <WebchatSettings theme={{ brand: { color: "green" } }} />;
  }
}
```
## Testing

The pull request...

- [X] has unit tests
- [ ] has integration tests


packages/botonic-react/src/utils.js
```diff
@@ -126,3 +126,14 @@ export const getWebchatElement = () =>
...
+export const serializeRegexs = (_, value) => {
+  if (value instanceof RegExp) return value.toString()
+  return value
+}
+
+export const rehydrateRegex = regex => {
+  const fragments = regex.match(/\/(.*?)\/([a-z]*)?$/i)
+  const rehydrated = new RegExp(fragments[1], fragments[2] || '')
+  return rehydrated
+}
```
marcpdw, line 139: Good one! I like the simplicity and the naming. Since we can't use typescript here, I'd suggest... 
- `regex` -> `regexStr`
- You can improve the regex with this tiny change `/\/(.[^\/]+)\/([a-z]{1})?$/i`
   - At least 1 character on the first part
   - Exclude the `/` to properly have the first part grouped
   - The second part only needs `{1}` character

Test it, it should work!

vanbasten17: it seems to work good! i have applied both changes, thanks!

packages/botonic-core/src/index.d.ts
```diff
@@ -21,6 +21,10 @@ export const INPUT: {
   DOCUMENT: 'document'
   LOCATION: 'location'
   CONTACT: 'contact'
+  BUTTON_MESSAGE: 'buttonmessage'
+  CAROUSEL: 'carousel'
+  CUSTOM: 'custom'
+  WEBCHAT_SETTINGS: 'webchat_settings'
```
dpinol, line 27: why webchat_settings contains an underscore and buttonmessage doesn't?

packages/botonic-react/src/components/webchat-settings.jsx
```diff
@@ -0,0 +1,66 @@
...
+export const normalizeWebchatSettings = settings => {
+  const {
+    theme,
+    blockInputs,
+    persistentMenu,
+    enableEmojiPicker,
+    enableAttachments,
+  } = settings
+  if (!theme.userInput) theme.userInput = {}
+  if (persistentMenu !== undefined) {
```
dpinol, line 52: did you choose "!== undefined" because you want to activate the "if" when persistentMenu is null?

packages/botonic-react/src/utils.js
```diff
@@ -126,3 +126,15 @@ export const getWebchatElement = () =>
 export const setWebchatElementHeight = newHeight => {
   getWebchatElement().style.height = newHeight
 }
+
+export const serializeRegexs = (_, value) => {
+  if (value instanceof RegExp) return value.toString()
+  return value
+}
+
+export const rehydrateRegex = regexStr => {
```
dpinol, line 135: I'd prefer the name deserialize for 2 reasons:
1) to be consistent with "serializeRegexs"
2) [hydration](https://stackoverflow.com/a/6991192/145289) means adding complementary data from a database (imagine you serialize only the ids, the hydration means fetching the models from the db which have these ids)

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -319,7 +334,12 @@ export const Webchat = forwardRef((props, ref) => {
     )
     if (!Array.isArray(blockInputs)) return false
     for (const rule of blockInputs) {
-      if (rule.match.some(regex => regex.test(input.data))) {
+      if (
+        rule.match.some(regex => {
```
dpinol, line 338: let's try yo reduce complexity to the minimum of every single function. We sholdn't wait until they are much too complex, we want it always easy :-)
What about creating a getBlockInputs() function which deserializes the regex?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -319,7 +334,12 @@ export const Webchat = forwardRef((props, ref) => {
     )
```
dpinol, line 334: in line 330, I'd add the word serialized

>  if is a text we check if it is a serialized RE

packages/botonic-react/src/utils.js
```diff
@@ -126,3 +126,15 @@ export const getWebchatElement = () =>
...
+  /* eslint-disable no-useless-escape */
+  const fragments = regexStr.match(/\/(.[^\/]+)\/([a-z]{1})?$/i)
```
dpinol, line 137: the regex is difficult to read. if you googled it, can you add a link?

packages/botonic-react/src/utils.js
```diff
@@ -126,3 +126,15 @@ export const getWebchatElement = () =>
 export const setWebchatElementHeight = newHeight => {
   getWebchatElement().style.height = newHeight
 }
+
+export const serializeRegexs = (_, value) => {
```
dpinol, line 130: serializeRegexs => serializeRegex

Since this needs to be called from stringify, I suggest
1) either you add a comment "to be used as 2s argument of JSON.stringify"
2) or we could replace it with a function which already calls JSON.Srtingify passing it as the 2n argument

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -169,6 +175,23 @@ export const Webchat = forwardRef((props, ref) => {
   const { theme } = webchatState
   const { initialSession, initialDevSettings, onStateChange } = props
   const [botonicState, saveState, deleteState] = useLocalStorage('botonicState')
+  const saveWebchatState = webchatState => {
+    // Serialization of regexs: https://stackoverflow.com/questions/12075927/serialization-of-regexp
```
dpinol, line 179: imho this comment should be in serializeRegexs.
Maybe it's better to have a "function serializeWithRegexs(object): string" which does 
{
const serializeRegex= (_, value) {...}
return JSON.stringify(object,  serializeRegex)
}

packages/botonic-react/src/components/webchat-settings.jsx
vanbasten17: a really good observation! 💯  yes, this is done in order to allow nullish values. Let's say you want to disable the persistentMenu again, so you should pass `persistentMenu={nulll}` prop.

vanbasten17:
Thanks for the feedback @dpinol , it has been very useful, I have applied all the serialization/deserialization improvements you suggested

packages/botonic-core/src/index.d.ts
vanbasten17: renamed to be consistent with current implementations

PR #807, opened 2020-06-11
fix: only render links coming from text elements

vanbasten17:

## Description
* Call renderlinks only on text elements, bypass React elements.

## Context
When rendering a custom message with user props, it was treated as if it can only be a text. This was breaking a custom rate React component rendered from the user perspective.

### DIFF
--- a/packages/botonic-react/src/components/message.jsx
+++ b/packages/botonic-react/src/components/message.jsx
@@ -115,7 +115,10 @@ export const Message = props => {
   let textChildren = React.Children.toArray(children).filter(
     e => ![Button, Reply].includes(e.type)
   )
-  if (from === 'user') textChildren = textChildren.map(e => renderLinks(e))
+  if (from === 'user')
+    textChildren = textChildren.map(e =>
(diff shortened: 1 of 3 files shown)


packages/botonic-react/src/components/message.jsx
```diff
@@ -115,7 +115,11 @@ export const Message = props => {
   let textChildren = React.Children.toArray(children).filter(
     e => ![Button, Reply].includes(e.type)
   )
-  if (from === 'user') textChildren = textChildren.map(e => renderLinks(e))
+  if (from === 'user')
+    textChildren = textChildren.map(e => {
+      if (typeof e === 'string') return renderLinks(e)
+      return e
```
marcpdw, line 121: I'll be a bit picky and suggest a simpler syntax if you please:

`return (typeof e === 'string') ? renderLinks(e) : e`

vanbasten17: you are a truly "picky" blinder, but I agree that it's simpler, fixed! 💯

PR #823, opened 2020-06-17
Botonic Plugin Google Analytics

asastre:
## Description
New plugin to track user interaction or bot's behaviour to Google Analytics.

## Context
Tracking information to Google Analytics is very usual nowadays and for some companies there is the need to track the user interaction with the bot to Google Analytics.

This plugin provides a basic tracking with its method `track` and it has to be done manually, there is no default tracking through the `pre` or `post` methods.

See [Readme.md](https://github.com/hubtype/botonic/blob/8be257fd2b8b7b2701b5d420bd8d89a0c9dbb1cb/packages/botonic-plugin-google-analytics/README.md) file for documentation/examples.

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests
- [ ] doesn't need tests because... no tests needed


packages/botonic-plugin-google-analytics/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+The tracking needs to be done manually, enabling this plugin doesn't track any user interaction or bot's behaviour by default.
+You can do the tracking inside the `botonicInit` method, `render` method or even inside the [Webchat listeners](https://docs.botonic.io/concepts/webchat#webchat-listeners)
+
+```javascript
+static async botonicInit({ input, session, params, lastRoutePath, plugins }) {
+    const ga = plugins.google_analytics
+    
+    const eventName = 'botEvent'
+    const eventData = {
```
dpinol, line 36: naming consistency.
In the plugin the field is called eventFields

packages/botonic-plugin-google-analytics/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@botonic/plugin-google-analytics",
+  "version": "0.14.0",
+  "main": "src/index.js",
+  "scripts": {
+    "cloc": "../../scripts/qa/cloc-package.sh .",
+    "prepare": "node ../../preinstall.js",
+    "lint": "npm run lint_core -- --fix",
+    "lint_ci": "npm run lint_core",
```
dpinol, line 9: clone .github/workflows/botonic-plugin-dashbot-tests.yml to create a github actions CI check

dpinol: Add a new hook to .pre-commit-config.yaml so that lint is executed when we do git commit

packages/botonic-plugin-google-analytics/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.js'"
+  },
+  "dependencies": {
+    "@analytics/google-analytics": "^0.4.0",
```
dpinol, line 13: since we don't have integration tests, I would be conservatieve and use "~" for both dependencies

packages/botonic-plugin-google-analytics/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@botonic/plugin-google-analytics",
+  "version": "0.14.0",
```
dpinol, line 3: only use 0.14.0 if you're going to publish it after merge. Otherwise, 0.14.0-rc1

packages/botonic-plugin-google-analytics/README.md
```diff
@@ -0,0 +1,49 @@
...
+      category: 'chatbot',
```
dpinol, line 37: please document what eventData/eventFields can/should contain:
* only the fields in the example or any fields?
* only strings and numbers, also nested objects, also class instances?

packages/botonic-plugin-google-analytics/README.md
```diff
@@ -0,0 +1,49 @@
...
+      trackingId: 'YOUR_TRACKING_ID'
+    }
+  }
+]
+```
+
+## Use
+
+The tracking needs to be done manually, enabling this plugin doesn't track any user interaction or bot's behaviour by default.
+You can do the tracking inside the `botonicInit` method, `render` method or even inside the [Webchat listeners](https://docs.botonic.io/concepts/webchat#webchat-listeners)
```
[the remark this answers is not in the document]
vanbasten17: i'm not pretty sure about the usage inside the listeners itself (I've never used a plugin in this way), have you tried it @asastre ?

packages/botonic-plugin-google-analytics/README.md
vanbasten17: @dpinol is it possible to add an `index.d.ts` for this project with definitions, or it has to be strictly a ts project?

packages/botonic-plugin-google-analytics/package.json
vanbasten17: I think it will be published as it's needed for a project, so it's okay that it is `0.14.0`.

packages/botonic-plugin-google-analytics/README.md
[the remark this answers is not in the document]
asastre: @dpinol maybe is better to remove the `async` `await` from the `track` method because it doesn't make sense to stop the execution of the bot while doing the tracking and hence it could be called anywhere (`botonicInit`, `render` or webchat listeners)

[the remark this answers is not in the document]
asastre: @vanbasten17 I've tryed to call it inside the `onInit` method and it works. I'll use it here to track the impressions of the chatbot for example.

packages/botonic-plugin-google-analytics/README.md
asastre: @dpinol corrected

packages/botonic-plugin-google-analytics/package.json
asastre: @dpinol done

packages/botonic-plugin-google-analytics/README.md
asastre: @dpinol @vanbasten17 I enhanced the documentation and I added descriptions of the eventFields argument.

vanbasten17: looking nice! 👌

asastre:
@dpinol @vanbasten17 after doing some tests with the Google Analytics tracking I did a refactor of the `track` method to match correctly the fields of eventFields with the GA fields.

I've removed the `async`/`await` of `track` method because, as I said in a comment before, it doesn't make sense to stop the execution of the bot while doing the tracking.

I've also added more and better documentation of the `track` method.

packages/botonic-plugin-google-analytics/src/index.js
```diff
@@ -0,0 +1,40 @@
...
+export default class BotonicPluginGoogleAnalytics {
+  constructor({ botName, trackingId }) {
+    this.analytics = Analytics({
+      app: botName,
+      plugins: [
+        googleAnalytics({
+          trackingId: trackingId,
+        }),
+      ],
+    })
```
ericmarcos, line 13: You'll also need to call the `identify` method so that GA knows what user is sending the events. Something like
```
this.analytics.identify(session.user.id, {
  firstName: session.user.name,
  channel: session.user.provider,
  channel_id: session.user.provider_id
})
```
I'm not sure if you can set any params you want...
Also, I'm not sure if you can do this in the constructor or you'll nedd to do it in every `track` call.

ericmarcos:
> @dpinol @vanbasten17 after doing some tests with the Google Analytics tracking I did a refactor of the `track` method to match correctly the fields of eventFields with the GA fields.
> 
> I've removed the `async`/`await` of `track` method because, as I said in a comment before, it doesn't make sense to stop the execution of the bot while doing the tracking.
> 
> I've also added more and better documentation of the `track` method.

@asastre please leave the `async` and state in the docs that you must `await` this call. As we're running this in Lambda, the main "thread" can finish and the process terminated leaving async calls that are not awaited terminated before they end the execution.

packages/botonic-plugin-google-analytics/README.md
```diff
@@ -0,0 +1,51 @@
...
+2. Add it to the `src/plugins.js` file:
+
+**src/plugins.js**
+```javascript
+export const plugins = [
+  {
+    id: 'google_analytics',
+    resolve: require('@botonic/plugin-google-analytics'),
+    options: {
+      botName: 'YOUR_BOT_NAME',
```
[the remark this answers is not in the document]
asastre: @ericmarcos yes, I'll remove it :+1:

packages/botonic-plugin-google-analytics/README.md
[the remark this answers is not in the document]
dpinol: > @dpinol maybe is better to remove the `async` `await` from the `track` method because it doesn't make sense to stop the execution of the bot while doing the tracking and hence it could be called anywhere (`botonicInit`, `render` or webchat listeners)

You should not remove async, await. Or if you do, you should return the promise. Otherwise:
* you don't let the bot wait for the response in case they want to do it
* in massimo's bot where we track to dynamo, if you send the track when the lambda instance is restarted (it happens every few hours), then if you don't have the await, the track is never sent. I guess here would happen the same

packages/botonic-plugin-google-analytics/README.md
dpinol: @vanbasten17 good point, we have index.d.ts for core and react, right? Can you help Sastre we this?

asastre:
@dpinol @vanbasten17 @ericmarcos changes done. I restored the `async`/`await` in `track` method, as you said, in production the plugin will be executed in lambda and it needs to be sure it will execute.

dpinol:
@asastre review the prettier errors in CI checks. More important, your pre-commit should not have let you comit this file. Please verify that you did "pre-commit install" on your botonic checkout
/home/runner/work/botonic/botonic/packages/botonic-plugin-google-analytics/src/index.js
##[error]  62:35  error  Replace `this.userId({·session·}),·this.userTraits({·session·})` with `⏎······this.userId({·session·}),⏎······this.userTraits({·session·})⏎····`  prettier/prettier

packages/botonic-plugin-google-analytics/src/index.js
```diff
@@ -2,39 +2,64 @@ import Analytics from 'analytics'
 import googleAnalytics from '@analytics/google-analytics'
 
 export default class BotonicPluginGoogleAnalytics {
-  constructor({ botName, trackingId }) {
+  /**
+   * @param {Object} options - Options for the plugin
+   * @param {string} options.trackingId - Tracking ID for Google Analytics.
+   * @param {function({session: object}): string} [options.userId] - Method that returns a unique user ID as string.
```
dpinol, line 8: maybe it's more clear to call it getUserId & getUserTraits?

asastre:
@dpinol @ericmarcos I've made some changes/enhancements:

- Refactor name methods `defaultUserId` and `defaultUserTraits` to `getUserId` and `getUserTraits`.
- Refactor the option `trackManually` to `automaticTracking` to clarify its usage. If this option is not set, its default value will be `true`, enabling the automatic tracking on each user interaction.
- Added a new option: `eventFields` to enable to send custom event fields to the automatic tracking done by the plugin (only used if `automaticTracking` is not set or is set to `true`)

packages/botonic-plugin-google-analytics/README.md
asastre: @dpinol great! I'll have a look at it with @vanbasten17 in next PR

packages/botonic-plugin-google-analytics/src/index.js
```diff
@@ -10,8 +10,8 @@ export default class BotonicPluginGoogleAnalytics {
    * @param {boolean} [options.trackManually] - If set to true, no default tracking will be done in post method.
    */
   constructor(options) {
-    this.userId = options.userId ?? this.defaultUserId
-    this.userTraits = options.userTraits ?? this.defaultUserTraits
+    this.userId = options.userId ?? this.getUserId
```
dpinol, line 15: sorry I meant to rename options.userId to options.getUserId so that it's clear for the developer that it's a function. what do you think?
about the defaultUserId, defaultUserTraits functions, since the developers don't need to interact with them, their name is not so important. defaultGetUserId or getUserIdFromSession sound good to me

packages/botonic-plugin-google-analytics/src/index.js
```diff
@@ -0,0 +1,75 @@
...
+
+export default class BotonicPluginGoogleAnalytics {
+  /**
+   * @param {Object} options - Options for the plugin
+   * @param {string} options.trackingId - Tracking ID for Google Analytics.
+   * @param {function({session: object}): string} [options.userId] - Method that returns a unique user ID as string.
+   * @param {function({session: object}): object} [options.userTraits] - Method that returns the user traits as object.
+   * @param {boolean} [options.automaticTracking] - If set to false, no automatic tracking will be done in post method.
+   * @param {function({session: object, input: object, lastRoutePath: string}): object} [options.eventFields] - Method
+   *    that returns the eventFields to track as object (used only if automaticTracking is set to false).
```
dpinol, line 12: you meant "if automaticTracking is set to TRUE", right?

asastre: yes! Good eyes :eyes: ! Corrected.

packages/botonic-plugin-google-analytics/src/index.js
asastre: Oh! ok, now I understand and I do like the rename, it's clearer. Done! Thanks!

asastre:
> can you comment on why you removed the default getUserId?

I removed it because at first we were assuming that `clientId` and `userId` were the same for Google Analytics, but they are [different concepts](https://support.google.com/analytics/answer/6205850?hl=en#clientid-userid) and `userId` is meant to represent a logged user, so the plugin can't add one by default, the user must provide one if they want to track it.

asastre:
@dpinol @ericmarcos The plugin wasn't working properly when tracking on the server side loosing events randomly and never reaching Google Analytics.
After trying various tracking packages (`@analytics/google-analytics`, `universal-ga`) and even [proposing changes](https://github.com/DavidWells/analytics/pull/64) to one of them so the `clientId` parameter could be sent in the server side tracking (feature needed to keep the tracking flow for both client and server side), the tracking on the server side didn't work as expected, probably because it's executed in Lambda.
The new aproach (which is working fine) is to make direct requests to Google Analytics (using `axios`) sending the parameters needed each time.

packages/botonic-plugin-google-analytics/src/index.js
```diff
@@ -0,0 +1,76 @@
...
+      v: this.GA_API_VERSION,
+      tid: this.trackingId,
+      t: 'event',
+      ec: eventFields.category,
+      ea: eventFields.action,
+      el: eventFields.label,
+      ev: eventFields.value,
+    }
+    params.cid =
+      this.getGaClientId && this.getGaClientId({ session })
```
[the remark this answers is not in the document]
asastre: Totally agree, changed!

packages/botonic-plugin-google-analytics/src/index.js
```diff
@@ -0,0 +1,76 @@
...
+        ? (params.cid = this.getGaClientId({ session }))
+        : uuidv4()
+    if (this.getUserId && this.getUserId({ session }))
```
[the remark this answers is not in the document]
asastre: This case is similar but it's not the same because we only want to add the option `params.uid` if `getUserId` is set and returns something. 
What I've changed here is not to call 2 times `this.getUserId` as you well said in other comment.

PR #841, opened 2020-06-29
fix(react): refactor and remove moment dep

vanbasten17:

**Note**
In render section (return) the only relevant changes are: https://github.com/hubtype/botonic/pull/841/files#diff-a05365ee2d7479e601dd64cc443ef010R276-R282. The rest of changes within this return are due to wrapping it all inside a fragment `<></>`.

## Description
* Removed `moment` dependency. Managed default `webchat.botonic.js` size by 170KB.
* Refactored timestamps.
* Modify and adapt api for defining timestamps.

## Context
A needed action in order to keep reducing webchat bundle sizes.

Screenshot 2020-06-29 at 12 43 02

Timestamps are disabled by default.

You can enable them easily by:
* Setting `theme.message.timestamps.enable` to `true`. The format displayed by default for all locales will look like: `29 Jun, 12:40:07`. (formatted with [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString))
```
theme: {
    message: {
      timestamps: {
        enable: true,
      },
    },
  },
```
* Defining whether `style`, `format` or both properties:
```
theme: {
    message: {
      timestamps: {
        format: () => {
          // A function returning an string formatted date
          return new Date().toLocaleString();
        },
        style: {
          color: "blue",
        },
      },
    },
  },
```
## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests
- [X] doesn't need tests because... Removed dep and code reorganization

### DIFF
--- a/packages/botonic-react/src/components/message.jsx
+++ b/packages/botonic-react/src/components/message.jsx
@@ -2,7 +2,7 @@ import React, { useState, useEffect, useContext } from 'react'
 import styled from 'styled-components'
 import { v4 as uuidv4 } from 'uuid'
 
-import { isBrowser, isNode, INPUT } from '@botonic/core'
+import { isBrowser, INPUT } from '@botonic/core'
 import { resolveImage, ConditionalWrapper, renderComponent } from '../utils'
 
 import { WebchatContext, RequestContext } from '../contexts'
@@ -10,8 +10,8 @@ import { Button } from './button'
 import { Reply } from './reply'
 import { WEBCHAT, COLORS } from '../constants'
 import Fade from 'react-reveal/Fade'
-import moment from 'moment'
 import { renderMarkdown, getMarkdownStyle, renderLinks } from './markdown'
+import { resolveMessageTimestamps, MessageTimestamp } from './timestamps'
 
 const MessageContainer = styled.div`
   display: flex;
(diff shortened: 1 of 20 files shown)


packages/botonic-react/src/components/timestamps.jsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
```
dpinol, line 1: divide & conquer! :clap: :clap: :clap:

dpinol:
To documentate  => to document :-)

dpinol:
In the documentation it's not clear if you have to do both or only setting the flag to true
``` 
You can enable them easily by:
* setting...
* defining
``

**Comment by dpinol** [comment]
When you day:
`The format displayed by default will look like: 29 Jun, 12:40:07`
will this be the format for English locales or for all. If only for English, you could say:
`The format displayed by default for English locales will look like: 29 Jun, 12:40:07`

File: packages/botonic-react/src/components/message.jsx [hunk 2]
```diff
@@ -240,73 +217,70 @@ export const Message = props => {
...
+              ) : (
+                <BlobText blob={blob}>{textChildren}</BlobText>
+              )}
+              {buttons}
+              {blob && hasBlobTick() && getBlobTick(6)}
+              {blob && hasBlobTick() && getBlobTick(5)}
+            </Blob>
+          </MessageContainer>
+          {timestampsEnabled && (
+            <MessageTimestamp
```
dpinol, line 277: would it be possible to pass getThemeProperty to MessageTimestamp (or MessageTimestamp gets its from context) so that resolveMessageTimestamps is called from MessageTimestamp?
Then MessageTimestamp would be more autocontained and Message would be simpler

PR #845, opened 2020-06-30
fix(react): minor improvements

vanbasten17:

## Description
* Deep merging of `theme` properties at top-level in order to have properly merged the properties coming from `webchat/index.js` and the ones defined in `index.html`.
* Refactor `mobile` webchat configs and allow `mobileStyle` 
* Add message classNames to every message to be able to select them via css/scss.

## Context
Little improvements needed for some projects
* Using lodash.merge method in the same way we do it inside `webchat.jsx`
* Having classNames for messages.

## Testing

The pull request...

- [X] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-react/package.json
+++ b/packages/botonic-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/react",
-  "version": "0.13.0-alpha.1",
+  "version": "0.14.0-alpha.0",
   "description": "Build Chatbots using React",
   "main": "src/index.js",
   "scripts": {
--- a/packages/botonic-react/package.json
+++ b/packages/botonic-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/react",
-  "version": "0.13.0",
+  "version": "0.13.0-alpha.1",
   "description": "Build Chatbots using React",
   "main": "src/index.js",
   "scripts": {
--- a/packages/botonic-react/package.json
+++ b/packages/botonic-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/react",
-  "version": "0.14.0-alpha.0",
+  "version": "0.14.0-alpha.1",
   "description": "Build Chatbots using React",
   "main": "src/index.js",
   "scripts": {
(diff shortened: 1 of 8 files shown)


packages/botonic-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/react",
-  "version": "0.13.0",
+  "version": "0.13.0-alpha.1",
```
dpinol, line 3: this would mean going backwards

packages/botonic-react/src/webchat-app.jsx
```diff
@@ -175,7 +176,7 @@ export class WebchatApp {
       visibility,
       ...webchatOptions
     } = optionsAtRuntime
-    theme = { ...this.theme, ...theme }
+    theme = merge(this.theme, theme)
```
dpinol, line 179: it would be nice to have a unit test for this.
if you don't have time, you can create an empty one and mark it with test.skip so that it's highlighted

packages/botonic-react/package.json
vanbasten17: it was for testing purposes, will publish a `0.14.0-alpha.0`

dpinol: never look back :-D

vanbasten17: now updated :)

packages/botonic-react/src/webchat-app.jsx
vanbasten17: just added a test for this 👍

packages/botonic-react/tests/webchat.theme.test.js
```diff
@@ -94,3 +95,72 @@ describe('Updating webchat properties with settings ', () => {
...
+        },
+        header: { title: 'title' },
+        persistentMenu: [
+          { label: 'Start', payload: 'start' },
+          { label: 'Help', payload: 'help' },
+          { closeLabel: 'close' },
+        ],
+      },
+    }
+    const sut = merge(theme1, theme2)
```
dpinol, line 163: I see two typical issues in unit tests:
a) unit tests should test our code, not third party functions (which probably have their own unit tests). The only test where it makes sense, is when then 3rd party function is complex (and in this case it's better to wrap it and then test the wrapped function.
b) the unit test is very long and visually it's difficult to appreciate which fields are merged. 

I suggest to:
* leave this test reducing the theme to three fields tops
* adding an empty test.skip which  is called st like webchab_merges_theme_from_xx_with_theme_from_yy

PR #857, opened 2020-07-06
fix(react): disable text input on Cover Component shown

vanbasten17:

## Description
* Fix name in default initial state.
* Fix cover component: if cover component is shown, disable user input.

## Context
When cover component is shown on first interaction, user input should be disabled. This should be the expected behavior, as the purpose of the component is to retrieve information before the user starts interacting with webchat.

**Before:**
Screenshot 2020-07-06 at 16 12 06

**Now:**
Screenshot 2020-07-06 at 17 53 44

### DIFF
--- a/packages/botonic-react/src/webchat/webchat.jsx
+++ b/packages/botonic-react/src/webchat/webchat.jsx
@@ -418,12 +418,12 @@ export const Webchat = forwardRef((props, ref) => {
   }
 
   useEffect(() => {
-    if (CoverComponent)
-      if (
-        !botonicState ||
-        (botonicState.messages && botonicState.messages.length == 0)
-      )
-        toggleCoverComponent(true)
+    if (!CoverComponent) return
+    if (
+      !botonicState ||
+      (botonicState.messages && botonicState.messages.length == 0)
+    )
+      toggleCoverComponent(true)
   }, [])
 
   const coverComponent = () => {
@@ -511,6 +511,7 @@ export const Webchat = forwardRef((props, ref) => {
     getMessages: () => webchatState.messagesJSON,
     clearMessages: () => {
       clearMessages()
+      updateReplies(false)
     },
     getLastMessageUpdate: () => webchatState.lastMessageUpdate,
     updateMessageInfo: (msgId, messageInfo) => {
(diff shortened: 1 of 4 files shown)


packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -418,11 +418,12 @@ export const Webchat = forwardRef((props, ref) => {
   }
 
   useEffect(() => {
-    if (
-      !botonicState ||
-      (botonicState.messages && botonicState.messages.length == 0)
-    )
-      toggleCoverComponent(true)
+    if (CoverComponent)
```
dpinol, line 421: two nested ifs without {} is fragile.
Try "early return" to simplify

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -636,10 +637,17 @@ export const Webchat = forwardRef((props, ref) => {
...
-  const userInputEnabled = getThemeProperty(
-    'userInput.enable',
-    props.enableUserInput !== undefined ? props.enableUserInput : true
-  )
+  const isUserInputEnabled = () => {
+    let isUserInputEnabled = getThemeProperty(
+      'userInput.enable',
+      props.enableUserInput !== undefined ? props.enableUserInput : true
+    )
+    isUserInputEnabled =
```
dpinol, line 645: early return.
WHat do you think about..
```
if (!isUserInputEnabled) {
  return false
}
return !webchatState.isCoverComponentOpen
```

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -855,20 +878,13 @@ export const Webchat = forwardRef((props, ref) => {
...
-              {darkBackgroundMenu && (
-                <DarkBackgroundMenu
-                  style={{
-                    borderRadius: webchatState.theme.style.borderRadius,
-                  }}
-                />
-              )}
-              {persistentMenu()}
-            </div>
+            <DarkenBackground component={persistentMenu()} />
```
dpinol, line 881: :100:

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: I agree in both situations, now it's easier to read 👍

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -636,10 +638,16 @@ export const Webchat = forwardRef((props, ref) => {
...
-    'userInput.enable',
-    props.enableUserInput !== undefined ? props.enableUserInput : true
-  )
+  const isUserInputEnabled = () => {
+    const isUserInputEnabled = getThemeProperty(
+      'userInput.enable',
+      props.enableUserInput !== undefined ? props.enableUserInput : true
+    )
+    if (!isUserInputEnabled) return false
+    return !webchatState.isCoverComponentOpen
```
asastre, line 647: Could be simplified to:
```suggestion
    return isUserInputEnabled && !webchatState.isCoverComponentOpen
```

dpinol: @asastre @marcrabat depending on the case, the "early return" (see section at our [best practices](https://docs.google.com/document/d/1-qd2FdXbAe7XEbdTGyACeCs9HRCauCogocfEU-h-n-I/edit#heading=h.e072veoorbnb)) may be better.

otoh, Adria's version is shorter which is better. 

so it's a draw, which means that both versions are ok imho

PR #859, opened 2020-07-07
Contentful/multi environment contentful

dpinol:
Decorator for Contentful which can distribute the delivery requests amongst different spaces/environments.
The feature has been split into several pieces to make the code more modular and reusable.
See plugin.test.ts on how to use it


packages/botonic-plugin-contentful/src/contentful/cms-contentful.ts
```diff
@@ -0,0 +1,229 @@
...
+      this._carousel,
+      this._image,
+      this._startUp,
+    ].forEach(d => d.setFollowUp(followUp))
+    this._keywords = new KeywordsDelivery(delivery)
+    this._schedule = new ScheduleDelivery(delivery)
+    this._dateRange = new DateRangeDelivery(delivery)
+  }
+
+  button(id: string, context: Context = DEFAULT_CONTEXT): Promise<cms.Button> {
```
guinii, line 98: This function and `element, image, chitchat` should also be "async"?

packages/botonic-plugin-contentful/src/contentful/delivery-api.ts
```diff
@@ -200,3 +141,18 @@ export class ContentfulEntryUtils {
     })
   }
 }
+
+export function createContentfulClientApi(
```
guinii, line 145: Do we need here the "cache" variable? Or this new "Client" ignores the cache?

packages/botonic-plugin-contentful/src/contentful/cms-contentful.ts
dpinol: there's an eslint rule which complain if you add async when there's no await.
async is only required when there's an await

packages/botonic-plugin-contentful/src/contentful/delivery-api.ts
dpinol: this is the lowest level client to contentful, without the cache.

dpinol:
Merging with only 1 approve because other PRs depend on this one and this PR fixes a problem after changes in QA contentful contents

PR #878, opened 2020-07-20
fix(react): late update of user when using updateAllUserReferences

AinaVendrell:

## Description
Try to fix the late update of the session when using `updateAllUserReferences`

## Context

To save the user info when using the `coverComponent` we created a function called `updateAllUserReferences` that updates the session with the user information. The problem is that this function doesn’t update the session immediately so the user information is not available in the next action

It should look like:
Captura de Pantalla 2020-07-20 a les 18 20 24

But in the first action it looks like:
Captura de Pantalla 2020-07-20 a les 18 21 22

Try to use `Object.assign` instead of `updateUser` and `updateSession` but it doesn’t work in prod

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-react/src/webchat/webchat.jsx
+++ b/packages/botonic-react/src/webchat/webchat.jsx
@@ -784,11 +784,8 @@ export const Webchat = forwardRef((props, ref) => {
   }
 
   const updateAllUserReferences = user => {
-    updateUser({ ...webchatState.user, ...user })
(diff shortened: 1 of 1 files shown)


vanbasten17:
Closing. Solved with PR: https://github.com/hubtype/botonic/pull/973

PR #891, opened 2020-08-04
Contentful/import csv duplicate fields

dpinol:
Depends on #888 **Please review it first**

## Description

Tool to duplicate the value of reference fields into a new locale. Reference fields (assets or links to other contents) will typically be the same for all locales, but not always. So we initially link them all to the same target for all locales

## Context

To simplify the maintenance, sometimes we don't want the locales to fallback to other locales. In this case, the reference fields need to be set for each locale.  This needs to be done each time a space is translated to a new locale. So  import-csv-for-translators.sh  used to only set the new values for the translated text fields, but now will also duplicate the fields which have different values for each locale.
 
## Testing

The pull request needs integration tests :-(

### DIFF
--- a/packages/botonic-plugin-contentful/tests/tools/l10n/csv-import.test.ts
+++ b/packages/botonic-plugin-contentful/tests/tools/l10n/csv-import.test.ts
@@ -6,7 +6,7 @@ import {
 } from '../../../src/tools/l10n/csv-import'
 import { anything, instance, mock, when } from 'ts-mockito'
 import { testManageContentful } from '../../contentful/manage/manage-contentful.helper'
-import { SPANISH } from '../../../src/nlp'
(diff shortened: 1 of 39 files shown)


packages/botonic-plugin-contentful/tests/tools/l10n/csv-import.test.ts
```diff
@@ -63,13 +63,18 @@ test('TEST: StringFieldImporter test', async () => {
   class MockCms implements ManageCms {
     numCalls = 0
 
+    getDefaultLocale(): Promise<Locale> {
```
guinii, line 66: Promise<void> instead?

dpinol: what do you mean?

PR #892, opened 2020-08-06
Contentful/Locale improvements

dpinol:
## Description

1. Added support for locales which include the country (eg. es_ES). Now they are preprocessed as the root locale (without the country)

2. Added an option for Contentful plugin to be able to transform the name of the locale when sent to Contentful

## Context

* Sometimes we want to use locales which are not supported by the Contentful. Eg, English language for a non English speaking country. So what we do is using eg. en_DK, but converting to da_DK when calling Contentful APIs.

## Testing

Unit tests and integration tests added


dpinol:
> After this PR, all locales should be "es_ES" in order that everything works with "rootLocale" right?

nope, rootLocale takes the first 2 chars. It returns "es" for both "es" and "es-ES"

PR #894, opened 2020-08-07
fix(react): custom message types to support Reply

ericmarcos:
## Description

This PR fixes a crash that happened when using Custom Message Types together with Reply component.

Imagine you have this custom message:

```javascript
import React from 'react'
import { WebchatContext } from '@botonic/react'
import { customMessage } from './custom-message'

class MyMessage extends React.Component {
  static contextType = WebchatContext

  render() {
    return (
      <>
        <button
          onClick={() =>
            this.context.sendText(`Bye!`)
          }
        >Click me!</button>
        <p>{this.props.children}</p>
      </>
    )
  }
}

  name: 'my-message',
  component: MyMessage,
})
```

Now you can use it in your actions like this:

```javascript
import React from 'react'
import { Reply } from '@botonic/react'
import MyMessage from '../webchat/MyMessage'

export const CalendarAction = props => (
	<MyMessage>
		Hi
		<Reply payload='p'>Peperoni</Reply>
		<Reply payload='v'>Veggie</Reply>
	</MyMessage>
)
```

## Context

This problem was reported by some of our partners and users in our Slack channel

I used the same approach that `@botonic/react/src/components/services message.jsx`

See above

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-react/src/components/custom-message.jsx
+++ b/packages/botonic-react/src/components/custom-message.jsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { Message } from './message'
+import { Reply } from './reply'
 import { INPUT } from '@botonic/core'
 
 export const customMessage = ({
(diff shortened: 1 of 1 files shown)


dpinol:
@ericmarcos any comments on my feedback?

ericmarcos:
> @ericmarcos any comments on my feedback?

@dpinol what feedback?

dpinol (changes requested):
haha, I hadn't published my comments

asastre (approved):
Take a look at @dpinol's comments, the rest is ok!

vanbasten17 (approved):

PR #895, opened 2020-08-07
fix(react): visualize whatsapp messages in webchat

AinaVendrell:

## Description

Deactivate markdown in webchat when visualizing messages using whatsapp provider.

## Context
whatsapp messages that contain objects (buttons, carousels...) were displayed in webchat like:

Captura de Pantalla 2020-08-07 a les 15 16 58

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/nlp/stopwords/README.txt
+++ b/packages/botonic-plugin-contentful/src/nlp/stopwords/README.txt
@@ -1,0 +1,5 @@
+Default stopwords should not include the following:
+* Numbers
+* Words such as: When, Who, Where, Why
+* Yes, No
+* Single letters
--- a/packages/botonic-react/src/components/message.jsx
+++ b/packages/botonic-react/src/components/message.jsx
@@ -8,6 +8,7 @@ import {
   ConditionalWrapper,
   renderComponent,
   isDev,
+  isWebchatOrDev,
 } from '../utils'
 
 import { WebchatContext, RequestContext } from '../contexts'
@@ -97,6 +98,7 @@ export const Message = props => {
   const [state, setState] = useState({
     id: props.id || uuidv4(),
   })
+  const requestContext = useContext(RequestContext)
 
   const replies = React.Children.toArray(children).filter(e => e.type === Reply)
   const buttons = React.Children.toArray(children).filter(
@@ -279,7 +281,7 @@ export const Message = props => {
               }}
               {...otherProps}
             >
-              {markdown ? (
+              {markdown && (isWebchatOrDev(requestContext) || isFromUser) ? (
                 <BlobText
                   blob={blob}
                   dangerouslySetInnerHTML={{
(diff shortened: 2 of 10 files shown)


packages/botonic-react/src/components/message.jsx
```diff
@@ -279,7 +281,7 @@ export const Message = props => {
               }}
               {...otherProps}
             >
-              {markdown ? (
+              {markdown && !isWhatsapp(requestContext) ? (
```
dpinol, line 284: should this be disabled for all providers except dev & webchat?
if we can never go here with providers like telegram, add comment explaining why

guinii:
I think we are missing something here, in multichannel-text.jsx, we are setting `markdown=false` if the provider is whatsapp, so this code should not be triggered...I think it's the only case when we use markdown...

dpinol:
> I think we are missing something here, in multichannel-text.jsx, we are setting `markdown=false` if the provider is whatsapp, so this code should not be triggered...I think it's the only case when we use markdown...

@guinii very good point. I reverted the initial commit and suggested and implementation. 
@AinaVendrell If approved, please squash the PR so that we don't pollute the history with the initial implementation

packages/botonic-react/src/components/multichannel/multichannel.jsx
```diff
@@ -61,7 +61,11 @@ export const Multichannel = props => {
     newChildren = newChildren.map((c, index) =>
       index > 0 && typeof c === 'string' ? props.messageSeparator + c : c
     )
-    newChildren = <Text key={props.key}>{newChildren}</Text>
+    newChildren = (
+      <Text key={props.key} {...MULTICHANNEL_WHATSAPP_PROPS}>
```
guinii, line 65: Here, if we put this, it will always have markdown=false, right?? Then if for example you have a bot, working in webchat, the markdown will not be applied?

packages/botonic-plugin-contentful/src/nlp/stopwords/README.txt
```diff
@@ -0,0 +1,5 @@
+Default stopwords should not include the following:
```
guinii, line 1: Did you want to put this here? Or it was a misadd??

packages/botonic-react/src/components/multichannel/multichannel.jsx
dpinol: For "real webchat", we won't reach here due to previous guard:
```
  if (!isWhatsapp(requestContext)) {
    return props.children
  }

```
When debugging "as whatsapp" from dev machine, we will reach here and the flag will disable the transformation from markdown syntax to html.

packages/botonic-plugin-contentful/src/nlp/stopwords/README.txt
dpinol: yes, not related to the PR, but since it's only documentation... :-)

guinii (approved):

PR #899, opened 2020-08-11
feat(react): coverComponentProps

AinaVendrell:

## Description
Add `coverComponent` props

## Context
Since the `coverComponent` is shown automatically the first time that a user interacts with the bot there was no possibility to pass properties to this component

Converts the `coverComponent` to an object that contains a `component` and `props`

Instead of defining the cover component as:
```
import CustomCover from '../webchat/custom-cover'

export const webchat = {
  coverComponent: CustomCover
} 
```
Now it can be defined as: 
```
import CustomCover from '../webchat/custom-cover'

export const webchat = {
  coverComponent: {
    component: CustomCover,
    props: { ... },
  },
} 
```

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests


packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -427,8 +427,18 @@ export const Webchat = forwardRef((props, ref) => {
...
+      'coverComponent.props',
+      props.coverComponent && props.coverComponent.props
+    )
+
     if (CoverComponent && webchatState.isCoverComponentOpen)
-      return <CoverComponent closeComponent={closeCoverComponent} />
+      return (
+        <CoverComponent
+          closeComponent={closeCoverComponent}
+          coverComponentProps={coverComponentProps}
```
asastre, line 439: I think it would be better if the attributes of `coverComponentProps` are sent directly to the `coverComponent` instead of sending them encapusalted with another props (`coverComponentProps`)
```suggestion
          {...coverComponentProps}
```

asastre (changes requested):
@AinaVendrell when you say in the description 'Now it can be defined as:' I thought that this new declaration of the coverComponent was optional and you could declare it both ways, the old and the new one, but looking into the code I can see only de new declaration working.

Since I've seen it's possible to load de correct configuration of Contentful's plugin inside the `coverComponent` using an environment variable and without needing to send the configuration parameters through props, I don't know if this new feature it's a must to be implemented.
In my opinion, I think it's great to have de possibility to send information to the `coverComponent` through props, but since it's a thing that maybe won't be much used, maybe it should be nice to have backwards compatibility and ensure that the old declaration of the `coverComponent` will work too.

What do you think @ericmarcos, @dpinol, @vanbasten17 ?

dpinol:
@AinaVendrell can you follow this up?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -411,10 +411,14 @@ export const Webchat = forwardRef((props, ref) => {
     )
   }
 
-  const CoverComponent = getThemeProperty(
-    'coverComponent',
-    props.coverComponent
-  )
+  const getCoverComponent = () => {
+    return getThemeProperty(
+      'coverComponent.component',
```
dpinol, line 416: use constant

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -430,8 +434,18 @@ export const Webchat = forwardRef((props, ref) => {
   }, [])
 
   const coverComponent = () => {
+    const coverComponentProps = getThemeProperty(
+      'coverComponent.props',
```
dpinol, line 438: use constant

dpinol (changes requested):

ericmarcos:
@AinaVendrell I don't understand the utility of this. What can you achieve with this feature that you couldn't do before?

vanbasten17 (changes requested):
@asastre I also think it's nice to be able to pass props to the component. Maybe it's not very hard to implement a way to make this backward-compatible by defining the old way and having also this new definition. If it's not affordable, then I'll preserve the old compat.

asastre (approved):
Nice :clap:!!

dpinol (approved):

PR #903, opened 2020-08-14
Fix/check childRoutes in getRoute

AinaVendrell:

## Description
The function `getRoute` now checks if the input matches with a childRoute

## Context
Some tests of `botonic/core` were failing because the function `getRoute` wasn't checking the childRoutes

Call `getRoute` recursively to check if the input matches with a childRoute

## Testing

The pull request...

- [X] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-core/src/router.js
+++ b/packages/botonic-core/src/router.js
@@ -144,24 +144,34 @@ export class Router {
     return undefined
   }
 
-  getRoute(input, routes, session) {
+  getRoute(input, routes, session, previousValue) {
     const computedRoutes = isFunction(routes)
       ? routes({ input, session })
       : routes
     /* Find the route that matches the given input, if it match with some of the entries,
       return the whole Route of the entry with optional params captured if matcher was a regex */
     let params = {}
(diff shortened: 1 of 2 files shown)


ericmarcos (changes requested):
Sorry, I think I didn't explain well the desired behaviour.
When the router gets a "path payolad", that is, a payload in the format `__PATH_PAYLOAD__<path>` (for example `__PATH_PAYLOAD__initial/step-2`) this means that we should ignore `lastRoutePath` value and simply return the route identified by <path>.

The 3 tests that were skipped shouldn't be changed, they reflect the intended behaviour (notice how you changed the 3rd test to match this implementation).

I think the solution could be very simple: just at the beginning of the `processInput` function, check if we're dealing with a "path payload" (`__PATH_PAYLOAD__<path>`) and just go find that <path> ignoring the `lastRoutePath`. We already have a function `getRouteByPath` that I think you could reuse. Else, just leave the rest of logic as is.

ericmarcos (approved):
Nice!

vanbasten17 (approved):

PR #905, opened 2020-08-14
Botonic Core - Export getAvailableAgentsByQueue

asastre:
## Description
Added the export of `getAvailableAgentsByQueue` and remove no needed body from request.

## Context
Couldn't use the method `getAvailableAgentsByQueue` because was not exported.

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-core/src/handoff.js
+++ b/packages/botonic-core/src/handoff.js
@@ -152,7 +152,6 @@ export async function getAvailableAgentsByQueue(session, queueId) {
     },
     method: 'post',
     url: `${baseUrl}/v1/queues/${queueId}/get_available_agents/`,
-    data: { queue_id: queueId },
   })
   return resp.data
 }
--- a/packages/botonic-core/src/index.js
+++ b/packages/botonic-core/src/index.js
@@ -6,6 +6,7 @@ export {
   HandOffBuilder,
   storeCaseRating,
   getAvailableAgents,
+  getAvailableAgentsByQueue,
   getAgentVacationRanges,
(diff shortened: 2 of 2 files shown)


packages/botonic-core/src/handoff.js
```diff
@@ -152,7 +152,6 @@ export async function getAvailableAgentsByQueue(session, queueId) {
     },
     method: 'post',
     url: `${baseUrl}/v1/queues/${queueId}/get_available_agents/`,
-    data: { queue_id: queueId },
   })
```
dpinol, line 155: I guess line removed by mistake
@guinii reviewers should stop these issues!

dpinol (changes requested):

packages/botonic-core/src/handoff.js
asastre: @dpinol No, it's not removed by mistake and @guinii was aware of this line removed. We saw doing tests with Postman that the body in this request is not needed (in fact, the queueId is already in the URL itself!) and that's why I removed it

dpinol: @asastre @guinii my sincere apologies!

dpinol (approved):

packages/botonic-core/src/handoff.js
asastre: @dpinol no problem! It's good that you see these (sometimes right and sometimes wrong) things :eyes:!!

PR #915, opened 2020-08-27
Contentful/error recovery

dpinol:
Depends on #913  Please **review it first** 

## Description

* Encapsulate logic to throw or not throw depending on context.resumeErrors
* Normally throw when URL is empty 
* Limit concurrency to aid debugging
## Testing

The pull request...

-  has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/tools/validate-all-contents.ts
+++ b/packages/botonic-plugin-contentful/src/tools/validate-all-contents.ts
@@ -1,4 +1,5 @@
 import { CMS, ContextWithLocale, TOP_CONTENT_TYPES } from '../cms'
+import * as Parallel from 'async-parallel'
 
 export class ContentsValidator {
   constructor(readonly cms: CMS) {}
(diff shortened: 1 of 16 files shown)


packages/botonic-plugin-contentful/src/tools/validate-all-contents.ts
```diff
@@ -10,11 +11,13 @@ export class ContentsValidator {
...
    * - There's a bug on the CMS plugin
    */
-  async validateAllTopContents(context: ContextWithLocale) {
-    await Promise.all(
-      TOP_CONTENT_TYPES.map(ct => {
-        return this.cms.topContents(ct, context)
-      })
+  async validateAllTopContents(context: ContextWithLocale): Promise<void> {
+    // topContents will internally limit concurrency
+    const concurrency = context.concurrency && 1
```
guinii, line 16: Why this `&& 1`?

guinii (approved):

packages/botonic-plugin-contentful/src/tools/validate-all-contents.ts
dpinol: we want to limit to 1 when not specified because the call to topContents will already execute in parallel

asastre (approved):

PR #920, opened 2020-09-02
Contentful/async error reporting

dpinol:
## Description

* When promises executed with async-parallel library, retrieve the list of error occurred for each item and included in the top thrown exception
* Show locale in Error Reporting decorator
* flag to optionally log CMS calls 

## Context

* Users typically leave inconsistent contents in contentful. This PR provide way to report with detail all discovered issues

New logCalls flag in Contentful plugin config to log each call performed to contentful (useful during debugging)

## Testing

-  has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/tools/validate-all-contents.ts
+++ b/packages/botonic-plugin-contentful/src/tools/validate-all-contents.ts
@@ -1,5 +1,5 @@
 import { CMS, ContextWithLocale, TOP_CONTENT_TYPES } from '../cms'
-import * as Parallel from 'async-parallel'
+import { asyncEach } from '../util/async'
 
 export class ContentsValidator {
   constructor(readonly cms: CMS) {}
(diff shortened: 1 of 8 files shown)


packages/botonic-plugin-contentful/src/tools/validate-all-contents.ts
```diff
@@ -12,12 +12,12 @@ export class ContentsValidator {
...
-    // topContents will internally limit concurrency
-    const concurrency = context.concurrency && 1
-    await Parallel.each(
+    await asyncEach(
+      context,
       TOP_CONTENT_TYPES,
       ct => this.cms.topContents(ct, context),
-      concurrency
+      // topContents will internally manage concurrency
+      1
```
asastre, line 20: If `topContents` will manage concurrency internally, will it be better if this parameter is ignored in this call in order to be undefined?

asastre (approved):
All good! Just one small observation.

packages/botonic-plugin-contentful/src/tools/validate-all-contents.ts
dpinol: If we specify undefined, it will request all content models simultaneously, which is what we don't want.
We want to get first all texts in parallel, then all carousels in parallel...That's why we request concurrency=1. Call me if it's not clear

guinii (approved):

PR #924, opened 2020-09-04
feat(contentful): content validator only if reachable

dpinol:
## Description

Flag so that  ContentValidator only validates contents which are reachable.
Check empty text in Text contents
 
## Context

Some contents are not available for some locales. We don't want ContentValidator to complain when they have empty fields.

Added a flag in Context to to automatically fix the buttons when no shortText is available. Setting this flag is now possible to detect these cases.

ContentsValidator is a useful class because contentful dashboard allow leaving contents in an inconsistent state.
See how to use in unit test https://github.com/hubtype/botonic/blob/master/packages/botonic-plugin-contentful/tests/tools/validate-all-contents.test.ts

## Testing

- has integration tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/contentful/contents/button.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/contents/button.ts
@@ -67,8 +67,9 @@ export class ButtonDelivery extends ContentDelivery {
     const entry = await this.delivery.getEntry(id, context)
     const entryType = ContentfulEntryUtils.getContentModel(entry)
     if (isOfType(entryType, TopContentType)) {
-      return ButtonDelivery.fromContentReference(
-        entry as contentful.Entry<CommonEntryFields>
+      return this.fromContentReference(
+        entry as contentful.Entry<CommonEntryFields>,
+        context
       )
     }
     if (entryType === ButtonDelivery.BUTTON_CONTENT_TYPE) {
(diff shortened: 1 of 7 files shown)


guinii (approved):

packages/botonic-plugin-contentful/src/contentful/contents/button.ts
```diff
@@ -95,12 +96,14 @@ export class ButtonDelivery extends ContentDelivery {
...
-    entry: contentful.Entry<CommonEntryFields>
+  private fromContentReference(
+    entry: contentful.Entry<CommonEntryFields>,
+    context: cms.Context
   ): cms.Button {
     const fields = entry.fields
     let text = fields.shortText
-    if (!text) {
+    const fixMissingData = context.fixMissingData ?? true
+    if (!text && fixMissingData) {
```
asastre, line 106: @dpinol If `fixMissingData` is set to `false`, no message will be logged, is this the expected behaviour? Also, the Button will be created with an empty text, may this cause any errors?

asastre (approved):
Take a look at the message I've sent, the rest seems ok!

packages/botonic-plugin-contentful/src/contentful/contents/button.ts
dpinol: if the developer manually sets the flag to false, I don't see why we should log anything. Probably developers will only do it when they set empty data purposely, and then they don't want spam in logs.

>Button will be created with an empty text, may this cause any errors?
Yes, that might crash. I fixed it at  4c4812c391d203ba2c42d22f85dbf33aa21c6d8e on the lastest PR. thanks!

PR #926, opened 2020-09-07
feat(contentful): im/export URLS for translators

dpinol:
## Description
* In tools for import/export content texts for translators, we had forgotten the URLs. Now added
* Also, now URLs are automatically converted to the target country. eg. when importing to fr_LU URLs such as https:/.../es/...html are transformed to https:/.../lu/...html

## Testing

The pull request...

- [has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/tools/l10n/csv-export.ts
+++ b/packages/botonic-plugin-contentful/src/tools/l10n/csv-export.ts
@@ -7,9 +7,10 @@ import {
   Element,
   StartUp,
   TopContent,
+  Url,
 } from '../../index'
 import { Locale } from '../../nlp'
-import { Text } from '../../cms'
+import { Text, ContentType } from '../../cms'
 import * as stream from 'stream'
 import * as fs from 'fs'
 import { promisify } from 'util'
@@ -19,7 +20,7 @@ import { ContentField, ContentFieldType } from '../../manage-cms/fields'
 
 const finished = promisify(stream.finished)
 
-class I18nField {
+export class I18nField {
   constructor(readonly name: ContentFieldType, readonly value: string) {}
 }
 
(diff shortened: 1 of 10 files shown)


packages/botonic-plugin-contentful/src/tools/l10n/csv-export.ts
```diff
@@ -68,8 +72,7 @@ export class CsvExport {
   }
 
   async *generate(cms: CMS, from: Locale): AsyncGenerator<CsvLine> {
-    console.error('TODO: pending to export URL models')
-    for (const model of BOTONIC_CONTENT_TYPES) {
+    for (const model of [...BOTONIC_CONTENT_TYPES, ContentType.URL]) {
```
guinii, line 75: TODO: add in "MessageContentType" the URL type, so it can be used in other parts in the future?

guinii (approved):

packages/botonic-plugin-contentful/src/tools/l10n/csv-export.ts
dpinol: would be nice, but since URL is not a React message, I decided not to do it

packages/botonic-plugin-contentful/tests/tools/l10n/csv-export.test.ts
```diff
@@ -34,6 +35,21 @@ test('TEST: ContentToCsvLines.getCsvLines Text', () => {
   ])
 })
 
+test('TEST: ContentToCsvLines.getCsvLines URL', () => {
```
asastre, line 38: It would be nice to add a test for URLS with a /es/ in its URL and confirm that PostProcessor is working corrrectly

asastre (commented):
Why is there the need to translate the language from the URL when exporting contents from Contentful?
I think it's a little bit confusing if your exported content is not the same as it is in Contentful. 
I don't think it will ever happen, but there is also a little chance of change incorrectly a part of the URL if the "/es/", "/en/", etc. is not referred to the language.

packages/botonic-plugin-contentful/tests/tools/l10n/csv-export.test.ts
dpinol: yes, but run out of time :-(

dpinol:
> Why is there the need to translate the language from the URL when exporting contents from Contentful?
> I think it's a little bit confusing if your exported content is not the same as it is in Contentful.
> I don't think it will ever happen, but there is also a little chance of change incorrectly a part of the URL if the "/es/", "/en/", etc. is not referred to the language.

Yes, that's why if you don't specify the target_locale argument, it will not perform this conversion.
I improved the documentation.
thanks for the comment!

PR #935, opened 2020-09-16
fix(react): fix change button text color on hover

asastre:
## Description
Rename theme property `hoverText` to `hoverTextColor` in `button.jsx` so changing button's text color on hover could work

## Context
Changing the text color of a button when the mouse was over it (`:hover`) wasn't working.

There are 2 different variable names that match the same theme property, `hoverTextColor` in [`constants.js`](https://github.com/hubtype/botonic/blob/master/packages/botonic-react/src/constants.js#L80) and `hoverText` in [`button.jsx`](https://github.com/hubtype/botonic/blob/master/packages/botonic-react/src/components/button.jsx#L82). 
I decided to use `hoverTextColor` because it's more representative of the usage of the property and also because that's the one that appears in the [documentation](https://docs.botonic.io/templates/template-custom-webchat/#webchat-styles)

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-react/src/components/button.jsx
+++ b/packages/botonic-react/src/components/button.jsx
@@ -79,7 +79,7 @@ export const Button = props => {
       ? getThemeProperty('button.hoverBackground', COLORS.CONCRETE_WHITE)
       : getThemeProperty('button.style.background', COLORS.SOLID_WHITE)
     const buttonTextColor = hover
-      ? getThemeProperty('button.hoverText', COLORS.SOLID_BLACK)
+      ? getThemeProperty('button.hoverTextColor', COLORS.SOLID_BLACK)
       : getThemeProperty('button.style.color', COLORS.SOLID_BLACK)
 
     return (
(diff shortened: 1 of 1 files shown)


packages/botonic-react/src/components/button.jsx
```diff
@@ -79,7 +79,7 @@ export const Button = props => {
       ? getThemeProperty('button.hoverBackground', COLORS.CONCRETE_WHITE)
       : getThemeProperty('button.style.background', COLORS.SOLID_WHITE)
     const buttonTextColor = hover
-      ? getThemeProperty('button.hoverText', COLORS.SOLID_BLACK)
+      ? getThemeProperty('button.hoverTextColor', COLORS.SOLID_BLACK)
```
dpinol, line 82: please take the opportunity to use the constants from botonic/packages/botonic-react/src/constants.js (where the value is correct)

dpinol (changes requested):

packages/botonic-react/src/components/button.jsx
asastre: Done it @dpinol! 
Yes, better using constants. I've also changed the other `getThemeProperty` calls whithin `button.jsx` and now they are using constants too.

ericmarcos (approved):

dpinol (approved):
thank you for using the constants!

PR #936, opened 2020-09-16
feat(react): config storage

AinaVendrell:

## Description

Define where to store the `botonicState` using the variable `storage`

## Context
In some cases it may be better to store the `botonicState` in the `sessionStrorage` or not storing it

Change [@rehooks/local-storage](https://www.npmjs.com/package/@rehooks/local-storage)  for [react-storage-hooks](https://www.npmjs.com/package/react-storage-hooks) since it allows to store the application state in the `sessionStorage` or in the `localStorage`

By default the `botonicState` will be stored in the `localStorage`. 

Code to change the `storage`:
```
export const webchat = {
  storage: {localStorage|sessionStorage|null} 
}
```
difference between `localStorage` and `sessionStorage`: https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage

## Testing

The pull request...

- [X] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-react/src/webchat/webchat.jsx
+++ b/packages/botonic-react/src/webchat/webchat.jsx
@@ -5,7 +5,7 @@ import React, {
   forwardRef,
 } from 'react'
 import Textarea from 'react-textarea-autosize'
-import { useLocalStorage } from '@rehooks/local-storage'
+import { useStorageState } from 'react-storage-hooks'
 import { v4 as uuidv4 } from 'uuid'
 import UAParser from 'ua-parser-js'
 import { isMobile, params2queryString, INPUT } from '@botonic/core'
@@ -190,10 +190,25 @@ export const Webchat = forwardRef((props, ref) => {
   const theme = merge(webchatState.theme, props.theme)
   const { initialSession, initialDevSettings, onStateChange } = props
   const isOnline = useNetwork()
-  const [botonicState, saveState, deleteState] = useLocalStorage('botonicState')
+  const getThemeProperty = _getThemeProperty(theme)
+
+  const getStorage = () => {
+    const storage = getThemeProperty(
+      WEBCHAT.CUSTOM_PROPERTIES.storage,
+      props.storage
+    )
+    if (storage === undefined) return localStorage
+    return storage
+  }
+
+  const [botonicState, saveState, writeError] = useStorageState(
+    getStorage(),
+    'botonicState'
+  )
+
   const saveWebchatState = webchatState => {
-    saveState(
-      stringifyWithRegexs({
+    getStorage() &&
+      saveState({
         user: webchatState.user,
         messages: webchatState.messagesJSON,
         session: webchatState.session,
@@ -202,11 +217,8 @@ export const Webchat = forwardRef((props, ref) => {
         lastMessageUpdate: webchatState.lastMessageUpdate,
         themeUpdates: webchatState.themeUpdates, // can contain regexs
       })
-    )
   }
   const deviceAdapter = new DeviceAdapter()
-
-  const getThemeProperty = _getThemeProperty(theme)
 
   const handleAttachment = event => {
     if (!isAllowedSize(event.target.files[0].size)) {
@@ -239,7 +251,7 @@ export const Webchat = forwardRef((props, ref) => {
   const resendUnsentInputs = async () =>
     props.resendUnsentInputs && props.resendUnsentInputs()
 
-  // Load initial state from localStorage
+  // Load initial state from storage
   useEffect(() => {
     let {
       user,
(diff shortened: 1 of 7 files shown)


packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -421,7 +434,8 @@ export const Webchat = forwardRef((props, ref) => {
     if (!CoverComponent) return
     if (
       !botonicState ||
-      (botonicState.messages && botonicState.messages.length == 0)
+      (JSON.parse(botonicState).messages &&
+        JSON.parse(botonicState).messages.length == 0)
```
ericmarcos, line 438: Is `JSON.parse` really needed? I had a quick review of the library you're using here (https://github.com/soyguijarro/react-storage-hooks) and it seems it already parses the result from the storage automatically.

If you really need to parse it manually, please do it once right after the botonicState is loaded to avoid repeating a potential costly operation.

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -190,24 +190,36 @@ export const Webchat = forwardRef((props, ref) => {
...
   const isOnline = useNetwork()
-  const [botonicState, saveState, deleteState] = useLocalStorage('botonicState')
+  const getThemeProperty = _getThemeProperty(theme)
+
+  const getStorage = () => {
+    const storage = getThemeProperty('storage', props.storage)
+    if (storage === 'none') return null
+    if (storage === 'sessionStorage') return sessionStorage
+    return localStorage
+  }
```
ericmarcos, line 200: Let's let the user provide the storage class directly. So in their project they can do:

```
webchat = {
   ...,
   storage: window.localStorage
}
```

This will allow for more flexibility in the future to provide custom storages.

If they don't want the state to be saved:
```
webchat = {
   ...,
   storage: null,
}
```
it's more natural than a string with value "none"

ericmarcos (changes requested):

packages/botonic-react/tests/webchat/storage.test.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
```
dpinol, line 1: rename to storage.test.jsx

packages/botonic-react/tests/webchat/storage.test.js
```diff
@@ -0,0 +1,42 @@
...
+    await act(async () => {
+      TestRenderer.create(<Webchat />)
+    })
+    expect(localStorage.getItem('botonicState')).not.toBeNull()
+    expect(sessionStorage.getItem('botonicState')).toBeNull()
+  })
+
+  it('Stores botonicState in the localStorage', async () => {
+    await act(async () => {
+      TestRenderer.create(<Webchat storage={'localStorage'} />)
```
dpinol, line 21: I think you don't need the brackets at {'localStorage'}
try 
`<Webchat storage='localStorage'}`

packages/botonic-react/tests/webchat/storage.test.js
```diff
@@ -0,0 +1,42 @@
...
+  afterEach(() => {
+    localStorage.clear()
+    sessionStorage.clear()
+  })
+
+  it('Stores botonicState in the localStorage by default', async () => {
+    await act(async () => {
+      TestRenderer.create(<Webchat />)
+    })
+    expect(localStorage.getItem('botonicState')).not.toBeNull()
```
dpinol, line 15: maybe you can check here the contents of botonicState if it's not tested anywhere else?
if required, only check the contents in 1 test

dpinol (changes requested):

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -190,10 +190,22 @@ export const Webchat = forwardRef((props, ref) => {
   const theme = merge(webchatState.theme, props.theme)
   const { initialSession, initialDevSettings, onStateChange } = props
   const isOnline = useNetwork()
-  const [botonicState, saveState, deleteState] = useLocalStorage('botonicState')
+  const getThemeProperty = _getThemeProperty(theme)
+
+  const getStorage = () => {
+    const storage = getThemeProperty('storage', props.storage)
+    if (storage === undefined) return localStorage
```
dpinol, line 197: be aware that "null === undefined".
Which means that the unit test where you call  
`TestRenderer.create(<Webchat storage={null} />)
`
might break in the future if getThemeProperty implementation is changed.
I think it's better to use == undefined

dpinol (changes requested):

packages/botonic-react/src/webchat/webchat.jsx
AinaVendrell: In case `storage == null` it should return null and then the botonicState will not be stored anywhere. Otherwise, if `storage == undefined` the botonicState will be stored in the localStorage. If I use `null == undefined`, it does not work as desired.

ericmarcos (approved):
Looks good to me!

vanbasten17 (approved):
Nice one with cool tests!! 👏 👏

PR #951, opened 2020-09-22
feat(react): Enable coverComponent toggle through Botonic

asastre:
## Description
Make the methods `openCoverComponent`, `closeCoverComponent` and `toggleCoverComponent` available through `Botonic` object in browser so the `coverComponent` can be shown/hide manually.

## Context
Once the user has gone through the `coverComponent`, there is no way to show it again without not removing the cookies.

Make the methods `openCoverComponent`, `closeCoverComponent` and `toggleCoverComponent` available both in `webchatRef` and `WebchatApp`.

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-react/src/webchat-app.jsx
+++ b/packages/botonic-react/src/webchat-app.jsx
@@ -138,6 +138,18 @@ export class WebchatApp {
     this.webchatRef.current.toggleWebchat()
   }
 
+  openCoverComponent() {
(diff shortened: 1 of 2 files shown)


vanbasten17 (approved):
LGTM 👍

dpinol (approved):

Issue #928, opened 2020-09-08
Adding name and version fields to generated package.json


chopfitzroy:
**Describe the bug**

Currently the `cli` does not add `name` or `version` to the `package.json` this can cause [issues](https://github.com/lerna/lerna/issues/2127) when working with tools like `lerna`.

**To Reproduce**
Steps to reproduce the behaviour:
1. Generate new project
2. Open `package.json`
3. `name` and `version` are not present

**What is the expected behaviour?**

Fields would be added, name based on name given when using `cli` and version could default to something like `0.0.0` or `1.0.0`.

**Screenshots**

**About the environment**
Which versions of [**botonic** packages](https://github.com/hubtype/botonic/tree/master/packages), and which browser/OS or environment (development/production) are affected by this issue? Did this work in previous versions of [**botonic** packages](https://github.com/hubtype/botonic/tree/master/packages)? 

Using latest `cli` version.

**Additional context**

vanbasten17:
We have added both name and version to every template in the new [released package](https://www.npmjs.com/package/@botonic/cli/v/0.14.0). Thanks for reporting! 😊

PR #962, opened 2020-09-29
chore(templates/react): add webpack entries to all templates

vanbasten17:
_Set as [Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) if it's not ready to be merged_.  

[PR best practices Reference](https://blog.codeminer42.com/on-writing-a-great-pull-request-37c60ce6f31d/)

## Description
~As mentioned in the tech daily, applying the trick we did for other providers projects to every template, by adding the corresponding `webchat-entry.js` in `@botonic/react` src.~ (Ignore commit: https://github.com/hubtype/botonic/pull/962/commits/36ea33c4841ba63bc04ef9a813c7747c5c6326ed) 
- **Important:** Reverted pusher to version `5.1.1` in `@botonic/core` as newer versions `> 5.1.1` introduce breaking changes that make webchat to not work.
- Added `webpack-entries` with every entry file. Then the alias like `BotonicProject` won't be used anymore, so the pipeline is less "magical" (the developer will be conscious that these files are to configure webpack)
- Added name and version to every template as reported: https://github.com/hubtype/botonic/issues/928

## Context
The expected generated `webchat.botonic.js` after a bot build was larger than expected, causing performance issues (extra download cost when loading the bot code).

We have to take into account that developers upgrading from older versions than 0.14.0 must add the `webpack-entries` folder to their root project and do the proper modifications in their `webpack.config.js`.

## Next steps
Try webpack v5.


packages/botonic-cli/templates/blank/webpack.config.js
```diff
@@ -51,6 +51,7 @@ const resolveConfig = {
   extensions: ['*', '.js', '.jsx', '.ts', '.tsx'],
   alias: {
     BotonicProject: path.resolve(__dirname, 'src'),
+    webchat: path.resolve(__dirname, 'src', 'webchat'),
```
asastre, line 54: Why is this alias needed? When is it used?

vanbasten17: This is used in order to import `webchat` from an outer package, in our case `@botonic/react` with no need to use relative paths. Here the line in question: 
https://github.com/hubtype/botonic/blob/a76b97494d7deb0948ee05a6b00d9dc2390ca1d1/packages/botonic-react/src/webchat-entry.js#L3

vanbasten17:
It solves https://github.com/hubtype/botonic/issues/928

packages/botonic-cli/templates/childs/webpack-entries/dev-entry.js
```diff
@@ -0,0 +1,19 @@
+import { DevApp } from '@botonic/react'
```
dpinol, line 1: I would leave all templates as they are except the ones about webchat & webviews for several reasons:
1) minimize changes
2) to make examples simpler, to avoid scaring developers if templates are too complex 
3) in case we refactor botonic in the future, it will be easier to modify 1 file per bot rather than 4 files per bot

packages/botonic-cli/templates/childs/webpack-entries/dev-entry.js
```diff
@@ -0,0 +1,19 @@
...
+
+import { routes } from '../src/routes'
+
```
dpinol, line 4: why some many blank lines?

dpinol (changes requested):

packages/botonic-cli/templates/childs/webpack-entries/dev-entry.js
ericmarcos: We were talking about this with @marcrabat yesterday, and we finally decided to go this way because:
1. It is a sure and simple way to enforce tree shaking of each artifact.
2. We avoid the "magic" of webpack aliases, I think that's more confusing for the developer than having some extra files with a clear purpose. It provides clarity about how things are built and also provides flexibility in case they want to modify the entry points.
3. This is actually more future proof than the old version, as we "eject" the entry files to the project code, we no longer need to keep that file in botonic source code, so we don't need to worry about backwards compatibility (that responsibility is no longer in botonic). Developers will be able to upgrade botonic versions in their old project without worrying about having to change their webpack config.

packages/botonic-cli/templates/childs/webpack-entries/dev-entry.js
vanbasten17: my fault, just removed them Dani, thanks!

ericmarcos (approved):

packages/botonic-cli/templates/childs/webpack-entries/dev-entry.js
dpinol: I agree with separating the entry files, but I was suggesting that for the templates which have no webchat nor webviews, it was a bit confusing having so many files.

asastre (approved):
Love this improvement :heart: :clap:!

packages/botonic-cli/templates/childs/webpack-entries/dev-entry.js
vanbasten17: I think that it's better to have all of them already set up in case the developer wants to add webviews or webchat later (which can be very common).  Otherwise the develeper should need to research these files somewhere and I think won't be favourable for his DX. I understand it can be more cumbersome to maintain, but all the files in `webpack-entries` should be the same for all templates and the entire folder is easy to replicate.

PR #965, opened 2020-09-29
docs: Components and Plugins sections refactored

annehubtype:
_Set as [Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) if it's not ready to be merged_.  

[PR best practices Reference](https://blog.codeminer42.com/on-writing-a-great-pull-request-37c60ce6f31d/)

## Description

The documentation sections have been reorganised into 4 main menu entries: 
- User Guide with high level content.
- Reference Guide with a list of components and concepts.
- Plugins where it is explained how to use existing plugins, create new ones and the list of available plugins.
- An FAQ page with updated content based on user's feedback.

Thanks for your feedback and suggestions.

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests


vanbasten17 (changes requested):
I have splitted this into things I think that are more prioritary, and the other ones that are less prioritary:

**More prior**
* "Understanding the project" is duplicated in user guide and reference guide. Should it be like this?
* I suggest to move “”How to make your bot compatible with IE11 in “Webchat” section at the end of this section. For example, you can add a link inside the table where we have “11 (with polyfills)”. In this way, we assure the developer it’s focused in the main content of the Webchat explanations.
* I suggest to move the section “Childs” somewhere inside the “Routes” section. Childs is not a “concept” but a predefined template, but the explanations there can be added somehow into “Routes”.
* In “Webchat SDK API” in “Webchat” we need to add the following new methods: `openCoverComponent`, `closeCoverComponent` and `toggleCoverComponent`.
* In “Customizing Webchat” in “Webchat” I wouldn’t remove most of the information we already have here https://docs.botonic.io/templates/template-custom-webchat, as it’s great to know what are many of the properties for customizing the webchat and brief examples of how to use them.
* I think it would be better to have Element, Title, Subtitle and Pic (in this order) documented within the same section of Carousel component. I think we are difficulting the reading of the dev otherwise.
* Add a code snippet of `Custom Message`, `Webchat Settings` usage?
* Review changelog for things pending to documentate (classNames for messages, ability to define where the botonicState is stored, new timestamps logic, … ). If we can introduce it for the release would be nice, otherwise, let’s create a Linear task. Let me know if you have any doubts :)

**Less prior**
* As the section `deployment` has been moved to user reference, maybe I would put more emphasis in the final part of “Create a Conversational App from Scratch“, to highlight even more the fact of deploying the bot just before moving to “Integrate a Bot with Webchat and Messaging Apps”. Also it can be useful to link this emphatized part to the “Deployment” section for further information.
* In the concept “Human Handoff” it would be nice if we can add some kind of tables with the options available to do the handoff (https://github.com/hubtype/botonic/blob/9b15c2bdffc548d61c27760ae8393fb0d9261625/packages/botonic-core/src/index.d.ts#L142) and other simple tables explaining the methods available to request information from agents: like `getAvailableAgentsByQueue`, `getOpenQueues`, etc. (You can find some definitions here: https://github.com/hubtype/botonic/blob/master/packages/botonic-core/src/index.d.ts)
* I think that “Introduction to Botonic Plugins” doesn’t add additional info in the way it is separated from other sections. Maybe it makes more sense to have it just at the beginning of “Prepare the plugin”. What do you think?


Thanks for the huge effort of restructuring the docs, now it looks more intuitive!! I really love the new reference tables and the plugins library section 😍 💯

PR #985, opened 2020-10-08
Contentful/validation improvements

dpinol:
Please review commit by commit

-  logger injection in ErrorReportingCMS: Not we allow customizing the sink of the error messages.Improvements in prints of recursive exception stacks

-  Improvements in Content.validate of text & shortText fields
    - Validate text.text also if content has no keywords (it could be accessed through a button)
    - Before, when shortText was empty, the contentful driver set it with the value of the name field (to ensure buttons did not show blank texts). Now this is managed by the Button model, so that MessageContent.validate() can report it

- Check that contentful button targets are available 

- Added asserts to asserts in ContentsValidator unit test

## To document
**breaking change**:  We remove the Context.fixMissingData flag. It was confusing because:
    * we already set shortText with the name field when it's missing
    * we have the resumeErrors option for contentful


asastre (approved):

packages/botonic-plugin-contentful/tests/contentful/contents/button.test.ts
```diff
@@ -0,0 +1,9 @@
+import { testContentful } from '../contentful.helper'
+import { ENGLISH } from '../../../src/nlp'
+
+test('TEST: contentful button', async () => {
```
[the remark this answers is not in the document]
dpinol: @asastre 
well, seen! I added asserts

packages/botonic-plugin-contentful/src/cms/cms-error.ts
```diff
@@ -201,29 +209,24 @@ export class ContentfulExceptionWrapper {
...
-        }
+  processError(contentfulError: Error): string {
+    let err = ''
+    if (contentfulError instanceof MultiError) {
+      err += ' Due to:'
+      for (const e of reduceMultiError(contentfulError)) {
+        err += this.processError(e)
       }
+    } else if (this.logStack && contentfulError.stack) {
+      err += contentfulError.stack.split('\n').slice(0, 5).join('\n')
```
guinii, line 226: Maybe a bit "hacky" Why are you slicing here?

guinii (approved):

packages/botonic-plugin-contentful/src/cms/cms-error.ts
dpinol: we don't want to spam the logs with too long callstack

PR #987, opened 2020-10-09
Contentful - Update content's Code when importing CSV from translators

asastre:
## Description
Add an extra parameter `overwriteCodes` to the script `import-from-translators.sh` in order to be able to update all the content's Codes in bulk during the CSV import.

## Context
There is no way to change the content's Codes (unique key forced by the user for each content) in bulk with this plugin.

I've used the actual import script to add this new feature to be able to change the Codes in bulk while importing the translated contents.

If you only want to update the Codes without changin the main contents you can just export the contents into a CSV with the script `export-for-translators.sh`, change all the Codes you want to update and then import this file with the script `import-from-translators.sh` with the options `writeMode` set to `OVERWRITE` or `OVERWRITE_AND_PUBLISH` and `overwriteCodes` to `true`.

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
@@ -60,15 +60,25 @@ export class ManageContentful implements ManageCms {
     context: ManageContext,
     contentId: ContentId,
     fieldType: ContentFieldType,
-    value: any
+    value: any,
+    code?: string
   ): Promise<void> {
     const environment = await this.getEnvironment()
     const oldEntry = await environment.getEntry(contentId.id)
+
     const field = this.checkOverwrite(context, oldEntry, fieldType, true)
     oldEntry.fields[field.cmsName][context.locale] = value
+    if (code) this.replaceCode(code, oldEntry)
+
     // we could use this.deliver.contentFromEntry & IgnoreFallbackDecorator to convert
     // the multilocale fields returned by update()
     await this.writeEntry(context, oldEntry)
+  }
+
+  replaceCode(code: string, entry: Entry) {
+    for (const locale in entry.fields.name) {
+      entry.fields.name[locale] = code
+    }
   }
 
   async removeAssetFile(
--- a/packages/botonic-plugin-contentful/bin/l10n/import-from-translators.sh
+++ b/packages/botonic-plugin-contentful/bin/l10n/import-from-translators.sh
@@ -25,7 +25,8 @@ WRITE_MODE=$7
 #  OVERWRITE_AND_PUBLISH: overwrites previous value and publishes it (only for new spaces!)
 
 DUPLICATE_REFERENCES=$8
+OVERWRITE_CODES=$9
 
 ../../node_modules/.bin/ts-node --files	src/tools/l10n/import-csv-from-translators.ts \
 	"$SPACE_ID" "$ENVIRONMENT" "$DELIVER_TOKEN" "$CONTENTFUL_MANAGEMENT_TOKEN" "$LOCALE" "$CSV_FILENAME" \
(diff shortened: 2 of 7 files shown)


packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
```diff
@@ -60,17 +60,27 @@ export class ManageContentful implements ManageCms {
     context: ManageContext,
     contentId: ContentId,
     fieldType: ContentFieldType,
-    value: any
+    value: any,
+    code?: string
```
dpinol, line 64: I don't think we need a special argument for the code. We should only add a new ContentFieldType

packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
```diff
@@ -60,17 +60,27 @@ export class ManageContentful implements ManageCms {
...
   ): Promise<void> {
     const environment = await this.getEnvironment()
     const oldEntry = await environment.getEntry(contentId.id)
+
     const field = this.checkOverwrite(context, oldEntry, fieldType, true)
     oldEntry.fields[field.cmsName][context.locale] = value
```
dpinol, line 70: code is not localizable, so this might not work.
Maybe you should have a isFieldLocalizable function and the here assign directly (without [context.locale]) for codes.

packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
```diff
@@ -60,17 +60,27 @@ export class ManageContentful implements ManageCms {
...
+    if (code) this.replaceCode(code, oldEntry)
+
     // we could use this.deliver.contentFromEntry & IgnoreFallbackDecorator to convert
     // the multilocale fields returned by update()
     await this.writeEntry(context, oldEntry)
   }
 
+  replaceCode(code: string, entry: Entry) {
+    for (const locale in entry.fields.name) {
```
dpinol, line 79: since code is not localizable, I'm surprised this works

dpinol (changes requested):

packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
dpinol: to avoid making 1 call to change the code per field, this method could validate if the previous value of the entry already has the desired value

packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
asastre: Yes, it's not localizable, but this is how `entry.fields.name` is formed, is an object with a `key: value` pair where the `key` is the default locale.
I did the `for` just in case the Code field was localizable.

packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
asastre: This line is for replacing the content of the ContentFieldType selected and this is localizable. The replace of the Code is done in `replaceCode` method (next line)

packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
asastre: Actually it's not making a new call to change the Code, it's using the actual call that change the contents to change the Code too.

asastre:
PR closed because makes the code too complex and mixes 2 separate functionalities (import translations and change Codes in bulk) in the same script. 
In the future, if changing codes in bulk is needed, this should have to be reviewed and refactored.

PR #988, opened 2020-10-09
refactor(react): move user refs to session.user

vanbasten17:
## Description
_This is an enhancement PR. We must consider if we want to have this implemented like this._
* Moving logic regarding `user` to `session.user` to avoid confusions and update it altogether with session.

## Context
We were  having two references to the same user in our webchatState. One was managed by the `user` object, and the other one by `session.user`. To avoid confusions, we will be always taking the user coming from the session variable.

## Testing
I have tested it in prod environment and dev and it seems to work properly, but we agreed with @asastre to test it before with some bots.
The pull request has **unit tests**


packages/botonic-react/tests/webchat/storage.test.jsx
```diff
@@ -13,9 +13,9 @@ describe('TEST: storage', () => {
       TestRenderer.create(<Webchat />)
     })
     const botonicState = JSON.parse(localStorage.getItem('botonicState'))
-    expect(botonicState.user).toHaveProperty('id' && 'name')
+    expect(botonicState.session.user).toHaveProperty('id' && 'name')
```
marcpdw, line 16: Oh, cool! Jasmine does not have this `toHaveProperty` check! Still, is this expression correct to check for these two properties? Wouldn't it be... ?
```
expect(botonicState.session.user).toHaveProperty(['id', 'name'])
```

marcpdw (approved):
LGTM!

asastre (approved):
Looks good! I've applied this changes into 2 bots and both seem to work as expected :clap: !

packages/botonic-react/tests/webchat/storage.test.jsx
vanbasten17: it seems to not work well by passing it like an array. I've just leave it as is 👍

PR #1000, opened 2020-10-14
fix(react): customMessage not to crash if bad children

dpinol:
Avoid customMessage to crash if they have bad children. Some bots may store no-react objects in children

### DIFF
--- a/packages/botonic-react/src/components/custom-message.jsx
+++ b/packages/botonic-react/src/components/custom-message.jsx
@@ -23,7 +23,7 @@ export const customMessage = ({
       )
       return { replies, childrenWithoutReplies }
     } catch (e) {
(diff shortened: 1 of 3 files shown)


guinii (approved):

ericmarcos (approved):

asastre (approved):

vanbasten17 (approved):

packages/botonic-react/src/components/custom-message.jsx
```diff
@@ -11,14 +11,25 @@ export const customMessage = ({
   const CustomMessage = props => (
     <Message {...defaultProps} {...props} type={INPUT.CUSTOM} />
   )
+
+  const SplitChildren = props => {
```
vanbasten17, line 15: lowercase?

PR #1001, opened 2020-10-18
Contentful/content

dpinol:
Improvements in Contentful error reporting
* improve ContentsValidator. Better logs and get a list of errors
* Add ResourceId (content type and id) to CmsException
* Improve contents of the CmsException messages
* Improve  toString of Button contents


packages/botonic-plugin-contentful/src/cms/exceptions.ts
```diff
@@ -1,12 +1,32 @@
...
+      return undefined
+    }
+    if (this.reason instanceof MultiError) {
+      return this.reason.list.map(e => e.message).join('\n')
+    }
+    if (this.reason instanceof Error) {
+      return this.reason.message
+    }
+    return String(this.reason)
+  }
```
vanbasten17, line 28: just a matter of "style", have you considered to save the `{ }` and do the returns in the same lines? Same in other places. Which style guide should we follow in general?

vanbasten17 (approved):
LGTM! 👍

packages/botonic-plugin-contentful/src/cms/exceptions.ts
dpinol: there's notable consensus on using braces. See https://softwareengineering.stackexchange.com/questions/16528/single-statement-if-block-braces-or-no

packages/botonic-plugin-contentful/src/cms/callback.ts
```diff
@@ -126,13 +140,11 @@ export class ContentId implements ValueObject {
...
+}
 
-  toString(): string {
-    return `${this.model} with id '${this.id}'`
-  }
-
-  equals(other: ContentId): boolean {
-    return this.model === other.model && this.id === other.id
+export class AssetId extends ResourceId {
+  constructor(id: string, readonly assetType: string | undefined) {
```
asastre, line 146: Why `AssetId` has the parameters in a different order (reversed order) than `ResourceId` and `ContentId`?
If it's because `assetType` can be `undefined`, why not define it this way?
`constructor(id: string, readonly assetType?: string) {`

If the `assetType` parameter must be declared in the constructor (even if it's undefined), I would change the order of the parameters to match the other classes.

asastre (approved):
Just one small comment, the rest seems ok!

packages/botonic-plugin-contentful/src/cms/callback.ts
dpinol: because in asset we have 2 type: the resourceType is asset and the assetType is eg. image

elozano98 (approved):
Good job! :+1:

PR #1015, opened 2020-10-22
Cli/fix & test templates and examples

dpinol:
* fix imports in cli templates which were broken after 0.14 change in entry files
* add validation of templates and examples in CI
* minor fixes
* rename .js files to .jsx which use react


.github/workflows/botonic-cli-tests.yml
```diff
@@ -30,3 +30,72 @@ jobs:
...
+    runs-on: ubuntu-latest
+    env:
+      PACKAGE: botonic-cli
+    steps:
+      - name: Checking out to current branch
+        uses: actions/checkout@v1
+      - name: Setting up node
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12'
```
vanbasten17, line 44: should we setup it with v10?

packages/botonic-cli/templates/intent/tests/app.test.js
```diff
@@ -13,7 +16,8 @@ const o = new BotonicOutputTester(app)
 test('TEST: (404) NOT FOUND', async () => {
   await expect(i.text('whatever')).resolves.toBe(
     o.text(
-      `Enter the generated JSON key for dialogflowV2 in plugins.js to test the bot. Try typing "hello" to start the bot.`
+      // replace with  'Try typing "hello" to start the bot.' after configuring dialogflow
```
vanbasten17, line 19: did you forget to remove it?

vanbasten17 (approved):
Nice!!! 💯

ericmarcos (approved):
Well done!

.github/workflows/botonic-cli-tests.yml
dpinol: why? 12 is the lts and our target version

packages/botonic-cli/templates/intent/tests/app.test.js
dpinol: no, the expected text was wrong (it was concatenating the normal message and the one which appears when dialogflow is not configure

PR #1017, opened 2020-10-23
fix(react): fix non-boolean attribute test

vanbasten17:
## Description
* Fixed error appearing in some tests: `Received true for a non-boolean attribute markdown`. Caused by styled-components: https://github.com/styled-components/styled-components/issues/1198#issue-262022540
* Updated snapshots


marcpdw (approved):
What's the reason behind the use of a number instead?

dpinol (approved):

PR #1019, opened 2020-10-23
feat(react): add 'storageKey' webchat setting #BOT-133

ericmarcos:
## Description

Add a new webchat setting `storageKey` that indicates what key name to use in order to store the webchat state in localStorage (or sessionStorage).
It can be any "stringable" object, or a function, in that case the return value of the function will be used. If a falsy object is passed then the default value "botonicState" will be used.

Note: I also took the chance to refactor a little bit the "storage" functionality by removing some unnecessary code.

## Context

Some clients need to have different webchat SDKs under the same domain. For example:
* customer.com/help --> Bot about FAQs
* customer.com/product --> Live agent chat for pre-sales

The problem is that localStorage is defined at domain level, so different instances of the webchat SDK would end up sharing the state (history of messages, etc), which is not desirable.
By adding a configurable localstorage key we can now have multiple SDKs without this conflicts.

Just define a 'storageKey' param in wechat settings:
```
const webchat = {
  storageKey: 'myCustomBotonicStateKey'
}
```

or using a function:
```
const webchat = {
  storageKey: () => 'myCustomBotonicStateKey'
}
```

## Testing

The pull request...

- has unit tests

### DIFF
--- a/packages/botonic-react/tests/webchat/storage.test.jsx
+++ b/packages/botonic-react/tests/webchat/storage.test.jsx
@@ -45,3 +45,34 @@ describe('TEST: storage', () => {
     expect(sessionStorage.getItem('botonicState')).toBeNull()
   })
 })
+
+describe('TEST: storageKey', () => {
+  afterEach(() => {
+    localStorage.clear()
+    sessionStorage.clear()
+  })
+
+  it('Stores botonicState in the localStorage key defined in the settings', async () => {
+    await act(async () => {
+      TestRenderer.create(<Webchat storageKey='myCustomKey' />)
+    })
+    expect(localStorage.getItem('botonicState')).toBeNull()
+    expect(localStorage.getItem('myCustomKey')).not.toBeNull()
+  })
+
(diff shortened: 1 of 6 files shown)


vanbasten17 (approved):

packages/botonic-react/tests/webchat/storage.test.jsx
```diff
@@ -45,3 +45,34 @@ describe('TEST: storage', () => {
...
+    expect(localStorage.getItem('botonicState')).toBeNull()
+    expect(localStorage.getItem('myCustomKey')).not.toBeNull()
+  })
+
+  it('Stores botonicState in the default localStorage key if storageKey is null', async () => {
+    await act(async () => {
+      TestRenderer.create(<Webchat storageKey={null} />)
+    })
+    expect(localStorage.getItem('botonicState')).not.toBeNull()
+    expect(localStorage.getItem('myCustomKey')).toBeNull()
```
dpinol, line 76: line not necessary

dpinol (approved):

PR #1020, opened 2020-10-24
Feat/shadow dom #BOT-120

ericmarcos:
**TLDR**

A new feature that avoids webchat CSS conflicts with the host page of your website.
1. Add `shadowDOM: true` to your project `src/webchat/index.js`:
```javascript
export const webchat = {
  shadowDOM: true, // false by default
  theme: {...},
  ...
}
```
2. Replace the `styleLoaderConfig` on your `webpack.config.js` with:
```javascript
const stylesLoaderConfig = {
  test: /\.(scss|css)$/,
  use: [{loader: 'style-loader', options: {
    insert: function(element) {
      if(!window._botonicInsertStyles)
        window._botonicInsertStyles = []
      window._botonicInsertStyles.push(element)
    }
  }}, 'css-loader', 'sass-loader'],
}
```

-------

This PR contains several changes, sorry I couldn't break it up into different PRs but all changes are needed in order to make this work.

#### ShadowDOM 

* **New `shadowDOM` parameter on WebchatApp, DevApp constructors**. This param indicates whether or not to activate the shadowDOM feature. Users will define this param in their `webchat` object (`src/webchat/index.js`):

```javascript
export const webchat = {
  shadowDOM: true, // false by default
  theme: {...},
  ...
}
```

When true, Botonic will render all the webchat inside a shadowDOM node so the webchat styles are isolated from the rest of the host page.

* **New Webpack config for styles**. By default, webpack, and concretely its `style-loader` plugin, adds styles to `document.head`. When using shadowDOM we need to inject these styles in the `shadowRoot` node of our app. We achieve this using the [insert](https://webpack.js.org/loaders/style-loader/#insert) option of `style-loader`, which is a function that receives the styles to be injected. We store these styles in a global variable `window._botonicInsertStyles` because the  shadowRoot element is not yet created when this function is called.

* **Dynamically injecting styles**. We need to inject the styles into the shadowRoot node after it's created (obviously), but before the rest of the webchat component is mounted, otherwise some styles will not be applied properly (like the ones needed by `simplebar`. That's why i do this logic in a new `useComponentWillMount` hook. Also because of `simplebar` (and potentially other deps the developers could add in their projects), even if `shadowDOM: true` styles need to be added **also** in `document.head` as it uses auxiliar DOM elements (invisble nodes that are created/destroyed dynamically at document level to calculate stuff) that need global styles. Another issue is that `styled-components` inject styles at runtime too, so we need to tell the library where to inject them using [StyleSheetManager](https://styled-components.com/docs/api#stylesheetmanager). Last, our dependency `emoji-picker-react` adds styles directly to `head`, so we need to manually find those and inject them in our `shadowRoot`. We could avoid that last part if we used dependencies published as ES6 so imported styles would be injected via Webpack/style-loader ([emoji-mart](https://github.com/missive/emoji-mart) looks like a solid alternative...)

  Note: One of the biggest problems I faced here is that React has a [long-standing bug](https://stackoverflow.com/questions/43261940/render-react-into-shadow-root) that makes it impossible to render an app in a child node of a shadowRoot. It can only be in a shadowRoot directly. This was a problem because I was injecting the styles manually outside the realm of React in the same node where react would mount the app DOM and had a lot of weird behaviour (DOM nodes disappearing, etc). Eventually, `useComponentWillMount` solved the issue as it allowed me to inject the styles inside the React realm.

* **`DebugTabPortal`** is a new component that wraps `DebugTab` (previously `FixedTab`, renamed it for clarity). `DebugTabPortal` renders the "Botonic Dev Console" in the `document.body` regardless of the `shadowDOM` option using a [React portal](https://reactjs.org/docs/portals.html), otherwise it would be rendered inside the shadowRoot node and not look good.

* **`src/webchat/devices` and `src/util/dom` refactors**. Some methods in these files were using `document.getElementById` which doesn't work if our elements are hidden inside a shadowRoot node. So we need to pass the host node as a parameter. I took the chance to simplify some parts of these files.

#### Configurable Host ID and auto-create host element

Added a new `hostId` option to `WebchatApp` so that we know on what element we have to create the `shadowRoot`. By default we normally use "root", but sometimes that's not possible and we use another id, so we need to pass it with this new option. I took the chance to make WebchatApp create the root node if it's not present in the page, and to find it automatically if it's not passed to the render method:

Before:
```javascript
<script src="./webchat.botonic.js">
<div id="root"></div>
<script type="text/javascript">
  document.addEventListener('DOMContentLoaded', function (event) {
    Botonic.render(document.getElementById('root'), {appId: 'XYZ'})
  })
</script>
```

After:
```javascript
<script src="./webchat.botonic.js">
<script type="text/javascript">
  document.addEventListener('DOMContentLoaded', function (event) {
    Botonic.render(null, {appId: 'XYZ'})
  })
</script>
```

I made it backward compatible so the old way is still valid.
The end goal with this is to simplify as much as possible the embed script for the simplest use case, and allow more verbose options for more complex use cases.

Next steps would be:
* Get rid of the `DOMContentLoaded` listener in the embed code, we've seen this failing some times in slow pages where the DOM is already initialized before executing this code. I've made some initial progress in the `WebchatApp` constructor using a new `onDOMLoaded` util.
* Allow the first param of `render` to be the config object instead of a DOM node, so we could do `Botonic.render({appId: 'XYZ'})`
* Inject the appId at deploy time, so we could get rid of the render call altogether in the embed script. So only with `<script src="./webchat.botonic.js">` everything would work.

#### New Tests for WebchatApp and DevApp

I added some very basic tests for `WebchatApp` and `DevApp` to make sure the new options `hostId`, `shadowDOM` and the Portal refactor work as expected.

----------------

# OLD DESCRIPTION:

## Description

Add a new webchat setting `shadowDOM` (boolean) that indicates whether the whole webchat component should be encapsulated in a shadow DOM node (https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM)
If set to true, the webchat styles will be isolated from the rest of the page styles, avoiding conflicts with the host page.
If not set, the default value is false, which is the same behaviour we had before.
A function that returns a boolean is also allowed.

**IMPORTANT**: When this setting is enabled the animations **will not work**. That's because of a [bug in framer-motion](https://github.com/framer/motion/issues/379), the animation library we're using. My opinion is to move forward with this PR, document clearly this limitation and let the developer decide. In the future we can see if we can patch the library or try a different one, or wait till they fix it...

## Context

CSS conflicts with the host websites is a recurring problem with our clients.

Using the library [react-shadow](https://github.com/Wildhoney/ReactShadow)

In your `webchat/index.js`:
```
export const webchat = {
  shadowDOM: true
}
```

or using a boolean function:
```
export const webchat = {
  shadowDOM: () => true
}
```

## Testing

The pull request...

- doesn't need tests because...

### DIFF
--- a/packages/botonic-react/src/util/dom.js
+++ b/packages/botonic-react/src/util/dom.js
@@ -1,15 +1,17 @@
+import { WEBCHAT } from '../constants'
+
 export const getScrollableArea = webchatElement => {
   const getArea = area => {
     const botonicScrollableContent = webchatElement.querySelector(
-      '#botonic-scrollable-content'
+      WEBCHAT.SELECTORS.SCROLLABLE_CONTENT
     )
     const scrollableArea =
       botonicScrollableContent && botonicScrollableContent.querySelector(area)
     return scrollableArea
   }
   return {
-    full: getArea('.simplebar-content'),
-    visible: getArea('.simplebar-content-wrapper'),
+    full: getArea(WEBCHAT.SELECTORS.SIMPLEBAR_CONTENT),
+    visible: getArea(WEBCHAT.SELECTORS.SIMPLEBAR_WRAPPER),
   }
 }
 
@@ -30,17 +32,17 @@ export const scrollToBottom = ({
 }
 
 export const getWebchatElement = host =>
-  host && host.querySelector('#botonic-webchat')
+  host && host.querySelector(`#${WEBCHAT.DEFAULTS.ID}`)
 
 // https://stackoverflow.com/questions/9457891/how-to-detect-if-domcontentloaded-was-fired
-export const onDOMLoaded = f => {
+export const onDOMLoaded = callback => {
   if (/complete|interactive|loaded/.test(document.readyState)) {
     // In case the document has finished parsing, document's readyState will
     // be one of "complete", "interactive" or (non-standard) "loaded".
-    f()
+    callback()
   } else {
     // The document is not ready yet, so wait for the DOMContentLoaded event
-    document.addEventListener('DOMContentLoaded', f, false)
+    document.addEventListener('DOMContentLoaded', callback, false)
   }
 }
 
--- a/packages/botonic-react/src/util/dom.js
+++ b/packages/botonic-react/src/util/dom.js
@@ -43,3 +43,11 @@ export const onDOMLoaded = f => {
     document.addEventListener('DOMContentLoaded', f, false)
   }
 }
+
+export const isShadowDOMSupported = () => {
+  try {
+    return document.head.createShadowRoot || document.head.attachShadow
+  } catch (e) {
+    return false
+  }
+}
--- a/packages/botonic-react/src/util/dom.js
+++ b/packages/botonic-react/src/util/dom.js
@@ -1,23 +1,26 @@
-export const getBotonicScrollableContent = () => {
-  return document.getElementById('botonic-scrollable-content')
-}
-
-export const getScrollableArea = () => {
+export const getScrollableArea = webchatElement => {
   const getArea = area => {
-    const botonicScrollableContent = getBotonicScrollableContent()
+    const botonicScrollableContent = webchatElement.querySelector(
+      '#botonic-scrollable-content'
+    )
     const scrollableArea =
-      botonicScrollableContent &&
-      botonicScrollableContent.getElementsByClassName(area)[0]
+      botonicScrollableContent && botonicScrollableContent.querySelector(area)
     return scrollableArea
   }
   return {
-    full: getArea('simplebar-content'),
-    visible: getArea('simplebar-content-wrapper'),
+    full: getArea('.simplebar-content'),
+    visible: getArea('.simplebar-content-wrapper'),
   }
 }
 
-export const scrollToBottom = ({ timeout = 200, behavior = 'smooth' } = {}) => {
-  const frame = getScrollableArea().visible
+export const scrollToBottom = ({
+  timeout = 200,
+  behavior = 'smooth',
+  host,
+} = {}) => {
+  const webchatElement = getWebchatElement(host)
+  if (!webchatElement) return
+  const frame = getScrollableArea(webchatElement).visible
   if (frame) {
     setTimeout(
       () => frame.scrollTo({ top: frame.scrollHeight, behavior: behavior }),
@@ -26,9 +29,17 @@ export const scrollToBottom = ({ timeout = 200, behavior = 'smooth' } = {}) => {
   }
 }
 
-export const getWebchatElement = () =>
-  document.getElementById('botonic-webchat')
+export const getWebchatElement = host =>
+  host && host.querySelector('#botonic-webchat')
 
-export const setWebchatElementHeight = newHeight => {
-  getWebchatElement().style.height = newHeight
+// https://stackoverflow.com/questions/9457891/how-to-detect-if-domcontentloaded-was-fired
+export const onDOMLoaded = f => {
+  if (/complete|interactive|loaded/.test(document.readyState)) {
+    // In case the document has finished parsing, document's readyState will
+    // be one of "complete", "interactive" or (non-standard) "loaded".
+    f()
+  } else {
+    // The document is not ready yet, so wait for the DOMContentLoaded event
+    document.addEventListener('DOMContentLoaded', f, false)
+  }
 }
--- a/packages/botonic-react/src/webchat-app.jsx
+++ b/packages/botonic-react/src/webchat-app.jsx
@@ -17,6 +17,7 @@ export class WebchatApp {
     enableAttachments,
     enableUserInput,
     enableAnimations,
+    shadowDOM,
     defaultDelay,
     defaultTyping,
     storage,
@@ -36,6 +37,7 @@ export class WebchatApp {
     this.enableAttachments = enableAttachments
     this.enableUserInput = enableUserInput
     this.enableAnimations = enableAnimations
+    this.shadowDOM = shadowDOM
     this.defaultDelay = defaultDelay
     this.defaultTyping = defaultTyping
     this.storage = storage
@@ -199,6 +201,7 @@ export class WebchatApp {
       enableUserInput,
       enableAnimations,
       enableEmojiPicker,
+      shadowDOM,
       defaultDelay,
       defaultTyping,
       storage,
@@ -219,6 +222,7 @@ export class WebchatApp {
     enableAttachments = enableAttachments || this.enableAttachments
     enableUserInput = enableUserInput || this.enableUserInput
     enableAnimations = enableAnimations || this.enableAnimations
+    shadowDOM = shadowDOM || this.shadowDOM
     defaultDelay = defaultDelay || this.defaultDelay
     defaultTyping = defaultTyping || this.defaultTyping
     storage = storage || this.storage
@@ -241,6 +245,7 @@ export class WebchatApp {
         enableAttachments={enableAttachments}
         enableUserInput={enableUserInput}
         enableAnimations={enableAnimations}
+        shadowDOM={shadowDOM}
         storage={storage}
         storageKey={storageKey}
         defaultDelay={defaultDelay}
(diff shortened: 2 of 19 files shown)


vanbasten17 (approved):

marcpdw (approved):

asastre (commented):
What happens if we're importing a CSS file into the `webchat/index.js` to load some CSS? Will this CSS be displayed correctly inside the Shadow DOM?
Will we need to add them in a different way (through webpack as it's noted [here](https://github.com/Wildhoney/ReactShadow))? If so, I think this should be explained in the Botonic documentation.

ericmarcos:
> What happens if we're importing a CSS file into the `webchat/index.js` to load some CSS? Will this CSS be displayed correctly inside the Shadow DOM?
> Will we need to add them in a different way (through webpack as it's noted [here](https://github.com/Wildhoney/ReactShadow))? If so, I think this should be explained in the Botonic documentation.

@asastre that's a very good point. This encapsulation will work for sure if your custom components are built with `styled-components`, however, I didn't test yet other ways to import CSS. Actually, I'm testing now this branch with customMessage types and I'm getting some errors. This PR needs more testing! I'll leave it as NOT MERGE for now.

packages/botonic-react/src/util/dom.js
```diff
@@ -1,23 +1,26 @@
...
-      botonicScrollableContent &&
-      botonicScrollableContent.getElementsByClassName(area)[0]
+      botonicScrollableContent && botonicScrollableContent.querySelector(area)
     return scrollableArea
   }
   return {
-    full: getArea('simplebar-content'),
-    visible: getArea('simplebar-content-wrapper'),
+    full: getArea('.simplebar-content'),
+    visible: getArea('.simplebar-content-wrapper'),
```
vanbasten17, line 12: should we add these references as constants? also `@botonic-scrollable-content`, etc.

packages/botonic-react/src/util/dom.js
```diff
@@ -26,9 +29,17 @@ export const scrollToBottom = ({ timeout = 200, behavior = 'smooth' } = {}) => {
...
-export const getWebchatElement = () =>
-  document.getElementById('botonic-webchat')
+export const getWebchatElement = host =>
+  host && host.querySelector('#botonic-webchat')
 
-export const setWebchatElementHeight = newHeight => {
-  getWebchatElement().style.height = newHeight
+// https://stackoverflow.com/questions/9457891/how-to-detect-if-domcontentloaded-was-fired
+export const onDOMLoaded = f => {
+  if (/complete|interactive|loaded/.test(document.readyState)) {
```
vanbasten17, line 37: use `document.readyState.exec(/complete|interactive|loaded/)` instead?

packages/botonic-react/src/util/dom.js
```diff
@@ -26,9 +29,17 @@ export const scrollToBottom = ({ timeout = 200, behavior = 'smooth' } = {}) => {
...
+    // In case the document has finished parsing, document's readyState will
+    // be one of "complete", "interactive" or (non-standard) "loaded".
+    f()
+  } else {
+    // The document is not ready yet, so wait for the DOMContentLoaded event
+    document.addEventListener('DOMContentLoaded', f, false)
```
vanbasten17, line 43: give an explicit name to f? if refers to fire()?

packages/botonic-react/src/webchat/devices/scrollbar-controller.js
```diff
@@ -57,25 +56,23 @@ export class ScrollbarController {
...
 
   toggleOnMouseWheelEvents() {
-    const webchat = getWebchatElement()
     if (this.hasScrollbar()) {
-      webchat.onmousewheel = {}
+      this.webchat.onmousewheel = {}
       return
     }
-    webchat.onmousewheel = e => e.preventDefault()
+    this.webchat.onmousewheel = e => e.preventDefault()
```
vanbasten17, line 63: we should test these changes in the simulator, will try to do it in an iOS simulator 👍

vanbasten17 (approved):
GAS!! 🚀

vanbasten17:
I will try to use this in some botonic-examples (as the styles were conflicting with the ones from docusaurus)

packages/botonic-react/src/util/dom.js
```diff
@@ -26,9 +29,17 @@ export const scrollToBottom = ({ timeout = 200, behavior = 'smooth' } = {}) => {
   }
 }
 
-export const getWebchatElement = () =>
-  document.getElementById('botonic-webchat')
+export const getWebchatElement = host =>
+  host && host.querySelector('#botonic-webchat')
```
asastre, line 33: Again here, as @vanbasten17 said, I think it would be nice to have all this ids and classes as constants.

packages/botonic-react/src/webchat-app.jsx
```diff
@@ -47,6 +52,36 @@ export class WebchatApp {
...
+  createRootElement() {
+    // Create root element <div id='root'> if not exists
+    // Create shadowDOM to root element if needed
+    let host = document.getElementById(this.hostId)
+    if (!host) {
+      host = document.createElement('div')
+      host.id = this.hostId
+      if (document.body.firstChild)
+        document.body.insertBefore(host, document.body.firstChild)
+      else document.body.appendChild(host)
```
asastre, line 77: Why inserting the root element as the first child? Why not doing only `document.body.appendChild(host)`?

packages/botonic-react/src/webchat/header.jsx
```diff
@@ -19,7 +19,7 @@ const Header = styled(Flex)`
   z-index: 2;
 `
 
-const ImageContainer = styled(Flex)`
+const ImageContainer = styled.div`
```
asastre, line 22: Is `display: flex;` missing here?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -279,6 +285,35 @@ export const Webchat = forwardRef((props, ref) => {
   const resendUnsentInputs = async () =>
     props.resendUnsentInputs && props.resendUnsentInputs()
 
+  // Load styles stored in window._botonicInsertStyles by Webpack
+  useComponentWillMount(() => {
+    if (!window._botonicInsertStyles || !window._botonicInsertStyles.length)
+      return
```
asastre, line 291: Even if we don't have styles in `_botonicInsertStyles` we want to load the `emoji-picker-react` styles, right? If so, this early return should be placed after injecting the emoji-picker-react styles.

packages/botonic-react/tests/webchat/webchat.addmessages.test.jsx
```diff
@@ -24,7 +24,7 @@ describe('Adding webchat messageComponent', () => {
       )
     })
     await act(async () => {
-      render(<Webchat webchatHooks={result.current} />)
+      render(<Webchat host={document.body} webchatHooks={result.current} />)
```
asastre, line 27: Why you have to add this `host` prop here?

asastre (changes requested):
Amazing work adding the shadowDOM feature! :100: :100: :clap:!! 
Just some comments/questions before being approved :wink:

packages/botonic-react/src/util/dom.js
ericmarcos: it's the same no?

packages/botonic-react/src/util/dom.js
ericmarcos: f = function, but yeah, maybe I'll put a more descriptive name

packages/botonic-react/src/webchat-app.jsx
ericmarcos: It's just a cosmetic thing: to put the root tag at the top of the body, so it's easy to find.

packages/botonic-react/src/webchat/header.jsx
ericmarcos: I think it's not necessary, but I'll review

packages/botonic-react/src/webchat/webchat.jsx
ericmarcos: true... maybe I'll just return early if `shadowDOM` is false

packages/botonic-react/tests/webchat/webchat.addmessages.test.jsx
ericmarcos: because now utils functions need the host as a parameter, otherwise tests would fail. Maybe I'll just add `document.body` as the default host inside Webchat component...

PR #1022, opened 2020-10-26
refactor(contentful): CMS.(top)contents to use generics

dpinol:
CMS.contents & CMS.topContents now use generics so that now they can return the specific subtype of Content desired

### DIFF
--- a/packages/botonic-plugin-contentful/src/cms/cms-error.ts
+++ b/packages/botonic-plugin-contentful/src/cms/cms-error.ts
@@ -111,14 +111,14 @@ export class ErrorReportingCMS implements CMS {
       .catch(this.handleError('contentsWithKeywords', context))
   }
 
-  topContents(
+  topContents<T extends TopContent>(
     model: TopContentType,
     context?: Context,
     filter?: (cf: CommonFields) => boolean,
     paging?: PagingOptions
-  ): Promise<TopContent[]> {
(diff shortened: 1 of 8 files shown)


packages/botonic-plugin-contentful/src/cms/cms-error.ts
```diff
@@ -111,14 +111,14 @@ export class ErrorReportingCMS implements CMS {
...
+  topContents<T extends TopContent>(
     model: TopContentType,
     context?: Context,
     filter?: (cf: CommonFields) => boolean,
     paging?: PagingOptions
-  ): Promise<TopContent[]> {
+  ): Promise<T[]> {
     return this.cms
-      .topContents(model, context, filter, paging)
+      .topContents<T>(model, context, filter, paging)
```
asastre, line 121: Is it necessary to add here the `<T>`?

asastre (approved):
Nice :clap:!!

packages/botonic-plugin-contentful/src/cms/cms-error.ts
dpinol: yes, otherwise TS will complain that the function is declared as returning T[], but it returns TopContents[]

elozano98 (approved):
Good job! :+1:

PR #1036, opened 2020-10-27
Contentful/multiple field update

dpinol:
## Description
Now the csv importer of texts coming from content translator groups all the fields from the same content. Advantages:
* It enables updating all the fields for the same content in a single API call to the CMS. In this PR we have modified the ManageCms interface to be able to update multiple fields in a single call.
* In the future it will be easier to implement the feature of renaming the content name/code field
* In a future PR we will detect when contents don't make sense for the imported locale. In this case, we'll remove the references (buttons) leading to this content for this particular locale.

## Context

We want to be able to implement the last of the 3 points above.

We have modified the ManageCms interface to be able to update multiple content fields in a single call.

## Testing

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/tools/l10n/csv-import.ts
+++ b/packages/botonic-plugin-contentful/src/tools/l10n/csv-import.ts
@@ -1,22 +1,20 @@
 import { ManageCms } from '../../manage-cms/manage-cms'
 import * as cms from '../../cms'
-import {
-  BotonicContentType,
-  CMS,
-  CmsException,
-  ContentId,
-  ContentType,
-} from '../../cms'
+import { CMS, ContentId, ContentType } from '../../cms'
 import * as fs from 'fs'
 import parser from 'csv-parse'
-import { isOfType } from '../../util/enums'
-import { CONTENT_FIELDS, ContentFieldType } from '../../manage-cms/fields'
+import {
+  CONTENT_FIELDS,
+  ContentField,
+  ContentFieldType,
+  ContentFieldValueType,
+  FIELDS_PER_CONTENT_TYPE,
+} from '../../manage-cms/fields'
 import { ManageContext } from '../../manage-cms/manage-context'
+import { isOfType } from '../../util/enums'
+import { BotonicContentType } from '../../cms/cms'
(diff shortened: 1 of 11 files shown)


guinii (approved):

packages/botonic-plugin-contentful/src/tools/l10n/csv-import.ts
```diff
@@ -1,22 +1,20 @@
 import { ManageCms } from '../../manage-cms/manage-cms'
 import * as cms from '../../cms'
-import {
-  BotonicContentType,
-  CMS,
-  CmsException,
-  ContentId,
-  ContentType,
-} from '../../cms'
+import { CMS, ContentId, ContentType } from '../../cms'
```
elozano98, line 3: Multiple imports from `../../cm`. Maybe `CMS`, `ContentId` and `ContentType` are the only ones needed.

packages/botonic-plugin-contentful/tests/tools/l10n/csv-import.test.ts
```diff
@@ -8,11 +8,11 @@ import { anything, instance, mock, when } from 'ts-mockito'
 import { testManageContentful } from '../../contentful/manage/manage-contentful.helper'
 import { Locale, SPANISH } from '../../../src/nlp'
 import * as cms from '../../../src/cms'
+import { ContentId, ContentType } from '../../../src/cms'
```
elozano98, line 11: Multiple imports from `../../../src/cm`. Maybe `ContentId` and `ContentType` are the only ones needed.

elozano98 (approved):
Good job! :+1:

packages/botonic-plugin-contentful/src/tools/l10n/csv-import.ts
dpinol: @elozano98 done in future PR after refactor. thanks!

packages/botonic-plugin-contentful/tests/tools/l10n/csv-import.test.ts
dpinol: @elozano98 done in future PR after refactor. thanks!

PR #1038, opened 2020-10-29
fix(react): replace react-storage-hooks dep (iOS versions not working fix)

vanbasten17:
https://linear.app/hubtype/issue/BOT-142/investigate-issues-with-ios-versions

## Description
* Replace `react-storage-hooks` for `react-use-storage`. 

## Context

iOS versions 13.4, 13.5, 13.6 and 13.7 where not working since something wrong in this external library (was entering in an infinite render loop). Now they are working as expected.

Created a custom `useStorageState` hook using this new lib. Created this hook into a separate file instead of `hooks.js` because it was conflicting with node build: `CustomEvent is not defined`, as bots were not loading `/dist/bot.js` as expected

Just update iOS supported versions in docs. 

## Testing

The pull request...

- has unit tests


packages/botonic-react/src/webchat/use-storage-state-hook.js
```diff
@@ -0,0 +1,9 @@
+import { useLocalStorage, useSessionStorage } from 'react-use-storage'
+
+export function useStorageState(storage, storageKey) {
+  const [botonicLocalState, saveLocalState] = useLocalStorage(storageKey)
```
dpinol, line 4: can you move these 2 lines right before they're used? or they're hooks that cannot be within "ifs"?

dpinol (approved):

asastre (approved):
LGTM! :rocket: :rocket: :rocket:

PR #1039, opened 2020-10-29
fix(react): fix missing storageKey in constructor

vanbasten17:
## Description

Add missing `storageKey` attribute in constructor.

## Context
`storageKey` was missing in `WebchatApp` constructor. I think it could be working when calling it from `index.html` but not when calling directly inside `webchat/index.js`.


ericmarcos (approved):
Absolutely!
GAS

asastre (approved):

PR #1041, opened 2020-11-01
fix(core/react): make sending unsent inputs more consistent

vanbasten17:
## Description
Be more consistent when resending inputs. 
* Filter now takes into account that `unsentInput` comes as a message property, not only the ack set to 0.
* Explicitly handle unsent messages by adding the ack 0.
* Once the message is updated remove unnecessary `unsentInput` property from a message.

## Context
We were having lots of these messages in the server: `Exception: 'message' not found in imp event data: ...`... This happened when in fact a message was sent properly but due to some race condition the `ack` continued being set to 0 and had no `unsentInput` property.

**E.g:**
```
for (const message of this.unsentInputs()) { // this.unsentInputs() --> Sometimes returns some messages without unsentInput property
      await this.postMessage(this.user, message.unsentInput) // --> was being called with this.postMessage(this.user, undefined), hence the error.
}
```

### DIFF
--- a/packages/botonic-core/src/hubtype-service.js
+++ b/packages/botonic-core/src/hubtype-service.js
@@ -98,7 +98,7 @@ export class HubtypeService {
   handleUnsentInput(message) {
     this.onEvent({
       action: 'update_message_info',
-      message: { id: message.id, unsentInput: message },
+      message: { id: message.id, ack: 0, unsentInput: message },
     })
   }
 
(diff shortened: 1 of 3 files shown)


ericmarcos (approved):
Nice catch!

asastre (approved):
:clap:

dpinol (approved):
merge!

PR #1067, opened 2020-11-05
Refactor/utils

vanbasten17:
## Description
Divide utils for maintainability:
* dom: utilities that access directly dom elements
* environment: functions related to webchat's execution environment
* logs: logging
* objects: JS objects manipulation
* react: rendering, wrappers, ...
* regexs: regex operations

## Context
Split 'utils' into several files depending on nature of utils. For maintainability and scalability.


packages/botonic-react/src/util/environment.js
```diff
@@ -0,0 +1,27 @@
...
+export const staticAsset = path => {
+  try {
+    const scriptBaseURL = document
+      .querySelector('script[src$="webchat.botonic.js"]')
+      .getAttribute('src')
+    const scriptName = scriptBaseURL.split('/').pop()
+    const basePath = scriptBaseURL.replace('/' + scriptName, '/')
+    return basePath + path
+  } catch (e) {
+    return path
```
dpinol, line 13: we should never have silent catches without a comment on why we do it

packages/botonic-react/src/util/environment.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export const resolveImage = src => {
+  if (isURL(src)) return src
+  return staticAsset(src)
+}
+
+export const isURL = urlPath => {
+  // @stephenhay (38 chars) from: https://mathiasbynens.be/demo/url-regex
+  const URL_PATTERN = /^(https?|ftp):\/\/[^\s/$.?#].[^\s]*$/
```
dpinol, line 24: imho I only define local constants when their name helps us to understand what they do. here it's obvious

packages/botonic-react/src/components/button.jsx
```diff
@@ -4,7 +4,7 @@ import styled from 'styled-components'
 import { params2queryString, INPUT } from '@botonic/core'
```
dpinol, line 4: in the future "the price to pay" to add an util is writing a unit test for them.
this will ensure that we don't create too many :-)

packages/botonic-react/src/util/regexs.js
```diff
@@ -0,0 +1,18 @@
+export const serializeRegex = (_, value) => {
+  // To be used as a 2nd argument (replacer) of JSON.stringify
```
dpinol, line 2: it doesn't make sense to export a util which nobody uses outside this file
I would move it as a lambda within stringifyWithRegexs

packages/botonic-react/src/util/regexs.js
```diff
@@ -0,0 +1,18 @@
...
+
+export const deserializeRegex = regexStr => {
+  /* eslint-disable no-useless-escape */
+  const fragments = regexStr.match(/\/(.[^\/]+)\/([a-z]{1})?$/i) // https://stackoverflow.com/a/33642463, https://github.com/hubtype/botonic/pull/805#discussion_r437309074
+  const deserialized = new RegExp(fragments[1], fragments[2] || '')
+  return deserialized
+}
+
+export const stringifyWithRegexs = object => {
+  // Serialization of regexs: https://stackoverflow.com/questions/12075927/serialization-of-regexp
```
dpinol, line 16: documentation should be with /** */ before the class/function first line
Serialization of regexs -> Serialization of objects containing regexs ?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -58,11 +55,35 @@ import {
...
+  if (splittedAction.length <= 1) return undefined
+  return JSON.parse(splittedAction[1])
+}
+
+/*
+ * returns the value of a property defined in bot's theme based on WEBCHAT.CUSTOM_PROPERTIES dictionary.
+ * it gives preference to nested defined properties (e.g.: header.style) over plain properties (e.g.: headerStyle).
+ * if property doesn't exist, returns the defaultValue
+ */
+export const _getThemeProperty = theme => (
```
dpinol, line 72: no export. No _

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -58,11 +55,35 @@ import {
...
 
-import merge from 'lodash.merge'
-import { useAsyncEffect } from 'use-async-effect'
+export const getParsedAction = botonicAction => {
+  const splittedAction = botonicAction.split('create_case:')
+  if (splittedAction.length <= 1) return undefined
+  return JSON.parse(splittedAction[1])
+}
+
+/*
```
dpinol, line 67: /* is not jsdoc https://jsdoc.app/about-getting-started.html
Start comments with Capital

ericmarcos (approved):

dpinol (changes requested):
:clap:

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: export because it is used for tests, _ to distinguish it from local scoped getThemeProperty

packages/botonic-react/src/util/environment.js
vanbasten17: added console.error 👍

packages/botonic-react/src/components/button.jsx
vanbasten17: sure, will add them in next PRs 👍

dpinol (approved):

PR #1070, opened 2020-11-06
Fix Issue: Params request fail in Oauth 3.1.0+

Khaeshah:
## Description
Fix calls to `/o/token` to follow current oauth specs
## Context

In oauth version 3.1.0 the bug that allowed making a post query with params was [fixed](https://github.com/oauthlib/oauthlib/pull/667). This caused a botonic login fail in this new package version in our backend, as it was using params.

It is retrocompatible as a middleware has been added in our backend to prevent that case.

Now we're using data instead of params


vanbasten17 (approved):
LGTM, i will assure that this is bwd compatible, ask me for help by testing it 👍

ericmarcos (approved):

PR #1072, opened 2020-11-09
fix(core): router

AinaVendrell:
## Description
Fix a bug in getting child routes that was added in https://github.com/hubtype/botonic/pull/903. When the bot was receiving a payload like : payload: '__PATH __PAYLOAD_2' and also had a lastRoute the lastRoute wasn't being checked and it wasn't able to find the route.

## Testing

The pull request...

- has unit tests

### DIFF
--- a/packages/botonic-core/src/router.js
+++ b/packages/botonic-core/src/router.js
@@ -10,16 +10,17 @@ export class Router {
 
   // eslint-disable-next-line complexity
   processInput(input, session = {}, lastRoutePath = null) {
-    let routeParams = undefined
+    let routeParams = {}
     if (input.payload && input.payload.includes('__PATH_PAYLOAD__')) {
       const pathParam = input.payload.split('__PATH_PAYLOAD__')
-      const route = this.getRouteByPath(pathParam[1], this.routes)
-      routeParams = { route }
+      routeParams.route = this.getRouteByPath(pathParam[1], this.routes)
     }
     const pathParams = this.getOnFinishParams(input)
     let brokenFlow = false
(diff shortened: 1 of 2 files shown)


vanbasten17 (approved):
LGTM 👍 we would be able to do more changes when we have improved the tests a little more

packages/botonic-core/src/router.test.js
```diff
@@ -225,6 +225,14 @@ describe('Process input (v<0.9)', () => {
       'Flow1.2'
     )
   })
+  test('old protocol: path payload input, 2nd level route with lastRoutePath', () => {
```
dpinol, line 228: please extend the PR description to explain what was the bug.
Also, I don't understand where is lastRoutePath used here

dpinol (changes requested):

packages/botonic-core/src/router.js
```diff
@@ -149,34 +150,24 @@ export class Router {
...
-          if (previousValue) currentValue = previousValue.concat('/', value)
-          const match = this.matchRoute(r, key, currentValue, input, session)
-          if (!match && r.childRoutes && !childRoute) {
-            childRoute = this.getRoute(
-              input,
-              r.childRoutes,
-              session,
-              currentValue
-            )
-          }
```
asastre, line 174: Are we losing something removing this code?

AinaVendrell: I added this in the old PR because I thought it was necessary but it was not.

asastre (approved):

PR #1086, opened 2020-11-09
feat(core): add new request matcher, fix minor issues

vanbasten17:
## Description
* Added new `request` matcher
* Check that `route.path` exists, causing 'cannot read path of undefined', failing with https://github.com/hubtype/botonic/pull/1072

To use this new matcher:
```javascript
{
    path: 'hi',
    request: request =>
      request.input.data === 'hi' && request.session.organization === '',
    action: Hi,
  },
```
## Testing

The pull request...

**- has unit tests**

### DIFF
--- a/packages/botonic-core/src/index.d.ts
+++ b/packages/botonic-core/src/index.d.ts
@@ -119,15 +119,18 @@ export interface Session {
   user: SessionUser
 }
 
-type StringMatcher = RegExp | string | ((data: string) => boolean)
+type InputMatcher = (input: Input) => boolean
 type ParamsMatcher =
   | { [key: string]: string }
   | ((params: { [key: string]: string }) => boolean)
 type SessionMatcher = (session: Session) => boolean
-type InputMatcher = (input: Input) => boolean
+type RequestMatcher = (input: Input, session: Session) => boolean
+type StringMatcher = RegExp | string | ((data: string) => boolean)
+
 type RouteMatcher =
   | InputMatcher
   | ParamsMatcher
+  | RequestMatcher
   | SessionMatcher
   | StringMatcher
 
--- a/packages/botonic-core/src/index.d.ts
+++ b/packages/botonic-core/src/index.d.ts
@@ -124,7 +124,7 @@ type ParamsMatcher =
   | { [key: string]: string }
   | ((params: { [key: string]: string }) => boolean)
 type SessionMatcher = (session: Session) => boolean
-type RequestMatcher = (input: Input, session: Session) => boolean
+type RequestMatcher = (request: BotRequest) => boolean
 type StringMatcher = RegExp | string | ((data: string) => boolean)
 
 type RouteMatcher =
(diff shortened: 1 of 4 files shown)


packages/botonic-core/src/index.d.ts
```diff
@@ -119,15 +119,18 @@ export interface Session {
...
 }
 
-type StringMatcher = RegExp | string | ((data: string) => boolean)
+type InputMatcher = (input: Input) => boolean
 type ParamsMatcher =
   | { [key: string]: string }
   | ((params: { [key: string]: string }) => boolean)
 type SessionMatcher = (session: Session) => boolean
-type InputMatcher = (input: Input) => boolean
+type RequestMatcher = (input: Input, session: Session) => boolean
```
asastre, line 127: Since we should be able here to work with all the request params I would go here with:
`type RequestMatcher = (request: BotRequest) => boolean`
To be able to do so [this](https://github.com/hubtype/botonic/blob/master/packages/botonic-core/src/core-bot.js#L69) should be changed to:
`[...(await this.routes({ input, session, lastRoutePath })), ...this.defaultRoutes],`

If the changes mentioned above are too many, at least I would pass the parameters as an object: { input: Input, session: Session }

dpinol, line 127: add 
`request?: RequestMatcher`
to the Route type

packages/botonic-core/src/router.js
```diff
@@ -205,6 +205,7 @@ export class Router {
       if (input.type == 'text') value = input.data
     } else if (prop == 'input') value = input
     else if (prop == 'session') value = session
+    else if (prop == 'request') value = { input, session }
```
asastre, line 208: If we do the changes explained in another comment, should we add here the `lastRoutePath`?

dpinol (changes requested):

packages/botonic-core/src/router.js
```diff
@@ -47,7 +47,7 @@ export class Router {
       //get in childRoute if one has path ''
       let defaultAction
       if (
-        !routeParams.route.path &&
+        !(routeParams.route && routeParams.route.path) &&
```
asastre, line 50: You're already checking `routeParams.route` in the next line.
Maybe you can add this check at the beginning instead of doing it in every `if`?

asastre (changes requested):

packages/botonic-core/src/index.d.ts
vanbasten17: added 👍

vanbasten17: i agree with you, i have modified the code in order to pass also the `lastRoutePath`, good catch 👍

packages/botonic-core/src/router.js
vanbasten17: sure, now simplified 💯

dpinol (approved):

asastre (approved):
:rocket:

PR #1087, opened 2020-11-09
React/error boundary

dpinol:
## Description

We want that if CustomMessage crashes, it will be replaced by the specified component (typically a simple error message). So far the solution only works if the failure occurs while deserializing the component. Pending to understand why it does not working yet if it fails during rendering of a new component.

## Context
CustomMessage can fail if there's a bug in their implementation. A typical case is when restoring a serialized component from the browser local storage.

React [ErrorBoundary](https://reactjs.org/docs/error-boundaries.html)

## Testing

The pull request...

- has integration tests

### DIFF
--- a/packages/botonic-react/src/components/components.js
+++ b/packages/botonic-react/src/components/components.js
@@ -1,0 +1,17 @@
+/**
+ * Get the name of an element (instantiated component)
+ * @param {React.Element} component
+ * @returns {string}
+ */
+export function getElementName(component) {
+  return component.type.name
+}
+
+/**
+ * Get the name of a component TYPE
+ * @param {React.ComponentType} component
+ * @returns {string}
+ */
+export function getComponentTypeName(component) {
+  return component.displayName || component.name || 'Component'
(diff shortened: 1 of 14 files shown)


packages/botonic-react/src/components/components.js
```diff
@@ -0,0 +1,17 @@
...
+}
+
+/**
+ * Get the name of a component TYPE
+ * @param {React.ComponentType} component
+ * @returns {string}
+ */
+export function getComponentTypeName(component) {
+  return component.displayName || component.name || 'Component'
+}
```
vanbasten17, line 17: move to `util/components.js`?

ericmarcos (approved):
This is so cool! 👏👏👏

vanbasten17 (approved):
💯 💯 💯

packages/botonic-react/src/components/components.js
dpinol: will do soon

PR #1091, opened 2020-11-11
docs: add gallery of Botonic examples 

vanbasten17:
## Description
* Add first approach for gallery/examples page.
* Fixed check of first render in botonic-react

## Important limitations
* When opening a specific example, Docusaurus styles are conflicting with the ones of the bot. To solve this issue, and after trying to isolate the dom element with `react-shadow`, `react-shadow-wrapper`, etc. I have come up with by simply removing the styles. This has as a handicap that we cannot inject the the navigation bar inside that page.
Also tried to take only the necessary styles from the navbar with this [extension](https://chrome.google.com/webstore/detail/css-used/cdopjfddjlonogibjahpnmjpoangjfff?hl=en) and load them after, with no successful results. To remove this styles when the site is built, we need to add the `styles.css` [filename generated](https://github.com/hubtype/botonic/pull/1091/commits/79e712c107f82458597e586835f3c081a5cfb85d). This name is **deterministic** depending on the styles.css file, so changing it implies to change again the name in the code atm.
* In order to built the website, you will see the usage of `typeof window !== undefined`, this is due to the fact that docusaurus does a build for both server and client side, hence these parts of logic.

## Example

Screenshot 2020-11-18 at 09 34 53


marcpdw (commented):
Code very neat for now. Good job!

docs/website/src/components/examples.js
```diff
@@ -0,0 +1,143 @@
...
+  </h1>
+)
+
+export const ExamplesSection = ({ cards }) => (
+  <section id='examples-section'>
+    <div className='max-w-6xl mx-auto px-4 sm:px-6 relative'>
+      <div className='relative pt-32 pb-10 md:pt-24 md:pb-16'>
+        <div className='w-full text-center'>
+          <ExamplesHeader />
+          <div className='max-w-3xl mx-auto'></div>
```
ericmarcos, line 136: Remove this empty div?

docs/website/src/components/examples.js
```diff
@@ -0,0 +1,143 @@
...
+    <GithubIcon
+      alt='View Source'
+      class='float-right mx-2 h-8 w-8 fill-current'
+    />
+  </a>
+)
+
+const createRootElement = () => {
+  const root = document.createElement('div')
+  const id = 'root'
```
ericmarcos, line 23: id='root' is usually for the main React app in the page. In this case it would be the docusaurus app, but I guess they use another name (otherwise you'd get an error in the console).
I prefer to change this name to something more custom like 'webchatExampleRoot'

vanbasten17: sure, i would change it or generate it randomly to avoid this kind of potential issues

docs/website/src/pages/examples/botonic-example.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+    removejscssfile('styles.css', 'css') // Dev
+    removejscssfile('styles.9c057c15.css', 'css') // Prod
+    window.onload = () => {
+      setTimeout(() => {
+        // eslint-disable-next-line no-undef
+        Botonic &&
+          // eslint-disable-next-line no-undef
+          Botonic.render(document.getElementById(rootId), runtimeOptions)
+      }, 500)
```
marcpdw, line 34: Could this timeout fail in any case? Why half a second? I guess we're simulating a `$(document).ready()` here, right? Maybe `0` would be enough?

docs/website/src/pages/examples/botonic-example.js
```diff
@@ -0,0 +1,63 @@
...
+        if (
+          allsuspects[i] &&
+          allsuspects[i].getAttribute(targetattr) != null &&
+          allsuspects[i].getAttribute(targetattr).indexOf(filename) != -1
+        )
+          allsuspects[i].parentNode.removeChild(allsuspects[i]) //remove element by calling parentNode.removeChild()
+      }
+    }
+    removejscssfile('styles.css', 'css') // Dev
+    removejscssfile('styles.9c057c15.css', 'css') // Prod
```
marcpdw, line 27: There's a lot happening here? Could it be encapsulated to a function with a defined scope?

marcpdw (commented):
Still good! I've added a few comments/questions to consider

docs/website/src/pages/examples/botonic-example.js
MarcosCA: good question

docs/website/static/css/main.css
```diff
@@ -487,7 +487,6 @@ blockquote {
 }
 
 .homeContainer .homeWrapper {
```
MarcosCA, line 489: can be just deleted?

MarcosCA (approved):

docs/website/src/components/examples.js
```diff
@@ -0,0 +1,104 @@
+/* eslint-disable import/no-unresolved */
+/* eslint-disable node/no-missing-import */
```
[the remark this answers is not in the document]
vanbasten17: these are all the imports of type `@docusaurus/module` like `@docusaurus/Head`. It seems that for some reason they don't have these types defined, so the linter complains. Will try to remove all of them when docusaurus 2 is more mature

docs/website/src/pages/examples/botonic-example.js
vanbasten17: i agree, at first i've tried to extract it into another file, but had issues to run the build. I will try to take it a look again 👍

docs/website/src/pages/examples/botonic-example.js
vanbasten17: yes, at first it was defined to 0, but it seems that sometimes 0 is not enough (Botonic script is not ready to use), so I have updated it to half a second. I have tested this in the production build and seems to work as expected now.

docs/website/src/pages/examples/botonic-example.js
vanbasten17: moved, now working 👍

annehubtype (approved):

PR #1092, opened 2020-11-11
fix(react): do not call onClose on very first render

vanbasten17:
## Description
Do not call onClose event on the first app render.

## Context
The first time the app is rendered and the webchat is closed, the `onClose` event is triggered but in fact we are not closing it. Adding a solution to this in order to only call it when the close button is clicked.

### DIFF
--- a/packages/botonic-react/src/webchat/webchat.jsx
+++ b/packages/botonic-react/src/webchat/webchat.jsx
@@ -214,6 +214,7 @@ export const Webchat = forwardRef((props, ref) => {
     setCurrentAttachment,
     // eslint-disable-next-line react-hooks/rules-of-hooks
   } = props.webchatHooks || useWebchat()
+  const firstUpdate = useRef(true)
   const theme = merge(webchatState.theme, props.theme)
   const { initialSession, initialDevSettings, onStateChange } = props
   const isOnline = useNetwork()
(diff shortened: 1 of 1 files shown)


packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -647,8 +648,12 @@ export const Webchat = forwardRef((props, ref) => {
   }
 
   useEffect(() => {
+    if (firstUpdate.current) {
+      firstUpdate.current = false
+      return
+    }
     if (webchatState.isWebchatOpen && props.onOpen) props.onOpen()
-    if (!webchatState.isWebchatOpen && props.onClose) {
+    if (!webchatState.isWebchatOpen && props.onClose && !firstUpdate.current) {
```
dpinol, line 656: do you need to check firstUpdate.current again?

dpinol (approved):

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: nope, not necessary, removed 👍

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -214,6 +214,7 @@ export const Webchat = forwardRef((props, ref) => {
     setCurrentAttachment,
     // eslint-disable-next-line react-hooks/rules-of-hooks
   } = props.webchatHooks || useWebchat()
+  const firstUpdate = useRef(true)
```
ericmarcos, line 217: Why are you using `useRef`? Wouldn't `useState` be more appropiate in this case?

vanbasten17: here the suggested solution: https://medium.com/anna-coding/the-way-to-check-if-its-the-first-time-for-useeffect-function-is-being-run-in-react-hooks-170520554067

ericmarcos (approved):

PR #1118, opened 2020-11-19
Contentful/cms info

dpinol:
* The method to get a contentful space's default locale has been moved to a new CmsInfo class
* New method to get all the locales of a contentful space


guinii (approved):

vanbasten17 (approved):

packages/botonic-plugin-contentful/src/contentful/delivery/cache.ts
```diff
@@ -31,5 +37,6 @@ export class CachedClientApi implements ReducedClientApi {
     this.getEntry = memoize(client.getEntry, options(2))
     this.getContentType = memoize(client.getContentType, options(1))
     this.getContentTypes = memoize(client.getContentTypes, options(0))
+    this.getLocales = memoize(client.getContentTypes, options(0))
```
asastre, line 40: This should be `client.getLocales` instead of `client.getContentTypes`, right?

asastre (changes requested):
Just one correction to be made

packages/botonic-plugin-contentful/src/contentful/delivery/cache.ts
dpinol: right! remove because it's now only called from CmsInfo
I added unit tests for CmsInfo

asastre (approved):

PR #1133, opened 2020-11-23
feat(react): multichannel Facebook text converter

AinaVendrell:
## Description

Split the text when the provider is Facebook  and the number of characters is larger than 640

## Context
Facebook messenger allows a maximum of 640 characters per message

## Testing

The pull request...

- has unit tests

### DIFF
--- a/packages/botonic-react/src/components/multichannel/multichannel-text.jsx
+++ b/packages/botonic-react/src/components/multichannel/multichannel-text.jsx
@@ -2,6 +2,7 @@ import React, { useContext } from 'react'
 
 import { RequestContext } from '../../contexts'
 import { Text } from '../text'
+import { MultichannelFacebook } from './facebook/facebook'
 import { MultichannelButton } from './multichannel-button'
 import { MultichannelContext } from './multichannel-context'
 import {
@@ -9,6 +10,7 @@ import {
   elementHasUrl,
   getMultichannelButtons,
   getMultichannelReplies,
+  isFacebook,
   isMultichannelButton,
   isMultichannelReply,
   isWhatsapp,
@@ -21,17 +23,19 @@ export const MultichannelText = props => {
 
   let elements = []
 
-  const getText = () => {
-    let text = undefined
-    if (typeof props.children == 'string') {
-      text = props.children
-    } else if (Array.isArray(props.children)) {
-      text = props.children[0]
-    }
+  const getText = children => {
+    children = Array.isArray(children) ? children : [children]
+    const text = children
+      .filter(e => e && !e.type)
+      .map(e => {
+        if (Array.isArray(e)) return getText(e)
+        else return String(e)
+      })
+      .join('')
     if (text == undefined) {
       return []
     }
-    return [text]
+    return [text].filter(t => t !== '') // to avoid line breaks when the carousel doesn't have title or subtitle
   }
 
   const getButtonsAndReplies = () =>
@@ -88,6 +92,26 @@ export const MultichannelText = props => {
       <Text {...MULTICHANNEL_WHATSAPP_PROPS} {...props}>
         {elements}
       </Text>
+    )
+  }
+  if (isFacebook(requestContext)) {
+    const text = getText(props.children)
+    const multichannelFacebook = new MultichannelFacebook()
+    const {
+      texts,
+      propsLastText,
+      propsWithoutChildren,
+    } = multichannelFacebook.convertText(props, text[0])
+    return (
+      <>
+        {texts &&
+          texts.map((e, i) => (
+            <Text key={i} {...propsWithoutChildren}>
+              {e}
+            </Text>
+          ))}
+        <Text {...propsLastText}>{propsLastText.children}</Text>
+      </>
     )
   } else {
     return <Text {...props}>{props.children}</Text>
--- a/packages/botonic-react/src/components/multichannel/multichannel.jsx
+++ b/packages/botonic-react/src/components/multichannel/multichannel.jsx
@@ -1,5 +1,6 @@
 import React, { useContext } from 'react'
 
+import { COMPONENT_TYPE } from '../../constants'
 import { RequestContext } from '../../contexts'
 import { Text } from '../text'
 import { deepMapWithIndex } from './deepmap-with-index'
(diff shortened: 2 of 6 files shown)


packages/botonic-react/src/components/multichannel/multichannel-facebook.jsx
```diff
@@ -0,0 +1,39 @@
...
+      const newProps = this.getNewProps(props, finalText)
+      return { texts, newProps }
+    }
+    return { newProps: props }
+  }
+
+  getNewProps(props, finalText) {
+    const newProps = { ...props }
+    Array.isArray(props.children)
+      ? (newProps.children[0] = finalText)
```
vanbasten17, line 18: beware this kind of logic (accessing directly with children[0]). In my opinion, it will be better to filter children by elements that are not React Components and then use the operator `join('')` to reagroup them.
You have an example in webchat's `text.jsx`.
```
const text = children
    .filter(e => !e.type) // <-- This
    .map(e => {
      if (Array.isArray(e)) return serializeText(e)
      else return String(e)
    })
    .join('')
```

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -89,6 +91,16 @@ export const MultichannelText = props => {
         {elements}
       </Text>
     )
+  } else if (isFacebook(requestContext)) {
```
vanbasten17, line 94: unnecessary else if, just do it an if below

packages/botonic-react/src/components/multichannel/multichannel.jsx
```diff
@@ -8,13 +8,29 @@ import { MultichannelCarousel } from './multichannel-carousel'
...
+  MULTICHANNEL_WHATSAPP_PROPS,
+} from './multichannel-utils'
 
 export const Multichannel = props => {
   const requestContext = useContext(RequestContext)
 
-  if (!isWhatsapp(requestContext)) {
+  if (!isWhatsapp(requestContext) && !isFacebook(requestContext)) {
     return props.children
+  } else if (isFacebook(requestContext)) {
```
vanbasten17, line 22: same here?

packages/botonic-react/src/components/multichannel/multichannel.jsx
```diff
@@ -8,13 +8,29 @@ import { MultichannelCarousel } from './multichannel-carousel'
...
+    const newChildren = deepMapWithIndex(props.children, child => {
+      if (child && child.type && child.type.name === 'Text') {
```
vanbasten17, line 24: Can you convert string literals like: `'Text'` to a constant?

packages/botonic-react/tests/components/multichannel/multichannel-facebook.test.jsx
```diff
@@ -0,0 +1,80 @@
...
+    props: {
+      children: 'Children',
+      url: 'https://botonic.io/',
+    },
+    type: 'Button',
+  }
+  const shortText = [
+    'Continetur devenietur id ex denegassem ponderibus imaginatio patrocinio.',
+  ]
+  const longText = [
```
vanbasten17, line 16: what if the text is three times the max length? have you covered it?

packages/botonic-react/tests/components/multichannel/multichannel-facebook.test.jsx
```diff
@@ -0,0 +1,80 @@
...
+  test('TEST: Split long text', () => {
+    const props = {
+      children: [longText[0]],
+    }
+    const { texts, newProps } = multichannelFacebook.convertText(
+      props,
+      longText[0]
+    )
+    const resultProps = {
+      children: lastSplitedText,
```
vanbasten17, line 36: rename every `splited` occurrence to `splitted`

packages/botonic-react/src/components/multichannel/multichannel-facebook.jsx
```diff
@@ -0,0 +1,39 @@
+import { MAX_CHARACTERS_FACEBOOK } from './multichannel-utils'
+
+export class MultichannelFacebook {
```
vanbasten17, line 3: Move this into `multichannel/facebook/facebook.jsx`? Then we would be able to move also whatsapp to their corresponding folder.

vanbasten17 (changes requested):
Looking good overall @AinaVendrell , good job!! 👏 👏 👏  Please take a look at the feedback, reach me out if you have any doubt

packages/botonic-react/tests/components/multichannel/multichannel-facebook.test.jsx
AinaVendrell: Yes it's covered. `longText` has almost 1400 characters, so in this case the text is divided in three messages

packages/botonic-react/src/components/multichannel/multichannel.jsx
```diff
@@ -8,13 +8,29 @@ import { MultichannelCarousel } from './multichannel-carousel'
...
+        return (
+          <MultichannelText key={child.key}>
```
dpinol, line 26: what happens with Text props? See whatsapp example below.
This function is too long.
maybe you can do st like:
```
function getConverters() {
if (isFacebook(requestContext)) { 
return {
'Text': convertText,
}
if (isWhatsapp(requestContext)) { 
return {
'Text': convertText,
'Carousel': convertCarousel
...
}
}
const converters = getConverters()
if converters[child.type.name]) {
return converters[child.type.name](child)
})
else
   return child
```

call me if you have doubts

dpinol (changes requested):

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -89,6 +91,16 @@ export const MultichannelText = props => {
...
+    const text = getText(props.children)[0]
+    const multichannelFacebook = new MultichannelFacebook()
+    const { texts, newProps } = multichannelFacebook.convertText(props, text)
+    return (
+      <>
+        {texts && texts.map((e, i) => <Text key={i}>{e}</Text>)}
```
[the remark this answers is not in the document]
vanbasten17: sure, are missing here, they should be added

packages/botonic-react/src/components/multichannel/multichannel-facebook.jsx
```diff
@@ -0,0 +1,39 @@
...
+    Array.isArray(props.children)
+      ? (newProps.children[0] = finalText)
+      : (newProps.children = finalText)
+    return newProps
+  }
+
+  splitText(originalText) {
+    const splitedText = originalText.split('\n')
+    const initialText = splitedText.shift()
+    const texts = [initialText]
```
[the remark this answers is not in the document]
AinaVendrell: yes but it would be added with an initial \n

AinaVendrell:
requested changes applied

vanbasten17 (approved):

dpinol (approved):

PR #1136, opened 2020-11-25
style(react): fix & add types

dpinol:
Added the following react typescript missing types:

- WebchatContext
- CoverComponent
- staticAsset
- isDev
- isProp
-

### DIFF
--- a/packages/botonic-react/src/components/index.d.ts
+++ b/packages/botonic-react/src/components/index.d.ts
@@ -1,6 +1,8 @@
 import * as React from 'react'
 import { ErrorInfo } from 'react'
 
+import { CoverComponentProps } from '../webchat'
+
 export type MessageType =
   | 'audio'
   | 'buttonmessage'
@@ -25,7 +27,9 @@ export interface MessageProps {
 
 export const Audio: React.FunctionComponent<MessageProps>
 export const Document: React.FunctionComponent<MessageProps>
+
 export const Message: React.FunctionComponent<MessageProps>
+export type Message = React.FunctionComponent<MessageProps>
 export const Video: React.FunctionComponent<MessageProps>
 
 export interface TextProps extends MessageProps {
@@ -50,8 +54,13 @@ export interface ButtonProps {
 
 export const Button: React.FunctionComponent<ButtonProps>
 
+export interface ReplyProps {
+  path?: string
+  payload?: string
+  children: string
+}
 export const Reply: React.FunctionComponent<ReplyProps>
-export type ReplyProps = ButtonProps
+export type Reply = React.FunctionComponent<ReplyProps>
 
 export interface PicProps {
   src: string
@@ -126,7 +135,7 @@ export interface ThemeProps extends StyleProp {
   message?: {
     bot?: BlobProps & ImageProp & StyleProp
     user?: BlobProps & StyleProp
-    customTypes?: React.Component[]
+    customTypes?: React.ComponentType[]
   } & StyleProp & {
       timestamps?: {
         enable?: boolean
@@ -170,10 +179,8 @@ export interface ThemeProps extends StyleProp {
 }
 
 export interface CoverComponentOptions {
-  coverComponent: {
-    component: React.Component
-    props: any
-  }
+  component: React.Component<CoverComponentProps>
+  props?: any
 }
 
 export interface WebchatSettingsProps {
@@ -199,12 +206,9 @@ export function createErrorBoundary<Props>(_?: {
   errorComponent: React.ComponentType
 }): ErrorBoundary<Props>
 
-export function customMessage<
-  Props,
-  CustomMessageComponent extends React.ComponentType<Props>
->(_: {
+export function customMessage<Props>(_: {
   name: string
-  component: CustomMessageComponent
+  component: React.ComponentType<Props>
   defaultProps?: Record<string, unknown>
   errorBoundary?: ErrorBoundary<Props>
 }): WrappedComponent<Props>
--- a/packages/botonic-react/src/index.d.ts
+++ b/packages/botonic-react/src/index.d.ts
@@ -217,11 +217,11 @@ export interface WebchatContextProps {
     property: string,
     defaultValue?: string
   ) => string | undefined
-  resolveCase: () => string
+  resolveCase: () => void
   theme: ThemeProps
   webchatState: WebchatState
   updateWebchatDevSettings: (settings: WebchatSettingsProps) => void
-  updateUser: (user: core.SessionUser) => string
+  updateUser: (user: core.SessionUser) => void
 }
 export const WebchatContext: React.Context<WebchatContextProps>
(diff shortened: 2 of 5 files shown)


packages/botonic-react/src/index.d.ts
```diff
@@ -198,11 +201,37 @@ export class WebchatApp {
   updateWebchatSettings(settings: WebchatSettingsProps): void
 }
 
+export interface WebChatContextProps {
```
asastre, line 204: I think `setReplies` is missing

packages/botonic-react/src/components/index.d.ts
```diff
@@ -25,7 +27,9 @@ export interface MessageProps {
 
 export const Audio: React.FunctionComponent<MessageProps>
 export const Document: React.FunctionComponent<MessageProps>
+
 export const Message: React.FunctionComponent<MessageProps>
+export type Message = React.FunctionComponent<MessageProps>
```
asastre, line 32: What's the difference between this line and the previous one and why exporting a const without assignment?

packages/botonic-react/src/components/index.d.ts
```diff
@@ -50,8 +54,13 @@ export interface ButtonProps {
...
 export const Button: React.FunctionComponent<ButtonProps>
 
+export interface ReplyProps {
+  path?: string
+  payload?: string
+  children: string
+}
 export const Reply: React.FunctionComponent<ReplyProps>
-export type ReplyProps = ButtonProps
+export type Reply = React.FunctionComponent<MessageProps>
```
asastre, line 63: Is this correct or it should be `React.FunctionComponent<ReplyProps>`?

packages/botonic-react/src/webchat/index.d.ts
```diff
@@ -1,7 +1,42 @@
+import * as core from '@botonic/core'
 import * as React from 'react'
 import { RefObject } from 'react'
 
-import { WebchatApp, WebchatArgs } from '../index'
+import { Reply, Webview } from '../components/index'
+import { Message, WebchatApp, WebchatArgs } from '../index'
+
+export interface WebchatState {
```
asastre, line 8: I think `themeUpdates` is missing?

asastre (changes requested):

packages/botonic-react/src/components/index.d.ts
dpinol: this is something typical of typescript.
Message is both:
*  a thing (imagine const Message = (props => ...)) that you can pass as parameter of a function so that the function can create an jsx element from it 
* and a type (image that you have a list of buttons you would declare them as buttons: Button[])

packages/botonic-react/src/components/index.d.ts
dpinol: yes, changed!

packages/botonic-react/src/webchat/index.d.ts
dpinol: yes, I added it and rebase, but I'm not sure. can you check? @asastre @vanbasten17  
```
export interface WebchatStateTheme {
  headerTitle: string
  brandColor: string
  brandImage: string
  triggerButtonImage: undefined
  textPlaceholder: string
  style: {
    fontFamily: string
  }
}

....
  theme: WebchatStateTheme
  themeUpdates: Partial<WebchatStateTheme>

```

packages/botonic-react/src/components/index.d.ts
dpinol: about the assignment, in d.ts you can declare types, but you can never assign values, as this is done in the .js files

packages/botonic-react/src/webchat/index.d.ts
asastre: Yes, I can see it now

packages/botonic-react/src/index.d.ts
asastre: It's still missing, that's because it's not needed?

packages/botonic-react/src/components/index.d.ts
```diff
@@ -170,10 +179,8 @@ export interface ThemeProps extends StyleProp {
 }
 
 export interface CoverComponentOptions {
-  coverComponent: {
-    component: React.Component
-    props: any
-  }
+  component: React.Component<CoverComponentProps>
+  props: any
```
vanbasten17, line 183: leave `props` as optional?

packages/botonic-react/src/index.d.ts
vanbasten17: `WebChat` --> `Webchat`

packages/botonic-react/src/index.d.ts
```diff
@@ -198,11 +201,37 @@ export class WebchatApp {
...
+  sendAttachment: (attachment: File) => string
+  sendPayload: (payload: string) => string
+  sendInput: (input: core.Input) => string
+  openWebview: (webviewComponent: Webview) => string
+  addMessage: (message: WebchatMessage) => string
+  updateMessage: (message: WebchatMessage) => string
+  updateReplies: (replies: boolean) => string
+  updateLatestInput: (input: core.Input) => string
+  closeWebview: () => string
+  toggleWebchat: () => string
```
vanbasten17, line 215: it's strange that many of these  returned values are strings when they should not return anything. 
Maybe because in the context are initialized with `''`?

packages/botonic-react/src/index.d.ts
```diff
@@ -198,11 +201,37 @@ export class WebchatApp {
...
+  getThemeProperty: (
+    property: string,
+    defaultValue?: string
+  ) => string | undefined
+  resolveCase: () => string
+  theme: ThemeProps
+  webchatState: WebchatState
+  updateWebchatDevSettings: (settings: WebchatSettingsProps) => void
+  updateUser: (user: core.SessionUser) => string
```
vanbasten17, line 224: here for example, i would expect `updateUser` to return `void` like the previous one

vanbasten17 (approved):
Amazing job!! 💯 💯 💯  Please, take a look at my feedback

packages/botonic-react/src/index.d.ts
dpinol: > It's still missing, that's because it's not needed?

I forgot to mention, it's deprecated in favour of updateReplies

dpinol: > `WebChat` --> `Webchat`

fixed, thanks

packages/botonic-react/src/components/index.d.ts
dpinol: done

packages/botonic-react/src/index.d.ts
dpinol: right, I'll change it to return void. and will also change the initialState

asastre (approved):

PR #1159, opened 2020-12-03
Chore/improve browser detection, remove react-children-utilities dep

vanbasten17:
## Description
- Refactored isBrowser, isNode from functions to constants (now botonic working with newest webpack versions).
- Removed `react-children-utilities` dependency (we were only using one method) and unnecessary `url` import: https://github.com/hubtype/botonic/pull/1159/commits/daee3ecec819540170bb164974a43530121e1d08

### DIFF
--- a/packages/botonic-core/package.json
+++ b/packages/botonic-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/core",
-  "version": "0.16.0-alpha.5",
+  "version": "0.16.0-alpha.6",
   "license": "MIT",
   "description": "Build Chatbots using React",
   "main": "src/index.js",
--- a/packages/botonic-core/package.json
+++ b/packages/botonic-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/core",
-  "version": "0.16.0",
+  "version": "0.16.0-alpha.5",
   "license": "MIT",
   "description": "Build Chatbots using React",
   "main": "src/index.js",
--- a/packages/botonic-core/package.json
+++ b/packages/botonic-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/core",
-  "version": "0.16.0-alpha.6",
+  "version": "0.16.0-alpha.7",
   "license": "MIT",
   "description": "Build Chatbots using React",
   "main": "src/index.js",
--- a/packages/botonic-core/src/router.js
+++ b/packages/botonic-core/src/router.js
@@ -1,5 +1,3 @@
-import * as url from 'url'
-
 import { RouteInspector } from './debug/inspector'
 import { isBrowser, isFunction, isNode } from './utils'
 
@@ -41,7 +39,7 @@ export class Router {
       if (pathParams) {
         let searchParams = ''
         if (isBrowser) searchParams = new URLSearchParams(pathParams)
-        if (isNode) searchParams = new url.URLSearchParams(pathParams)
+        if (isNode) searchParams = new URLSearchParams(pathParams)
         for (const [key, value] of searchParams) {
           routeParams.params
             ? (routeParams.params[key] = value)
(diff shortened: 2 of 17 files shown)


marcpdw (approved):
Good one 👍

packages/botonic-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/core",
-  "version": "0.16.0",
+  "version": "0.16.0-alpha.7",
```
asastre, line 3: Shouldn't be the version `0.16.1-alpha.7` instead of `0.16.0-alpha.7`? The `0.16.0` is already the latest version published for `@botonic/core`.

asastre (approved):
LGTM :+1:

ericmarcos (changes requested):

packages/botonic-core/src/utils.js
```diff
@@ -1,13 +1,11 @@
-export const isBrowser = () => {
-  return typeof window !== 'undefined' && !window.process
-}
+export const isBrowser =
```
dpinol, line 1: this is an API break. Please flag as "to document" and add doc for Anne.
Also, we shouldn't break compatibility in a patch upgrade.
delay until 0.17 or split into 2 PRs

dpinol (changes requested):

packages/botonic-core/src/utils.js
vanbasten17: it is not breaking current functioning with webpack 4

dpinol: what I mean is that you changed from a function to a constant, so the bots which used it will need to drop the `()`

vanbasten17: okay, now I got it, so should we do it later this change? I can revert the commit

packages/botonic-react/src/components/multichannel/deepmap-with-index.js
```diff
@@ -1,5 +1,6 @@
 import { Children, cloneElement, isValidElement } from 'react'
-import { hasComplexChildren } from 'react-children-utilities'
+
+import { hasComplexChildren } from '../../util/react'
 
 // react-children-utilities 2.1.0 fixes the issue in one deepMapFn call,
 // but not in the other one
```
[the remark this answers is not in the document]
vanbasten17: applied in https://github.com/hubtype/botonic/pull/1173

packages/botonic-core/src/router.js
```diff
@@ -1,5 +1,3 @@
-import * as url from 'url'
```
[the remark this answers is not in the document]
vanbasten17: we do not support Node < 10. I have applied related feedback here: https://github.com/hubtype/botonic/pull/1173

packages/botonic-core/src/utils.js
```diff
@@ -1,13 +1,11 @@
...
+  typeof window !== 'undefined' &&
+  !(window.process && Object.keys(window.process.env).length)
 
-export const isNode = () => {
-  return !isBrowser() && typeof process !== 'undefined'
-}
+export const isNode = !isBrowser && typeof process !== 'undefined'
```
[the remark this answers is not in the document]
vanbasten17: sure, will close this PR and do it in a webpackonian way

packages/botonic-core/src/router.js
```diff
@@ -40,8 +38,8 @@ export class Router {
     try {
       if (pathParams) {
         let searchParams = ''
-        if (isBrowser()) searchParams = new URLSearchParams(pathParams)
-        if (isNode()) searchParams = new url.URLSearchParams(pathParams)
+        if (isBrowser) searchParams = new URLSearchParams(pathParams)
```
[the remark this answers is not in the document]
vanbasten17: applied in https://github.com/hubtype/botonic/pull/1173

packages/botonic-core/package.json
vanbasten17: yes, somehow sb didn't updated it to 0.16.1, will change it in next alphas

vanbasten17:
Closing this PR, will try to use the webpack way to detect environments

PR #1173, opened 2020-12-09
chore(react): remove react-children-utilities dependency, remove url import

vanbasten17:
## Description

* Independent branch from https://github.com/hubtype/botonic/pull/1159 (applied feedback)
* Remove react-children-utilities dependency 
* Removed unnecessary url import and refactored piece of code

### DIFF
--- a/packages/botonic-react/src/util/react.js
+++ b/packages/botonic-react/src/util/react.js
@@ -1,4 +1,5 @@
 import { isBrowser, isNode } from '@botonic/core'
+import { Children, isValidElement } from 'react'
 
 import { mapObject } from './objects'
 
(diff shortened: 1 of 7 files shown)


MarcosCA (approved):

packages/botonic-react/src/util/react.js
```diff
@@ -18,3 +19,14 @@ export const mapObjectNonBooleanValues = obj => {
     return [key, value]
   })
 }
+
+const hasChildren = element =>
+  isValidElement(element) && Boolean(element.props.children)
+
+export const hasComplexChildren = element =>
+  isValidElement(element) &&
+  hasChildren(element) &&
```
asastre, line 28: In `hasChildren` you're calling again the `isValidElement` of `element`, I think the `isValidElement` from line 27 can be removed.

asastre (approved):

packages/botonic-react/src/util/react.js
vanbasten17: good catch! changed 👍

PR #1176, opened 2020-12-10
Chore/cli improvements

vanbasten17:
## Description
* fix(cli): make spawned processes to work on Windows: now commands which were using `spawn` like `serve` and `train` work as expected on Windows
* feat(cli): track o.s information for better analytics: cli to track useful information for us for solving future errors

I will add pre scripts in `botonic-examples` repo

### DIFF
--- a/packages/botonic-cli/src/commands/train.ts
+++ b/packages/botonic-cli/src/commands/train.ts
@@ -31,7 +31,7 @@ export default class Run extends Command {
     } catch (e) {
       console.log(
         colors.red(
-          `You don't have @botonic/plugin-nlu installed.\nPlease, install it by typing the following command:`
+          `You don't have @botonic/plugin-nlu installed.\nPlease, install it with the following command:`
         )
       )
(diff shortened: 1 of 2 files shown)


ericmarcos (approved):
Let's go!

packages/botonic-cli/src/commands/train.ts
```diff
@@ -29,8 +30,9 @@ export default class Run extends Command {
       await import(botonicNLUPath)
     } catch (e) {
       console.log(
-        `You don't have @botonic/plugin-nlu installed.\nPlease, install it by typing the following command:`
-          .red
+        colors.red(
+          `You don't have @botonic/plugin-nlu installed.\nPlease, install it by typing the following command:`
```
dpinol, line 34: =>Please, install it with:

asastre (approved):

packages/botonic-cli/src/commands/train.ts
vanbasten17: fixed, thanks! 👍

PR #1177, opened 2020-12-10
Codecov & Botonic workflows

manuelfidalgo:
## Description

1. **Setup codecov** for botonic:
    - Configure initial codecov setup
    - Upload coverage report to https://codecov.io/gh/hubtype/botonic
2. Trigger **Github workflows only when necessary**. 

## Context
1. Codecov
Configuring codecov will allow us to improve the coverage on every push (or at least keep it as the level of previous push)
The configuration in [codecov.yml](https://docs.codecov.io/docs/codecovyml-reference) includes an initial setup with the following:
    - Minimum coverage of 40% threshold
    - [Flags](https://docs.codecov.io/docs/flags) for the different projects in packages/ that have unit tests running. (Flags help us isolating reports in a monorepo setup)
You can see the reports in https://codecov.io/gh/hubtype/botonic

2. Trigger Github workflows only when necessary 
Prior to this PR, all workflows were triggered, no matter where the change was coming from.
Now, each workflow is triggered only when the change is originated in the path where the workflow operates. For instance, changes in packages/botonic-plugin-dynamodb/** will trigger botonic-plugin-dynamo-tests.yml workflow only. Of course, if we change a workflow file, it will be executed once it is pushed.

1. Codecov
    - Include a codecov config file with the initial setup
    - add codecov/codecov-action@v1 to worflows in order to upload coverage reports
2. Trigger Github workflows only when necessary 
    - Include paths section on the trigger pointing to the related codebase in the packages folder. Also it makes reference to the workflow itself, to make sure it is executed if changes on it are pushed.

You can see last generated reports at https://codecov.io/gh/hubtype/botonic

## Testing
I executed some tests to see that results from the different paths appear in the codecov page.
Regarding the event that triggers the workflows, I also did some tests but, as I have changed all workflows and paths sometimes are hard to validate, something might have fallen through the cracks, please let me know if you see that workflows are not being triggered upon your changes.

### DIFF
--- a/.github/workflows/botonic-cli-tests.yml
+++ b/.github/workflows/botonic-cli-tests.yml
@@ -3,6 +3,8 @@ name: Botonic cli tests
 on:
   push:
     paths:
+      - '*'
+      - 'packages/*'
       - 'packages/botonic-cli/**'
       - '.github/workflows/botonic-cli-tests.yml'
 
--- a/.github/workflows/botonic-cli-tests.yml
+++ b/.github/workflows/botonic-cli-tests.yml
@@ -1,6 +1,10 @@
 name: Botonic cli tests
 
-on: [push]
+on:
+  push:
+    paths:
+      - 'packages/botonic-cli/**'
+      - '.github/workflows/botonic-cli-tests.yml'
 
 jobs:
   botonic-cli-tests:
--- a/.github/workflows/botonic-cli-tests.yml
+++ b/.github/workflows/botonic-cli-tests.yml
@@ -14,13 +14,13 @@ jobs:
       PACKAGE: botonic-cli
     steps:
       - name: Checking out to current branch
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
       - name: Setting up node
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2-beta
         with:
           node-version: '12'
       - name: Setting up cache
-        uses: actions/cache@v1
+        uses: actions/cache@v2
         with:
           path: ~/.npm
           key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
--- a/.github/workflows/botonic-nlu-tests.yml
+++ b/.github/workflows/botonic-nlu-tests.yml
@@ -3,6 +3,8 @@ name: Botonic nlu tests
 on:
   push:
     paths:
+      - '*'
+      - 'packages/*'
       - 'packages/botonic-nlu/**'
       - '.github/workflows/botonic-nlu-tests.yml'
  
--- a/.github/workflows/botonic-nlu-tests.yml
+++ b/.github/workflows/botonic-nlu-tests.yml
@@ -10,6 +10,8 @@ jobs:
   botonic-plugin-nlu-tests:
     name: Botonic nlu tests
     runs-on: ubuntu-latest
+    env:
+      PACKAGE: botonic-nlu
     steps:
     - name: Checking out to current branch (Step 1 of 7)
       uses: actions/checkout@v1
@@ -27,8 +29,8 @@ jobs:
     - name: Install botonic-nlu (Step 4 of 7)
       run: (cd ./packages/botonic-nlu && npm install -D)
     - name: Build botonic-nlu (Step 5 of 7)
-      run: (cd ./packages/botonic-nlu && npm run build)
+      run: (cd ./packages/$PACKAGE && npm run build)
     - name: Install common packages dependencies (Step 6 of 7)
       run: npm install -D
     - name: Verify lint botonic-nlu (Step 7 of 7)
-      run: (cd ./packages/botonic-nlu && npm run lint_ci)
+      run: (cd ./packages/$PACKAGE && npm run lint_ci)
--- a/.github/workflows/botonic-nlu-tests.yml
+++ b/.github/workflows/botonic-nlu-tests.yml
@@ -14,13 +14,13 @@ jobs:
       PACKAGE: botonic-nlu
     steps:
     - name: Checking out to current branch (Step 1 of 7)
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
     - name: Setting up node (Step 2 of 7)
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v2-beta
       with:
         node-version: '12'
     - name: Setting up cache (Step 3 of 7)
-      uses: actions/cache@v1
+      uses: actions/cache@v2
       with:
         path: ~/.npm
         key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
--- a/.github/workflows/botonic-nlu-tests.yml
+++ b/.github/workflows/botonic-nlu-tests.yml
@@ -1,6 +1,10 @@
 name: Botonic nlu tests
  
-on: [push]
+on:
+  push:
+    paths:
+      - 'packages/botonic-nlu/**'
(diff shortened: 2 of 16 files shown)


ericmarcos (approved):
Nice!! 👏👏👏

.github/workflows/botonic-cli-tests.yml
```diff
@@ -1,6 +1,10 @@
 name: Botonic cli tests
 
-on: [push]
+on:
+  push:
+    paths:
+      - 'packages/botonic-cli/**'
+      - '.github/workflows/botonic-cli-tests.yml'
```
dpinol, line 7: are you sure that you need to include the workflow file itself?

.github/workflows/botonic-plugin-contentful-tests.yml
```diff
@@ -3,11 +3,8 @@ name: Botonic plugin-contentful tests
 on:
   push:
     paths:
-      - '*'
-      - '.github/**'
-      - 'packages/*'
       - 'packages/botonic-plugin-contentful/**'
```
dpinol, line 6: please restore '*'' and 'packages/*' There are there build configurations

.github/workflows/botonic-plugin-dashbot-tests.yml
```diff
@@ -21,8 +27,8 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-node-
     - name: Install botonic-plugin-dashbot (Step 4 of 6)
-      run: (cd ./packages/botonic-plugin-dashbot && npm install -D)
+      run: (cd ./packages/$PACKAGE && npm install -D)
```
dpinol, line 30: :clap:

.github/workflows/botonic-plugin-dashbot-tests.yml
```diff
@@ -21,8 +27,8 @@ jobs:
...
     - name: Install common packages dependencies (Step 5 of 6)
       run: npm install -D
     - name: Verify lint botonic-plugin-dashbot (Step 6 of 6)
```
dpinol, line 33: please remove "botonic-plugin-dashbot" from the name ( I already removed it from most fpackages, but please check if there are more), so that the workflow files are most similar

.github/workflows/botonic-react-tests.yml
```diff
@@ -31,4 +35,12 @@ jobs:
     - name: Verify lint
       run: (cd ./packages/$PACKAGE && npm run lint_ci)
     - name: Verify index.d.ts
-      run: scripts/qa/lint-d-ts.sh packages/botonic-core
+      run: scripts/qa/lint-d-ts.sh ./packages/$PACKAGE
+    
+    - name: Upload coverage to codecov
```
dpinol, line 40: can you try this example https://stackoverflow.com/a/59798779/145289  to see if you can have the codecov config in a single file?

dpinol (changes requested):
:clap:

.github/workflows/botonic-cli-tests.yml
manuelfidalgo: From what I understood from the [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths), a git diff is performed so I think it is needed.
I think I did a test on that but I will double check

.github/workflows/botonic-plugin-contentful-tests.yml
manuelfidalgo: ok, sorry, I thought they were not needed. I pushed a new commit with the changes

vanbasten17 (approved):
Nice!! 💯 go for it after taking a look at @dpinol suggestions!

.github/workflows/botonic-react-tests.yml
manuelfidalgo: From what I have read (https://github.com/actions/runner/issues/646), that would the case for reusing a pure bash/docker/JS action, not to call one action from another.
I didn't tried because of that link but I can give it a try.

.github/workflows/botonic-plugin-dashbot-tests.yml
manuelfidalgo: I created https://github.com/hubtype/botonic/pull/1177/commits/ba06acede628229d3aee2fe5e04eff27f8bcba85 for this

.github/workflows/botonic-plugin-contentful-tests.yml
```diff
@@ -4,10 +4,9 @@ on:
   push:
     paths:
       - '*'
```
dpinol, line 6: sorry, can you add '*' and 'packages/*' for all packages?

.github/workflows/botonic-nlu-tests.yml
```diff
@@ -1,11 +1,17 @@
...
  
 jobs:
   botonic-plugin-nlu-tests:
     name: Botonic nlu tests
     runs-on: ubuntu-latest
+    env:
+      PACKAGE: botonic-nlu
     steps:
     - name: Checking out to current branch (Step 1 of 7)
       uses: actions/checkout@v1
```
dpinol, line 17: make the most of this PR to try to upgrade to actions/checkout@v2, actions/cache@v2 and actions/setup-node@v2
luckily you can do a replace-all for this

dpinol (changes requested):

.github/workflows/botonic-cli-tests.yml
manuelfidalgo: I did some checks on https://github.com/hubtype/botonic/tree/chore/codecov-temp branch and I confirm that line is needed: if we remove it, changes on the .yml file won't trigger the workflow.

.github/workflows/botonic-nlu-tests.yml
manuelfidalgo: updated in https://github.com/hubtype/botonic/pull/1177/commits/6076e6a62da605e85b0a648245c0014480f2ac7e

.github/workflows/botonic-plugin-contentful-tests.yml
manuelfidalgo: included in https://github.com/hubtype/botonic/pull/1177/commits/7e317273c5bb55d5d76562a61db9d4b21bf14d58

dpinol (approved):
:clap:

.github/workflows/botonic-cli-tests.yml
```diff
@@ -1,6 +1,12 @@
 name: Botonic cli tests
 
-on: [push]
+on:
+  push:
+    paths:
+      - '*'
```
gbarba, line 6: I guess that a single `*` is not recursive so it doesn't means a change anywhere

.github/workflows/botonic-cli-tests.yml
```diff
@@ -1,6 +1,12 @@
 name: Botonic cli tests
 
-on: [push]
+on:
```
gbarba, line 3: You are not uploading the coverage for this package.
I see it isn't running any test (there isn't any `npm run test`).
I would prefer to also upload this coverage that will be 0%, but this will make visible that we don't have any testing for this package.
But this can go in a separate PR to get this merged

gbarba (approved):
I would like to see another PR that adds the codecov upload for all packages so we make explicit that we have these packages untested (and we can see progress in the future).
This, I guess, will require to add a `npm run test` to them, despite this will run nothing.
But confirm with @dpinol and botonic team

Also I see lot of repetition. maybe we can think in a refactor to put a "generic workflow" that is called with params for each workflow

PR #1183, opened 2020-12-14
Chore/webpack5 changes

vanbasten17:
## Description
* Improve browser/node detection.
* Use env variables to use tree shaking in webpack's compile time.
* Leave it backward compatible with webpack 4.
* Test that same botonic packages in webpack 4 and webpack 5 configs works
* Modify staticAsset to work with bwd code. (Now `staticAsset` won't be necessary to be used, webpack 5 resolves assets as an absolute url)
* Adapt `@botonic/nlu` exported constants to avoid using `path` or `os` dependencies (causing issues in WP5, not polyfilled)
* Modify and refactored publish script to allow publishing betas

## Context
Webpack 5 doesn't add polyfills for node by default as Webpack 4 does. This is done because in terms of front-end development Node variables (process) should not be mixed with front-end code.

In order to do a proper tree shaking in compile time, we will be declaring env variables like this in WP5:
```
    plugins: [
      new HtmlWebpackPlugin({
        template: path.resolve(BOTONIC_PATH, 'src', TEMPLATES.WEBCHAT),
        filename: 'index.html',
      }),
      imageminPlugin,
      new webpack.DefinePlugin({
        IS_BROWSER: true,
        HUBTYPE_API_URL: JSON.stringify(process.env.HUBTYPE_API_URL),
        WEBCHAT_PUSHER_KEY: JSON.stringify(process.env.WEBCHAT_PUSHER_KEY),
      }),
    ],
```
that are directly accessible by using the symbol `HUBTYPE_API_URL`, ... within the code.
To preserve webpack 4 compatiblity, we are checking if `typeof HUBTYPE_API_URL` is defined, otherwise we fall back to old logic supported by webpack 4.
Ref: https://stackoverflow.com/a/64612964
Defining env variables with `process.env.{ENV_VAR}` is not working properly in this new release of webpack


packages/botonic-core/src/hubtype-service.js
```diff
@@ -1,9 +1,18 @@
...
 import Pusher from 'pusher-js'
 
-const PUSHER_KEY = process.env.WEBCHAT_PUSHER_KEY || '434ca667c8e6cb3f641c'
-const HUBTYPE_API_URL = process.env.HUBTYPE_API_URL || 'https://api.hubtype.com'
-
+const PUSHER_KEY =
+  (typeof WEBCHAT_PUSHER_KEY !== 'undefined'
+    ? // eslint-disable-next-line no-undef
+      WEBCHAT_PUSHER_KEY
+    : typeof process !== 'undefined' && process.env.WEBCHAT_PUSHER_KEY) ||
```
dpinol, line 8: this is much too complex. I suggest create a getEnvVar(varName, defaultValue) to implement lines8 & 9

packages/botonic-core/src/hubtype-service.js
```diff
@@ -1,9 +1,18 @@
 import axios from 'axios'
 import Pusher from 'pusher-js'
 
-const PUSHER_KEY = process.env.WEBCHAT_PUSHER_KEY || '434ca667c8e6cb3f641c'
-const HUBTYPE_API_URL = process.env.HUBTYPE_API_URL || 'https://api.hubtype.com'
-
+const PUSHER_KEY =
+  (typeof WEBCHAT_PUSHER_KEY !== 'undefined'
```
dpinol, line 5: where is WEBCHAT_PUSHER_KEY defined?

packages/botonic-core/src/hubtype-service.js
```diff
@@ -1,9 +1,18 @@
...
+    ? // eslint-disable-next-line no-undef
+      WEBCHAT_PUSHER_KEY
+    : typeof process !== 'undefined' && process.env.WEBCHAT_PUSHER_KEY) ||
+  '434ca667c8e6cb3f641c'
+const HUBTYPE_URL =
```
dpinol, line 10: why did you rename it? I think it's less error prone to keep the same name

scripts/ci/publish.ts
```diff
@@ -20,9 +20,14 @@ const packagesDir = join(process.cwd(), C.PACKAGES_DIRNAME)
 const packagesList = sortPackagesByPreference(packagesDir)
 
 ;(async () => {
-  const version = await doAskVersionToPublish()
-  const confirmation = await doAskForConfirmation(version)
-  if (!confirmation) return
+  let version = ''
```
dpinol, line 23: move this to a getVersionAndConfirmation()

scripts/ci/utils.ts
```diff
@@ -77,10 +77,14 @@ export const bumpVersion = async (version, packagePath) => {
     await spawnProcess('npm', ['version', 'minor'], {
       onSuccess: logBumpedVersion,
     })
-  } else {
+  } else if (version === 'alpha' || version === 'rc') {
```
dpinol, line 80: what about beta?
also use ['alpha',...].includes(version)

scripts/ci/utils.ts
```diff
@@ -118,6 +122,7 @@ export const build = async () => {
 
 export const publish = async version => {
   console.log(` - Publishing ${version} version...`)
+
   if (version === 'rc' || version === 'alpha') {
```
dpinol, line 126: beta

dpinol (changes requested):

packages/botonic-core/src/hubtype-service.js
vanbasten17: `WEBCHAT_PUSHER_KEY` comes from:
```
new webpack.DefinePlugin({
        IS_BROWSER: true,
        HUBTYPE_API_URL: JSON.stringify(process.env.HUBTYPE_API_URL),
        WEBCHAT_PUSHER_KEY: JSON.stringify(process.env.WEBCHAT_PUSHER_KEY),
      }),
```
when using DefinePlugin, the variable is available at compile time to do the build

packages/botonic-core/src/hubtype-service.js
vanbasten17: I think so, but i can't have two symbols with the same name (one that references the webpack env variables, and one for declaring the variable

packages/botonic-core/src/hubtype-service.js
vanbasten17: refactored! 👍

scripts/ci/publish.ts
vanbasten17: refactored, also added support for publishing betas

packages/botonic-core/src/hubtype-service.js
vanbasten17: now refactored, using `getWebpackEnvVar`

packages/botonic-core/src/hubtype-service.js
vanbasten17: preserved same symbol, using it inside an anonymous function to handle the scope

dpinol (approved):

asastre (approved):
:clap: :clap: :clap: Nice!
Just 2 small corrections in texts.

packages/botonic-cli/README.md
```diff
@@ -66,10 +64,12 @@ EXAMPLE
   🚀 Bot deployed!
 ```
 
-_See code: [lib/commands/deploy.js](https://github.com/hubtype/botonic/blob/v0.16.0/lib/commands/deploy.js)_
+_See code: [lib/commands/deploy.js](https://github.com/hubtype/botonic/blob/v0.17.0-alpha.4/lib/commands/deploy.js)_
 
 ## `botonic help [COMMAND]`
 
+display help for botonic
```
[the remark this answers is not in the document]
vanbasten17: sure, this readme is autogenerated by oclif, i will take a closer look before publishing

PR #1232, opened 2021-01-05
Configure Jest to scan all files in src dirs when measuring coverage

manuelfidalgo:
## Description
Configure Jest to look for all files in src dir in order to measure coverage properly.
Same config has been propposed to bots-definition in https://github.com/metis-ai/bots-definition/pull/404

## Context
With the previous configuration, Jest would only look for items imported in the tests files, resulting in an unrealistic coverage. e.g. A src file without tests and not referenced by other tests, will not be taken into account for the coverage.

Adding src dir to the roots of the Jest configuration will force Jest to scan ALL src files no matter if they are imported or not.
In all jest config files, we have changed the line `roots: ['tests/'],` to `roots: ['src/', 'tests/'],`

## Testing
Tested in `packages/botonic-react` folder with `npm test`:
After applying the changes, we observe that coverage has slighltly decreased, since more files are taken into account for measuring the whole coverage

### DIFF
--- a/packages/botonic-core/jest.config.js
+++ b/packages/botonic-core/jest.config.js
@@ -1,0 +1,24 @@
+// Options about JS are for compiling @botonic .js/jsx files
+module.exports = {
+  roots: ['src/'],
+  testRegex: '(/tests/.*|(\\.|/)(test|spec))\\.(js|jsx)$',
+  testPathIgnorePatterns: [
+    'lib',
+    '.*.d.ts',
+    'tests/helpers',
+    '.*.helper.js',
+    'tests/__mocks__',
+  ],
+  collectCoverageFrom: ['src/**/*.{js,jsx,ts,tsx}', '!/node_modules/'],
+  transformIgnorePatterns: [
+    'node_modules/(?!@botonic|react-children-utilities).+\\.(js|jsx)$',
+  ],
+  moduleFileExtensions: ['js', 'jsx', 'json'],
+  snapshotSerializers: [],
+  modulePaths: ['node_modules', 'src'],
+  moduleNameMapper: {
+    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
+      '<rootDir>/tests/__mocks__/file-mock.js',
+    '\\.(css|less|scss|sass)$': 'identity-obj-proxy',
+  },
+}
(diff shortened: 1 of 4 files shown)


ericmarcos (approved):

dpinol (commented):
:clap
Please get coverage % before and after the change and add it to a need 2021/01/08 tab at https://docs.google.com/spreadsheets/d/1_33VpvHmgW5fXwoWzSgfCKHYsBIza1d3Es5F7BztXm4/edit#gid=1543842881
so that we can calculate how much we have improved during the last quarter

packages/botonic-plugin-contentful/jest.config.js
```diff
@@ -1,6 +1,6 @@
 // Options about JS are for compiling @botonic .js/jsx files
```
dpinol, line 1: I see there's no jest.config.js for core.
Can you add one with minimal contents? Not sure if by default it was already adding src to the roots

dpinol (changes requested):

packages/botonic-plugin-contentful/jest.config.js
manuelfidalgo: good catch @dpinol ! There were some files missing from src.
I added a jest config file for taking care of that.

dpinol (approved):

packages/botonic-core/jest.config.js
```diff
@@ -0,0 +1,24 @@
...
+  roots: ['src/'],
+  testRegex: '(/tests/.*|(\\.|/)(test|spec))\\.(js|jsx)$',
+  testPathIgnorePatterns: [
+    'lib',
+    '.*.d.ts',
+    'tests/helpers',
+    '.*.helper.js',
+    'tests/__mocks__',
+  ],
+  collectCoverageFrom: ['src/**/*.{js,jsx,ts,tsx}', '!/node_modules/'],
```
vanbasten17, line 12: I think that we don't have any file with this extension, but if we want to cover all Javascript scenarios, I would add also `mjs` extension. https://medium.com/passpill-project/files-with-mjs-extension-for-javascript-modules-ced195d7c84a

vanbasten17 (approved):
LGTM! Go for it! 💪

packages/botonic-core/jest.config.js
manuelfidalgo: thanks for the observation, I think we can add it when we have that kind of files

PR #1245, opened 2021-01-12
feat: Disabling buttons components once clicked

vanbasten17:
## Description

* Disabling buttons once clicked.
* As discussed, the properties are `autodisable` and `disabledstyle` (they are in lowercase to avoid annoying warnings, as these props will be rendered in the server side also).
* Created a class `ButtonsDisabler` with static methods inside `/util/webchat.js` to handle all the related logic of this new feature
* Added unit tests

## Context
We need a way to disable buttons in webchat once they are clicked for some customers

In this case, each button receives a `parentId` which is used to know what is the parent needed to have their buttons disabled. Once clicked, the message is updated with its buttons set with the property `disabled: {true|false}`. On next renders, the button will be rendered depending on this property. 

```
export const webchat = {
  theme: {
    style: {
      width: 500,
    },
    button: {
      disabledstyle: {
        opacity: 0.5,
        cursor: 'auto',
        pointerEvents: 'none',
        backgroundColor: 'green',
      },
      autodisable: true,
    },
  },
}
```
### Buttons
```
<Text>
    Here I display two types of buttons, the first one is a URL button and the second is a payload button:
     <Button payload='https://botonic.io' autodisable={false}>Visit botonic.io</Button>
     <Button payload='carousel'>Show me a carousel</Button>
</Text>
```
Screenshot 2021-01-28 at 12 30 29
Screenshot 2021-01-28 at 12 30 34

### Carousel
```
<Carousel>
          {movies.map((e, i) => (
            <Element key={e.name}>
              <Pic src={e.pic} />
              <Title>{e.name}</Title>
              <Subtitle>{e.desc}</Subtitle>
              {i === 0 ? (
                <Button payload={e.url}>Visit website</Button>
              ) : (
                <Button
                  payload={e.url}
                  autodisable={true}
                  disabledstyle={{ backgroundColor: 'red' }}
                >
                  KAKA
                </Button>
              )}
            </Element>
          ))}
        </Carousel>
```
Screenshot 2021-01-28 at 12 30 44
Screenshot 2021-01-28 at 12 30 52

As you can observe, properties defined at component level, overwrite the ones defined in the webchat theme. If `autodisable` is set in `webchat.theme`, then the components that do not have disabling properties will have the defined ones in the theme (appended in their props).

## Testing
* Has unit tests


packages/botonic-react/src/components/button.jsx
```diff
@@ -100,20 +125,22 @@ export const Button = props => {
...
-        style={{
-          ...buttonStyle,
-          color: buttonTextColor,
-          backgroundColor: buttonBgColor,
-        }}
-        bottom={props.bottomRadius}
-      >
-        {props.children}
-      </StyledButton>
+      enabled && (
```
asastre, line 128: We should check here the `diableButtons` value too, right? If we don't do it here, even if the feature is not enabled, we could hide buttons anyway.

packages/botonic-react/src/msg-to-botonic.jsx
```diff
@@ -169,13 +169,15 @@ function buttonsParse(buttons) {
...
     const webview = props.messenger_extensions ? props.url : props.webview
+    const enabled = props.enabled !== undefined ? props.enabled : true
     return (
       <Button
         key={i}
         payload={payload}
         url={url}
         target={target}
         webview={webview}
+        enabled={enabled}
```
asastre, line 180: Should be `parentId` property added here?

asastre (changes requested):
Looking good! Some things to take into account.

packages/botonic-react/src/msg-to-botonic.jsx
vanbasten17: yes, sure, added 👍

packages/botonic-react/src/components/button.jsx
```diff
@@ -65,6 +73,23 @@ export const Button = props => {
...
+      const parent = document.getElementById(parentId)
+      const parentButtons = Array.from(parent.getElementsByTagName('button'))
+      parentButtons.forEach(b => (b.style.display = 'none'))
+      const messageToUpdate = webchatState.messagesJSON.filter(
+        m => m.id == parentId
+      )[0]
+      const updatedButtons = {
+        buttons: messageToUpdate.buttons.map(b => ({
+          ...b,
+          ...{ enabled: false },
```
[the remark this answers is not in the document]
vanbasten17: do you mean just one of the two buttons, does this make sense?

packages/botonic-react/src/components/button.jsx
```diff
@@ -65,6 +73,23 @@ export const Button = props => {
       window.open(props.url, props.target || '_blank')
     }
     if (props.onClick) props.onClick()
+    if (disableButtons) {
+      // TODO: Extract this logic
+      const parent = document.getElementById(parentId)
```
ericmarcos, line 78: This will not work if `shadowDOM` is active. We should pass the `host` prop from Webchat down to the rest of the tree so that it can be used in Button and other components. I think passing it through the `WebchatContext.Provider` would be a nice solution.

packages/botonic-react/src/components/button.jsx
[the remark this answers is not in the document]
ericmarcos: @asastre @vanbasten17 let's think of generic ways to modify the behaviour, like passing `disabledStyle` props or stuff like that, so you could add `display: none`, or `opacity: 0.5` or do nothing (equal to "alwaysEnabled"). That would need additional complexity if we want that styles to be dynamic based on certain runtime info, so needs careful thinking... I would move forward with this implementation for now but juts as a 1st step towards a more generic solution.

packages/botonic-react/src/components/message.jsx
```diff
@@ -94,10 +94,26 @@ export const Message = props => {
...
-    e => e.type === Button
-  )
+  const withParentId = (element, parentId) => {
+    return {
+      ...element,
+      ...{
+        props: {
+          ...element.props,
+          ...{ parentId: parentId },
+        },
```
ericmarcos, line 104: As @asastre said before, you don't need to `...{a: b}`, you can just set these key:values directly

packages/botonic-react/src/components/button.jsx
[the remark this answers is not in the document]
ericmarcos: @marcrabat we shouldn't modify the DOM of other components directly, I think the right approach would be to modify some application state (probably in messagesJSON) of the elements you want to modify, then have these elements have a useEffect to observe these state and render accordingly.

vanbasten17:
Nice! I agree with both of you so next points will be:
* generic property to enable/disable feature
* specific prop: `disabledStyle` to enable/disable button specifically (it should take preference over the generic property in case it is defined)
* Not using DOM operations but directly props that updates the state component with `useEffect`.
* Using providers to expose `host` in case that shadowDOM is enabled.
* Implementation working with Custom Messages.

packages/botonic-react/src/util/webchat.js
```diff
@@ -0,0 +1,151 @@
...
+      if (nestedProperty !== undefined) return nestedProperty
+      const plainProperty = getProperty(theme, k)
+      if (plainProperty !== undefined) return plainProperty
+      return defaultValue
+    }
+  }
+  return undefined
+}
+
+export class ButtonsDisabler {
```
ericmarcos, line 29: This is not a "util". A util is a veeeery generic function that can be used in different contexts. This function is really specific to our Buttons, it deserves it's own file next to Button component definition.

packages/botonic-react/tests/webchat/buttons-disabler.test.jsx
```diff
@@ -0,0 +1,252 @@
...
+        <Element key={e.name}>
+          <Pic src={e.pic} />
+          <Title>{e.name}</Title>
+          <Subtitle>{e.desc}</Subtitle>
+          <Button {...e.props} />
+        </Element>
+      ))}
+    </Carousel>
+  ),
+}
```
ericmarcos, line 252: Nice work with tests!!

packages/botonic-react/src/util/webchat.js
```diff
@@ -0,0 +1,151 @@
...
+            props: {
+              ...n.props,
+              ...{ children: deepMapWithIndex(n.props.children, n => n) },
+            },
+          },
+        }
+      }
+      return n
+    })
+  }
```
ericmarcos, line 90: I think this implementation is way more complex than it should be. You should never mess directly with React internals, unless there's a very good reason, which I fail to see here.
Why can't we just have these props in our Webchat reducer `messagesJSON` which is just a plain and simple JSON representation of our components? Then have the Button components react to changes on that object?

packages/botonic-react/src/util/webchat.js
```diff
@@ -0,0 +1,151 @@
...
+      disabledProps.disabledstyle = JSON.stringify(props.disabledstyle)
+    return disabledProps
+  }
+
+  static withDisabledProps(props) {
+    return {
+      ...{ disabled: props.disabled },
+      ...{ autodisable: props.autodisable },
+      ...{ disabledstyle: props.disabledstyle },
+    }
```
ericmarcos, line 52: I see this all over, you don't need to create a new object to just deconstruct it. Simpler:

```
return {
  disabled: props.disabled,
  ...
}
```

ericmarcos (changes requested):

packages/botonic-react/src/util/webchat.js
vanbasten17: true, changed 👍

packages/botonic-react/src/util/webchat.js
```diff
@@ -0,0 +1,151 @@
...
+        setDisabled: additionalProps.setDisabled,
+        parentId: additionalProps.parentId,
+      }
+    }
+    return {
+      ...node,
+      props: {
+        ...node.props,
+        autodisable: node.props.autodisable,
+        disabledstyle: node.props.disabledstyle,
```
asastre: Are these 2 lines needed? Since both properties (`autodisable` and `disabledstyle`) are present in `node.props`, it should be enought with the previous line (`...node.props`), right?

packages/botonic-react/src/util/webchat.js
vanbasten17: moved to its own file inside components 👍

asastre (changes requested):
Just a few small comments/suggestions

packages/botonic-react/src/node-app.jsx
```diff
@@ -5,7 +5,10 @@ import { ReactBot } from './react-bot'
 export class NodeApp {
   constructor(options) {
     this.bot = new ReactBot({
-      renderer: args => this.renderNode(args),
+      renderer: args => {
+        args.request.theme = options.theme
```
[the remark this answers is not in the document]
vanbasten17: this was residual from trying things while developing it, leaving as it was, good catch

packages/botonic-react/src/components/button.jsx
```diff
@@ -109,6 +125,11 @@ export const Button = props => {
           ...buttonStyle,
           color: buttonTextColor,
           backgroundColor: buttonBgColor,
+          ...(props.disabled &&
+            autoDisable && {
+              ...WEBCHAT.DEFAULTS.BUTTON_DISABLED_STYLE,
+              ...disabledStyle,
```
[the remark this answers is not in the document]
vanbasten17: sure, not needed anymore!

packages/botonic-react/src/util/webchat.js
vanbasten17: sure, applied

asastre (approved):
:rocket:

packages/botonic-react/src/util/webchat.js
vanbasten17: As discussed we agreed to move forward, but it's true that Webchat code must be rethinked in order to decouple all the functionalities in a good way.

ericmarcos (approved):

PR #1263, opened 2021-01-21
fix(react): scroll to bottom on focus on non-iphone devices

vanbasten17:
## Description
* Scroll to bottom when focusing an input on non-iphone devices.


elozano98 (approved):
:+1:

asastre (approved):

PR #1289, opened 2021-02-01
fix(contentful): Reimplementation of the stemmer-uk without lookbehind regex.

elozano98:
## Description
The stemmer used for Ukrainian had lookbehind regex that was not supported by Safari, so it has been implemented without including that type of unsupported regex.

## Context

Is trying to fix this bug reported by @vanbasten17 in [this issue](https://github.com/axa-group/nlp.js/issues/741).

The stemmer has been reimplemented in Botonic contentful but replacing the lookbehind regex.

## Testing

The pull request...

- [x] has unit tests
- [x] has integration tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/nlp/stemmers/stemmer-uk.ts
+++ b/packages/botonic-plugin-contentful/src/nlp/stemmers/stemmer-uk.ts
@@ -1,0 +1,88 @@
+import Stemmer from '@nlpjs/core/src/stemmer'
+
+export class StemmerUk implements Stemmer {
+  public stem(tokens: string[]): string[] {
+    return tokens.map(token => this.stemToken(token))
+  }
+
+  private stemToken(token: string): string {
+    const vowelMatch = token.match(/[аеиоуюяіїє]/)
+    if (!vowelMatch) {
+      return token
+    } else if (vowelMatch.index != undefined) {
+      const start = token.slice(0, vowelMatch.index + 1)
+      token = token.slice(vowelMatch.index + 1)
+
+      if (token === '') {
+        return start
+      }
+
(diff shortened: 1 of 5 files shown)


vanbasten17 (approved):

asastre (approved):

packages/botonic-plugin-contentful/src/nlp/stemmers/stemmer-uk.ts
```diff
@@ -0,0 +1,88 @@
+import Stemmer from '@nlpjs/core/src/stemmer'
```
dpinol, line 1: in the future, when importing an external file to slightly modify it, please create a commit with the exact external version , and one commit with the change.
In this case, it's enough if you add a comment in the stemmer specifying what was the original line

packages/botonic-plugin-contentful/tests/nlp/stemmers/stemmer-uk.test.ts
```diff
@@ -0,0 +1,15 @@
+import { tokenizerPerLocale } from '../../../src/nlp'
+import { StemmerUk } from '../../../src/nlp/stemmers/stemmer-uk'
+
+test.each<any>([
+  ['розмовляючи', ['розмовляюч']],
+  ['говорити', ['говор']],
```
dpinol, line 6: can you comment which word is using the code which you have fixed?

dpinol (changes requested):
amazing!
can you create a PR for node-nlp, also? or a ticket for it if you have something urgent

vanbasten17:
we already have warned there @dpinol : https://github.com/axa-group/nlp.js/issues/741 👍

PR #1290, opened 2021-02-01
Chore/improve telemetry

vanbasten17:
## Description

* Track which template has been selected by the developer.
* Track errors when running botonic new.
* Remove build step after running `botonic new`. (It will improve DX and it does not make sense at all)

### DIFF
--- a/packages/botonic-cli/src/commands/new.ts
+++ b/packages/botonic-cli/src/commands/new.ts
@@ -12,7 +12,7 @@ import { promisify } from 'util'
 
 import { BotonicAPIService } from '../botonic-api-service'
 import { EXAMPLES } from '../botonic-examples'
-import { track } from '../utils'
+import { track, trackError } from '../utils'
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const exec = promisify(childProcessExec)
@@ -41,7 +41,6 @@ Creating...
   private botonicApiService = new BotonicAPIService()
 
   async run(): Promise<void> {
-    track('Created Botonic Bot CLI')
     const { args } = this.parse(Run)
     const userProjectName = args.name
     let selectedProjectName = ''
@@ -66,7 +65,9 @@ Creating...
         return
       }
     }
-
+    track('Created Botonic Bot CLI', {
+      selected_project_name: selectedProjectName,
+    })
     if (existsSync(userProjectName)) {
       rimraf.sync(userProjectName)
     }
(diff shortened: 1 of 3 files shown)


packages/botonic-cli/src/commands/new.ts
```diff
@@ -86,7 +87,11 @@ Creating...
         dir: userProjectName,
       })
     } catch (e) {
-      throw new Error(`Error downloading ${selectedProjectName}: ${String(e)}`)
+      const error = `Downloading Project: ${selectedProjectName}: ${String(e)}`
+      trackError('Created Botonic Bot CLI', {
```
dpinol, line 91: Created or Creating?

packages/botonic-cli/src/commands/new.ts
```diff
@@ -95,9 +100,16 @@ Creating...
       spinner: 'bouncingBar',
     }).start()
     const dependencyCommand = `npm install`
-    await exec(dependencyCommand)
+    try {
+      await exec(dependencyCommand)
+    } catch (e) {
+      const error = `Installing dependencies: ${String(e)}`
+      trackError('Created Botonic Bot CLI', {
```
dpinol, line 107: since we always want an error arg for trackError, I prefere that it has 3 arguments: eventName, error, and optional properties

packages/botonic-cli/src/commands/new.ts
dpinol: Or we can just name it "Failed creation ..." or "Not created..." and drop the initial "Error:" injected in trackError

dpinol (changes requested):

packages/botonic-cli/src/utils.ts
```diff
@@ -84,8 +84,11 @@ function getSystemInformation() {
     botonic_dependencies: getBotonicDependencies(),
   }
 }
-
+export function trackError(event: string, properties = {}): void {
+  track(`Error: ${event}`, properties)
+}
 export function track(event: string, properties = {}): void {
+  console.log({ event }, { properties })
```
asastre, line 91: Remove this console.log if not needed.

asastre (approved):
I agree with @dpinol comments, I think it will be better to change a little bit the naming of the errors.

ericmarcos (changes requested):

packages/botonic-cli/src/utils.ts
```diff
@@ -84,8 +84,11 @@ function getSystemInformation() {
     botonic_dependencies: getBotonicDependencies(),
   }
 }
-
+export function trackError(event: string, properties = {}): void {
+  track(`Error: ${event}`, properties)
```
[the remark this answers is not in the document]
vanbasten17: sure, makes sense, applied 👍

MarcosCA (approved):

Issue #1199, opened 2020-12-18
Self-hosted web app built but incorrectly


sampittko:
**Describe the bug**
When proceeding according to [docs](https://botonic.io/docs/deployment/standalone-js-bundle) in order to build standalone JS bundle for chatbot, the only output file is `/dist/main.js`.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create blank Botonic project
2. Proceed according to docs for standalone JS bundle deployment type
3. Build the project
4. See code generated by webpack that is unusable

**What is the expected behaviour?**
Generate `dist/webchat.botonic.js` which can be imported to any website.

**Screenshots**
Build log
![Screen Shot 2020-12-18 at 19 57 24](https://user-images.githubusercontent.com/38221262/102650453-40d5ed80-416b-11eb-942c-531cf92e782b.png)

Folder structure after reproducing steps from docs:
![Screen Shot 2020-12-18 at 19 54 06](https://user-images.githubusercontent.com/38221262/102650200-cc02b380-416a-11eb-9e54-b6a6fc6a4238.png)

**About the environment**
Botonic CLI v0.16.0
Node v15.4.0

**Additional context**
No additional dependencies, just what is included for blank Botonic project

ChristopherGS:
@dpinol I can confirm that we've also just encountered this same bug (tested on node v12 and v14, and on Windows 10 and MacOS). We end up with just the `main.js` file in the `dist` directory. I'll begin investigating now, would really appreciate any guidance.

ChristopherGS:
*update*
Progress so far: I've now fixed the webpack config so that the correct file is output. I updated webpack.config.js like so (note quite a few differences from the docs):

```
...
function botonicSelfHostedConfig(mode) {
  return {
    optimization: optimizationConfig,
    mode: mode,
    devtool: sourceMap(mode),
    target: 'web',
    entry: path.resolve('webpack-entries', 'self-hosted-entry.js'),
    module: {
      rules: [babelLoaderConfig, fileLoaderConfig(path.join('..', ASSETS_DIRNAME)), stylesLoaderConfig],
    },
    output: {
      path: path.resolve(__dirname, 'dist'),
      filename: 'webchat.botonic.js',
      library: 'Botonic',
      libraryTarget: 'umd',
      libraryExport: 'app',
      publicPath: './',
    },
    resolve: resolveConfig,
    plugins: [
      imageminPlugin,
      new webpack.DefinePlugin({
[...]

PR #1293, opened 2021-02-04
docs: update standalone deploy readme

ChristopherGS:
## Description

Updated Botonic Standalone JS deploy readme to address recent breaking changes from version 0.16.0 -> 0.17.0.

## Context

Following the minor version increase to 0.17.0 (and likely due to the updating of the webpack version), the instructions to deploy Botonic as a standalone JS bundle require updating. 

The full bug is described here:
Fixes [1199](https://github.com/hubtype/botonic/issues/1199)

In short, without these adjustments, the output file is incorrect.

Docs updated. Tested locally with both own project and "telco example" from [Botonic examples](https://github.com/hubtype/botonic-examples)

Main remaining question is whether the approach I detail passing in `webchat` to the `render` method of the `SelfHostedApp` class is advisable. It does work, but unsure if it is best practice/maintainable.

## Testing

The pull request...

- doesn't need tests because it's documentation.

### DIFF
--- a/docs/docs/deployment/standalone-js-bundle.md
+++ b/docs/docs/deployment/standalone-js-bundle.md
@@ -32,6 +32,7 @@ import { render } from 'react-dom'
 import { ReactBot } from '@botonic/react/src/react-bot'
 import { WebchatApp } from '@botonic/react/src/webchat-app'
 import { Webchat } from '@botonic/react/src/webchat'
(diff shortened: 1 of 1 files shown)


vanbasten17 (approved):

vanbasten17:
Thanks again for your contribution @ChristopherGS , merging it. Keep rocking it :)

PR #1305, opened 2021-02-09
Chore/improve telemetry

vanbasten17:
## Description
**Fixes to be included in a 0.18.1 release.**
* Track botonic new at the very beginning of its execution.
* Update `@oclif` dependencies in `package.json`
* Track if is interactive console, node version, path to botonic binary, $PATH and current working directory.
```
  is_tty: true,
  node_version: 'v12.20.1',
  binary_path: '/usr/local/bin/botonic',
```

Once approved: including these changes also in https://github.com/hubtype/botonic-examples

### DIFF
--- a/packages/botonic-cli/src/utils.ts
+++ b/packages/botonic-cli/src/utils.ts
@@ -54,9 +54,13 @@ if (isAnalyticsEnabled()) {
   analytics = new Analytics('YD0jpJHNGW12uhLNbgB4wbdTRQ4Cy1Zu')
 }
 
-function getBotonicCliVersion(): string {
+function isWindows() {
+  return process.platform === 'win32'
+}
+
+function execCommand(command: string) {
   try {
-    return String(execSync('botonic --version')).trim()
+    return String(execSync(command)).trim()
   } catch (e) {
     return String(e)
   }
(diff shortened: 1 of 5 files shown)


packages/botonic-cli/src/utils.ts
```diff
@@ -80,7 +84,16 @@ function getSystemInformation() {
     arch: os.arch(),
     timezone: Intl.DateTimeFormat().resolvedOptions().timeZone,
     timestamp: new Date().toISOString(),
-    botonic_cli_version: getBotonicCliVersion(),
+    is_tty: process.stdout.isTTY,
+    binary_path: isWindows()
+      ? execCommand('where botonic')
+      : execCommand('which botonic'),
```
ericmarcos, line 90: The purpose of `binary_path` is to learn where the framework is installed in case there's the install problem and `botonic` command is not found. So with this approach it will fail.
I propose to just get the current directory with the global variable `__dirname` (https://stackoverflow.com/questions/3133243/how-do-i-get-the-path-to-the-current-script-with-node-js)

ericmarcos (changes requested):

packages/botonic-cli/src/utils.ts
vanbasten17: done @ericmarcos ! 👍

packages/botonic-cli/src/utils.ts
```diff
@@ -54,9 +54,13 @@ if (isAnalyticsEnabled()) {
   analytics = new Analytics('YD0jpJHNGW12uhLNbgB4wbdTRQ4Cy1Zu')
 }
 
-function getBotonicCliVersion(): string {
+function isWindows() {
+  return process.platform === 'win32'
```
asastre, line 58: Small observation, I know it's the same, but I would use here `os.platform()` instead of `process.platform` just to keep using the same methods everywhere (we're using `os.platform()` later in `getSystemInformation()`)

asastre (approved):
LGTM

packages/botonic-cli/src/utils.ts
vanbasten17: nice observation, applied! 👍

ericmarcos (approved):

dpinol (approved):

PR #1306, opened 2021-02-10
fix(core/react): include lib as files to be published

vanbasten17:
## Description
* Lib was not published to npm because of missing in "files" property of `package.json` in core/react.

### DIFF
--- a/packages/botonic-cli/package.json
+++ b/packages/botonic-cli/package.json
@@ -40,8 +40,7 @@
     "/lib",
     "/npm-shrinkwrap.json",
     "/oclif.manifest.json",
-    "/scripts",
-    "/templates"
+    "/scripts"
   ],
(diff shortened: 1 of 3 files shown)


packages/botonic-cli/package.json
```diff
@@ -40,8 +40,7 @@
     "/lib",
     "/npm-shrinkwrap.json",
     "/oclif.manifest.json",
-    "/scripts",
-    "/templates"
```
asastre, line 44: Why removing `/templates`?

vanbasten17: these are not included anymore, as we have moved them to `botonic-examples`

asastre (approved):

ericmarcos (approved):

PR #1309, opened 2021-02-12
Botonic plugin contentful: add handoffMessage field

AlbertGom:
## Description
Added a new field named handoffMessage in cms queue. This field is used to show a message before a handoff and it is placed in desk queue contentful content model.

## Testing

The pull request...

- has unit tests


dpinol (approved):
:clap:

vanbasten17 (approved):

PR #1320, opened 2021-02-16
Fix return types for input method in NodeApp and CoreBot

asastre:
## Description
Fix return type for `input` method in `NodeApp` and `CoreBot`, these methods will return Promises.


vanbasten17 (approved):

dpinol (approved):
go @asastre ! :clap:

PR #1332, opened 2021-02-25
Chore/codecov plugin nlu #PLA-309

manuelfidalgo:
## Description
Configure botonic-plugin-nlu package to be taken into account when measuring coverage

## Context
Currently, we cannot compare the coverage of the whole botonic project with previous versions since codecov is not configured for all the packages.
This change will allow us to track the coverage of the nlu plugin with codecov.

- Configure codecov flag
- Configure workflow to upload coverage results
- Configure jest in botonic-plugin-nlu 
- Include a test to trigger coverage

Coverage can be seen in codecov: https://codecov.io/gh/hubtype/botonic/tree/4021c7af5c6b2732bbc11128bf6255b0fd881da3/packages/botonic-plugin-nlu/src

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/.github/workflows/botonic-nlu-tests.yml
+++ b/.github/workflows/botonic-nlu-tests.yml
@@ -29,7 +29,7 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-node-
     - name: Install dev dependencies 
-      run: (cd ./packages/botonic-nlu && npm install -D)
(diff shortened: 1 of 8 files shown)


dpinol (approved):
:clap:

vanbasten17 (approved):

PR #1333, opened 2021-02-28
chore(core): add typings for @botonic/core

vanbasten17:
## Description
* Adding types for @botonic/core in order to start migrating eventually pieces of code to TS.
* Move types into `src/types` preserving project structure. 
* Reference types location in `"types"` package.json's field.

**Tip**: Used the command `npx typescript src/**/*.js --declaration --allowJs --emitDeclarationOnly --outDir types;` to generate the initial structure for types.


packages/botonic-core/src/types/plugins.d.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Input, PluginConfig, Session } from '.'
+
+export declare function loadPlugins(plugins: PluginConfig<any>[]): any
+export declare function runPlugins(
+  plugins: PluginConfig<any>[],
+  mode: 'pre' | 'post',
+  input: Input,
+  session: Session,
+  lastRoutePath: string,
+  response?: any
```
dpinol, line 11: is this a BotResponse?

packages/botonic-core/src/types/plugins.d.ts
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable node/no-missing-import */
+import { Input, PluginConfig, Session } from '.'
+
+export declare function loadPlugins(plugins: PluginConfig<any>[]): any
```
dpinol, line 4: this returns a Plugin[]?

dpinol (changes requested):
amazing! :clap:

packages/botonic-core/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.18.1",
   "license": "MIT",
   "description": "Build Chatbots using React",
-  "main": "src/index.js",
+  "main": "lib/index.js",
```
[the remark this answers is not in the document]
vanbasten17: this was something we also talked with @asastre , removing atm and let's see how we coordinate this.

packages/botonic-core/src/types/plugins.d.ts
vanbasten17: sure

packages/botonic-core/src/types/plugins.d.ts
vanbasten17: yes

packages/botonic-core/package.json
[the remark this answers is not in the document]
asastre: Yes, before doing this change we should evaluate all possible implications.
It would be nice too to test if using lib instead of src improve the execution times of the tests, maybe the fact that the code need to be transpiled before the tests is what makes the tests take so long the first time to run.

packages/botonic-core/src/types/i18n.d.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable node/no-missing-import */
+/* eslint-disable filenames/match-regex */
```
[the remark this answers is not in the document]
vanbasten17: `i18n.d.ts` is detected as an invalid filename.

packages/botonic-core/src/types/index.d.ts
```diff
@@ -1,3 +1,14 @@
+/* eslint-disable node/no-missing-import */
```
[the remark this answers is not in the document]
vanbasten17: added `.d.ts` in tryExtensions key of no-missing-import rule and removed ignores

packages/botonic-core/src/types/index.d.ts
```diff
@@ -1,3 +1,13 @@
+export { Providers } from './constants'
+export { CoreBot } from './core-bot'
+export * from './debug'
+import { Inspector } from './debug'
+export * from './handoff'
+export { HubtypeService } from './hubtype-service'
+export { getString } from './i18n'
+export { Router } from './router'
+export * from './utils'
+
```
marcpdw, line 10: Is it sorting the imports by file instead of object? Do we want it like this?

IMO it's easier to read them by object.

packages/botonic-core/src/index.js
```diff
@@ -1,21 +1,11 @@
...
-  HandOffBuilder,
-  humanHandOff,
-  storeCaseRating,
-} from './handoff'
+export * from './handoff'
 export { HubtypeService } from './hubtype-service'
 export { getString } from './i18n'
 export { Router } from './router'
-export { isBrowser, isMobile, isNode, params2queryString } from './utils'
+export * from './utils'
```
marcpdw, line 8: Why are those `*` imports necessary?

marcpdw (approved):
This is huge, great job here!!! ⭐ 

Not for this PR, but for the next one: remember that `d.ts` files keep the comments you may add to them. So, documentation of _fields_, _parameters_, _objects_, ... can bring big help to our fans. Take `RxJS` for example and check how well-documented they have it!
Of course our stuff is not as complex, but if we want a 5 ⭐  repo, it's something we have to consider

packages/botonic-core/src/types/index.d.ts
vanbasten17: yes, it is sorting them by file right now. Maybe we should discuss it and when we have the Pending PRs cleaner, apply the result of the discussion to all the packages

packages/botonic-core/src/index.js
vanbasten17: It's a shorthand to export all the members exported in `./utils`, so instead to do `import {isBrowser, ...} from './utils'` and then export them, you can export all the functions directly with `*`

vanbasten17: they are exposed as `@botonic/react` uses them, also they can be useful for the developer.

packages/botonic-core/src/types/index.d.ts
vanbasten17: Created a task to talk about this: https://linear.app/hubtype/issue/BOT-319/discussion-sort-imports-by-file-vs-sort-imports-by-object

packages/botonic-core/src/types/constants.d.ts
```diff
@@ -0,0 +1,14 @@
+export const Providers: Readonly<{
+  Messaging: {
```
elozano98, line 2: Should this field name be lowercase?

packages/botonic-core/src/types/constants.d.ts
```diff
@@ -0,0 +1,14 @@
+export const Providers: Readonly<{
```
elozano98, line 1: Why don't use an `enum` for this?

packages/botonic-core/src/types/index.d.ts
```diff
@@ -162,102 +172,36 @@ export interface Route {
 }
 
 type RouteRequest = { input: Input; session: Session }
```
asastre, line 174: Why not exporting `RouteRequest`?

packages/botonic-core/src/types/constants.d.ts
vanbasten17: yep, we should consider how to name this in Botonic 1.0, atm, some packages are using it, so we should maintain it like this just to not break things.

packages/botonic-core/src/types/core-bot.d.ts
```diff
@@ -0,0 +1,46 @@
...
+  BotResponse,
+  Locales,
+  PluginConfig,
+  Routes,
+  Session,
+} from '.'
+import { Inspector } from './debug'
+import { Router } from './router'
+
+export interface BotOptions {
```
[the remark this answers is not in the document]
asastre: It's still repeated.

packages/botonic-core/src/types/plugins.d.ts
```diff
@@ -0,0 +1,11 @@
...
+export declare function runPlugins(
+  plugins: PluginConfig<any>[],
+  mode: 'pre' | 'post',
```
asastre, line 6: should this be set into constants?

packages/botonic-core/src/types/router.d.ts
```diff
@@ -0,0 +1,57 @@
...
+        retryAction?: undefined
+        defaultAction?: undefined
+      }
+    | {
+        action: any
+        params: any
+        retryAction: any
+        lastRoutePath: any
+        defaultAction?: undefined
+      }
```
asastre, line 35: Can these 3 type of returns be categorized into other types or interfaces? This way maybe it will be understandable why there can be 3 type of returns.

asastre (changes requested):
Great job :clap:! I like grouping all types in the `types` folder!
Some observations/questions/suggestions before approving :wink:

elozano98 (approved):
good job! 👍🏽

packages/botonic-core/src/types/core-bot.d.ts
```diff
@@ -0,0 +1,45 @@
...
+  /** The plugin configurations */
+  appId?: string
+  defaultDelay?: number
+  defaultRoutes?: Routes
+  defaultTyping?: number
+  inspector?: Inspector
+  locales: Locales
+  plugins?: PluginConfig<any>
+  renderer: any
```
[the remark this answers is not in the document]
vanbasten17: renderer is the abstract object in charge of defining how the actions will be rendered. In case of `@botonic/react`, there we define a ReactRenderer

packages/botonic-core/src/types/plugins.d.ts
vanbasten17: added as `PluginMode`type.

packages/botonic-core/src/types/router.d.ts
vanbasten17: yes, I would like to take a closer look into this in the next iteration of moving some files to TS.

vanbasten17: with the current code of router, it's hard to discriminate what kind of objects we can have here or if they can be optimized.

asastre (approved):
:rocket:

PR #1342, opened 2021-03-02
Chore/fix telemetry

vanbasten17:
## Description
* Refactored Botonic CLI tracking logic to reduce errors and noise
* Remove unnecessary dependencies (`rimraf`, `folderHash`)
* Keep adding types
* Add extra info for users going into postinstall catch flow
* Use uuids instead of random numbers
* Get correct homedir (platform agnostic)
* Add tests

Remove generic utils in favour of:
 - Centralize file system operations into `util/file-system.ts`
 - Processes (spawn, exec) handling into `util/processes.ts` 

## Context
Right now, track for botonic-cli were not working as expected. Since credentials were not loaded correctly, every time a user did an install which generated, let's say id: `1234`, when going on next steps of the funnel the id was different, causing inconsistent results in telemetry.


packages/botonic-cli/src/botonic-api-service.ts
```diff
@@ -274,21 +216,21 @@ export class BotonicAPIService {
     return axios({ method: 'post', url: url, data: signup_data })
   }
 
-  async saveBot(bot_name: string) {
+  async saveBot(botName: string): Promise<AxiosPromise> {
     const resp = await this.api(
       'bots/',
-      { name: bot_name, framework: 'framework_botonic' },
+      { name: botName, framework: 'framework_botonic' },
```
ericmarcos, line 222: Can you remove the `framework` param? It's not used in the API anymore.

ericmarcos (approved):
Amazing job!!


giphy (1)

packages/botonic-cli/src/analytics/credentials-handler.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  initialize(): void {
+    this.createIfNotExists()
+  }
+
+  generateId(): string {
+    return uuidv4()
+  }
+
+  createIfNotExists(): void {
```
dpinol, line 37: good name!
what about createPathIfNotExists?

packages/botonic-cli/src/util/file-system.ts
```diff
@@ -0,0 +1,44 @@
+import {
```
dpinol, line 1: good work killing utils.ts! :clap:

dpinol (changes requested):

packages/botonic-cli/src/util/processes.ts
```diff
@@ -0,0 +1,83 @@
+import { execSync, spawn } from 'child_process'
```
dpinol, line 1: I suggest
1) renaming this file to system-info.ts
2) moving sleep & spawnProcess to a new file called eg. system.ts

dpinol (changes requested):

packages/botonic-cli/src/analytics/credentials-handler.ts
```diff
@@ -0,0 +1,123 @@
...
+      console.warn('Credentials could not be overwritten')
+    }
+  }
+}
+
+export class GlobalCredentialsHandler extends CredentialsHandler {
+  constructor() {
+    const homeDirectory = isWindows()
+      ? homedir()
+      : execCommand('eval echo ~${SUDO_USER}')
```
dpinol, line 65: if execCommand fails, GlobalCredentialsHandler will do weird things and it will be difficult to diagnose.
I suggest in a future PR:
* call here a version of execCommand that does not catch 
* In the telemetry constructor, in new GlobalCredentialsHandler(), catch it and set a flag (maybe reuse isAnalyticsEnabled) to disable the telemetry because the GlobalCredentialsHandler() is screwed

dpinol (changes requested):

packages/botonic-cli/src/constants.ts
```diff
@@ -0,0 +1,17 @@
+export const BOTONIC_NPM_NAMESPACE = '@botonic'
+export const BOTONIC_HOME_DIRNAME = '.botonic'
+export const GLOBAL_CREDS_FILENAME = 'credentials.json'
+export const BOTONIC_PROJECT_PATH = process.cwd()
+export const BOT_CREDS_FILENAME = '.botonic.json'
+export const ANALYTICS_WRITE_KEY = 'YD0jpJHNGW12uhLNbgB4wbdTRQ4Cy1Zu'
+export const TRACKING_EVENTS = {
```
dpinol, line 7: Having ALL constants in a "constants.ts" file is like having a single utils.ts file, a "cajon de sastre". See https://docs.google.com/document/d/1-qd2FdXbAe7XEbdTGyACeCs9HRCauCogocfEU-h-n-I/edit#bookmark=id.3fjmm64cwiqc
But since most of them are "important" :+1: I only suggest moving the TRACKING_EVENTS to somewhere closer to the analytics scope.

packages/botonic-cli/src/analytics/telemetry.ts
```diff
@@ -0,0 +1,83 @@
...
+      const anonymousId = this.createAnonymousIdIfNotExists()
+      this.analyticsService.track({
+        event: event,
+        anonymousId: anonymousId,
+        properties: this.getTelemetryProperties(properties),
+      })
+    }
+  }
+
+  trackLoggedIn(properties = {}): void {
```
dpinol, line 45: My 2 cents about the names.
* Normally it makes sense to name events in the paste "creatED" since you normally want to launch them after the action has completed. 
* But here you launch the tracks before completing the action. So it would make more sense to name the functions in present (trackLogIn)
Since it's too late to change the literals sent to analytics, I would leave them and just add a comment explaining the misalignment in the code that defines the constants

dpinol (changes requested):

packages/botonic-cli/tests/analytics/telemetry.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      'framework_path',
+      'system_path',
+      'node_version',
+      'botonic_cli_version',
+      'botonic_dependencies',
+      'addedField',
+    ])
+  })
+  it('Do telemetry calls properly', () => {
+    const track = (Telemetry.prototype.track = jest.fn())
```
dpinol, line 51: it looks like your asserting the call on the same method that you call. am I wrong?

packages/botonic-cli/tests/analytics/telemetry.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    const trackCreated = (Telemetry.prototype.trackCreated = jest.fn())
+    const trackServed = (Telemetry.prototype.trackServed = jest.fn())
+    const trackTested = (Telemetry.prototype.trackTested = jest.fn())
+    const trackTrained = (Telemetry.prototype.trackTrained = jest.fn())
+    const trackInstalledBotonic = (Telemetry.prototype.trackInstalledBotonic = jest.fn())
+    const trackDeployed = (Telemetry.prototype.trackDeployed = jest.fn())
+    const trackError = (Telemetry.prototype.trackError = jest.fn())
+
+    telemetry.track('some event', {})
+    telemetry.trackLoggedIn()
```
dpinol, line 63: advantes of changing the order of he test to repeat the following for each method

```
const tracxxx = ...
telemetry.trackXXX
expect(...)

```
1) you would detect if 2 internal calls are swapped
2) if you want to add or remove a call, editing is much easier

packages/botonic-cli/tests/analytics/telemetry.test.ts
```diff
@@ -0,0 +1,85 @@
+import { assert } from 'console'
+
+import { Telemetry } from '../../src/analytics/telemetry'
+import { pathExists } from '../../src/util/file-system'
+
+process.env.BOTONIC_DISABLE_ANALYTICS = '1'
+assert(process.env.BOTONIC_DISABLE_ANALYTICS === '1')
```
dpinol, line 7: why is there an assert after setting the variable?

packages/botonic-cli/tests/util/file-system.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+const existingPath = process.cwd()
+const unexistingPath = 'unexistingPath'
+const dirToTest = join(process.cwd(), 'botonic-tmp')
+
+const withDirToTest = (
+  toTest: () => any,
+  { createDir } = { createDir: true }
+): boolean => {
+  let success = false
```
dpinol, line 24: in many usages you don't check the returned boolean, and hence you won't realize if toTest() fails. Also, you won't even see any error callstack.
I suggest 2 alternatives:
* creating/removing the dir in beforeEach/afterEach
* Remove the try/catch and manage the expected faiures wih a expect/toThrow

In general, always first try using the tools provided by your framework.

packages/botonic-cli/tests/util/file-system.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      expect.stringMatching(/botonic-tmp*/)
+    })
+  })
+  it('Copy content', () => {
+    const tmp1 = createTemp('botonic-tmp1')
+    createFile(join(tmp1, 'dummy-file.txt'), 'dummy content')
+    const tmp2 = createTemp('botonic-tmp2')
+    copy(tmp1, tmp2)
+    expect(readDir(tmp2)).toContain('dummy-file.txt')
+    remove(tmp1)
```
dpinol, line 105: place removes in finally

packages/botonic-cli/tests/util/file-system.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    remove(tmp2)
+  })
+  it('Reads/Writes JSON correctly', () => {
+    const tmpDir = createTemp('botonic-tmp')
```
dpinol, line 109: use withTempDir here?

dpinol (changes requested):

packages/botonic-cli/src/util/file-system.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function pathExists(path: string): boolean {
+  return existsSync(path)
+}
+
+export function readDir(path: string): string[] {
+  return readdirSync(path)
+}
+
+export function readJSON(path: string): any {
```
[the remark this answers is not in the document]
vanbasten17: the fact is that an object is not JSON if we are strict.
e.g.: 
```
{ dummy: 'content' }
```
is not:
```
{ "dummy": "content"}
```
so when using for example writeJSON
```
writeJSON(path, {
      dummy: 'content',
} as JSON)
``` 
won't work: `Conversion of type '{ dummy: string; }' to type 'JSON' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.`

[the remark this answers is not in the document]
vanbasten17: just read your comment for typing JSON 👍

packages/botonic-cli/src/util/file-system.ts
```diff
@@ -0,0 +1,44 @@
...
+  const fileContent = readFileSync(path, 'utf8')
+  if (!fileContent) return undefined
```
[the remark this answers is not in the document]
vanbasten17: this check is necessary in order to not doing JSON.parse('') of an empty content, o.w. it will cause an error which will break the entire developer flow.

packages/botonic-cli/src/analytics/credentials-handler.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (!pathExists(this.homePath)) create(this.homePath)
+  }
+
+  load(): any {
+    try {
+      return (
+        pathExists(this.pathToCredentials) && readJSON(this.pathToCredentials)
+      )
+    } catch (e) {
+      console.warn('Credentials could not be loaded')
```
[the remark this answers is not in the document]
vanbasten17: I decided not to print this exception because the developer flow continues working as expected despite of not being available of loading the credentials and don't want to show a large error within the flow (the handicap here, and this fail will be due to user permissions is that for every deploy, the user must login again).

packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -277,24 +278,24 @@ Uploading...
       text: 'Creating bundle...',
       spinner: 'bouncingBar',
     }).start()
-    // Why using rimraf instead of fs.rmdirSync?: https://stackoverflow.com/a/57834155/145289
-    if (fs.existsSync('tmp')) rimraf.sync('tmp')
-    fs.mkdirSync(join('tmp'))
-    copySync('dist', join('tmp', 'dist'))
+    if (pathExists('tmp')) remove('tmp')
```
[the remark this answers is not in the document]
vanbasten17: if for some reason the temporary directory is not removed, with createTempDir we are going to generate more than one temporary directories in the developer project. What about calling it `botonic-tmp` in order to minimize the probability of collisioning with an already existing `tmp`?

packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -277,24 +278,24 @@ Uploading...
...
+    create(join('tmp'))
```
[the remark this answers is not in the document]
vanbasten17: yes, it should create it in `join(process.cwd(), 'tmp')`. Fixed now. (by default it will create the dir in process.cwd())

packages/botonic-cli/tests/analytics/telemetry.test.ts
vanbasten17: taken from this example: https://stackoverflow.com/a/50108368, first do the assigment in the right parentheses, then assign into track variable.

packages/botonic-cli/tests/analytics/telemetry.test.ts
vanbasten17: not needed rewriting the variable, as it is already in our secrets, leaving the assert in order to catch that this is not executed if we have analytics enabled.

packages/botonic-cli/tests/analytics/telemetry.test.ts
vanbasten17: put them in order 👍

packages/botonic-cli/src/analytics/credentials-handler.ts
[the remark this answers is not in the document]
dpinol: I understand, but hiding exceptions is a no-no. I suggest implementing here  a simple function which only displays the error message but not the call stack, and only do it once per botonic execution to minimize spamming. I think you implemented something similar in botonic, but here it can be much simpler as a start.
```
let credentialsFailed = false
function logCredentialsFailure(msg: string, e) {
  if (!credentialsFailed) {
     console.warng(msg +  ":" + str(e)
    credentialsFailed = true
}
}

```

packages/botonic-cli/src/commands/deploy.ts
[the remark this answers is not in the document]
dpinol: Well, createTempDir could be in the top of the cli command with a finally to remove it. And leaving rubbish in tmp dir is not so bad because user will eventually remove it. I prefer rubbish there rather than in my bot project.
But in this PR let's **not do anything** please. Just create a linear with a copy of this discussion and let's think about it

packages/botonic-cli/tests/analytics/telemetry.test.ts
dpinol: call me , I don't get it

packages/botonic-cli/src/analytics/telemetry.ts
```diff
@@ -1,10 +1,24 @@
...
 
-import { ANALYTICS_WRITE_KEY, TRACKING_EVENTS } from '../constants'
+import { ANALYTICS_WRITE_KEY } from '../constants'
 import { AnalyticsService } from '../interfaces'
 import { getSystemInformation } from '../util/processes'
 import { GlobalCredentialsHandler } from './credentials-handler'
 
+const TRACKING_EVENTS = {
+  // BEWARE: Changing these names will screw analytics
+  // Despite the event is written in past tense, events are tracked at the very beginning of the action.
```
dpinol, line 10: :clap:

packages/botonic-cli/src/commands/deploy.ts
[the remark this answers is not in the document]
vanbasten17: annotated 👍

packages/botonic-cli/src/util/processes.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export async function sleep(ms: number): Promise<number> {
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
+export function execCommand(command: string): string {
+  try {
+    return String(execSync(command)).trim()
+  } catch (e) {
+    return String(e)
```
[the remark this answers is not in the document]
vanbasten17: added a function `execCommandSafe` which uses `execCommand` that can fail

packages/botonic-cli/src/analytics/credentials-handler.ts
vanbasten17: done 👍

packages/botonic-cli/src/analytics/telemetry.ts
```diff
@@ -0,0 +1,83 @@
...
+      ...properties,
+    }
+  }
+
+  track(event: string, properties = {}): void {
+    if (this.isAnalyticsEnabled) {
+      const anonymousId = this.createAnonymousIdIfNotExists()
+      this.analyticsService.track({
+        event: event,
+        anonymousId: anonymousId,
```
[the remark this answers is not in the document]
vanbasten17: typed 👍

packages/botonic-cli/src/analytics/credentials-handler.ts
```diff
@@ -0,0 +1,123 @@
...
+    this.createIfNotExists()
+    if (!pathExists(this.pathToCredentials) || !this.hasAnonymousId()) {
+      this.refreshAnonymousId()
+    }
+  }
+
+  getAnonymousId(): string {
+    try {
+      const content = this.load()
+      return content?.analytics?.anonymous_id
```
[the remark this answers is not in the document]
vanbasten17: what i've done: `getAnonymousId` returns `string or undefined`.
Then `refreshAnonymousId` returns the new `id` generated

packages/botonic-cli/src/analytics/credentials-handler.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  refreshAnonymousId(): void {
+    this.dump({
+      analytics: { anonymous_id: this.generateId() },
+    })
+  }
+
+  load(): GlobalCredentials {
+    return super.load() as GlobalCredentials
```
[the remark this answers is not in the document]
vanbasten17: typed with JSON 👍

packages/botonic-cli/src/analytics/credentials-handler.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  generateId(): string {
+    return uuidv4()
+  }
+
+  createIfNotExists(): void {
+    if (!pathExists(this.homePath)) create(this.homePath)
+  }
+
+  load(): any {
```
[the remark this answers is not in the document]
vanbasten17: adapted one example from the given resource:
```
export type JSONPrimitive = string | number | boolean | null
export type JSONValue = JSONPrimitive | JSONObject | JSONArray
export type JSONObject = { [member: string]: JSONValue }
export type JSONArray = JSONValue[]
```

vanbasten17:
feedback applied @dpinol , thanks for your exhaustive review!! 💯

packages/botonic-cli/src/analytics/credentials-handler.ts
```diff
@@ -111,11 +112,11 @@ export class BotCredentialsHandler extends CredentialsHandler {
     })
   }
 
-  load(): BotCredentials {
-    return super.load() as BotCredentials
+  load(): BotCredentials | undefined {
```
dpinol, line 115: same as previous load

packages/botonic-cli/tests/analytics/telemetry.test.ts
```diff
@@ -15,18 +16,25 @@ describe('TEST: Telemetry', () => {
   it('Initializes correctly Telemetry', () => {
     expect(telemetry.isAnalyticsEnabled).toBe(false)
     expect(telemetry.analyticsService).toBeTruthy()
-    expect(pathExists(telemetry.globalCredentialsHandler.homeDir)).toBeTruthy()
+    let homeDir = ''
+    if (telemetry.globalCredentialsHandler) {
+      homeDir = telemetry.globalCredentialsHandler.homeDir
```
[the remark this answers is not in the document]
vanbasten17: makes sense, applied 👍

dpinol (changes requested):
see comments

dpinol (approved):

PR #1343, opened 2021-03-02
Contentful/fix tests

dpinol:
## Description

Fix contentful tests:
*  The QA contentful space was reimported by mistake from an old export, which contained a different asset id. This PR changes an expectation to reflect the current asset id
*   ManageCms.updateFields test has been improved: Now we have individual tests for setting "", undefined and a non empty value. Now we finally set to undefined instead of "" to avoid that validator fails when it finds empty values. Not sure why the validator test didn't fail before. Maybe we had recently disabled English as the fallback of Spanish? Maybe

## Context

* Contentful behaves differently when a string field is set to "" or when set to undefined. "" is interpreted as enforcing empty text. Undefined is interpreted as the default value, which causes delivery to default to the fallback locale (if field is configured as i18nable). However, there's no way to make this distinction from the contentful.com console (where it's always interpreted as default to fallback locale) .

Improve ManageCms.updateFields readbility.

## Testing

- has integration tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/cms/contents.ts
+++ b/packages/botonic-plugin-contentful/src/cms/contents.ts
@@ -1,6 +1,6 @@
 import * as time from '../time'
 import { shallowClone, Stringable } from '../util/objects'
-import { Callback, ContentCallback, TopContentId } from './callback'
+import { Callback, ContentCallback, ContentId, TopContentId } from './callback'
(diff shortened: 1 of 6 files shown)


vanbasten17 (approved):

elozano98 (approved):
LGTM :+1:

PR #1355, opened 2021-03-08
fix(react): update lastMessageUpdate consistently with Webchat Settings

vanbasten17:
## Description
* Update logic to be consistent when resending WebchatSettings.

Linked PR: https://github.com/metis-ai/hubtype-backend/pull/552

### DIFF
--- a/packages/botonic-react/src/webchat/webchat.jsx
+++ b/packages/botonic-react/src/webchat/webchat.jsx
@@ -581,6 +581,7 @@ export const Webchat = forwardRef((props, ref) => {
     updateWebchatSettings: settings => {
       const themeUpdates = normalizeWebchatSettings(settings)
       updateTheme(merge(webchatState.theme, themeUpdates), themeUpdates)
+      updateLastMessageDate(currentDateString())
     },
   }))
 


asastre (commented):
Since `WebchatSettings` is not a message by itself I don't know if it make sense to update the last message date. What do you think?

vanbasten17:
@asastre it's true that is not a message itself, but when resending unread messages from backend, the last message update date should be taken into account in order to not send additional WebchatSettings messages.

asastre (approved):

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -581,6 +581,7 @@ export const Webchat = forwardRef((props, ref) => {
     updateWebchatSettings: settings => {
       const themeUpdates = normalizeWebchatSettings(settings)
       updateTheme(merge(webchatState.theme, themeUpdates), themeUpdates)
+      updateLastMessageDate(new Date())
```
dpinol, line 584: does this depend on #1360? there you're sending new Date().toISOString

dpinol (changes requested):

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: yes, it is related and it depends on it (but this one was created before). Changing this PR to draft

vanbasten17:
@asastre , you were right, there is no need to update the lastMessageDate. When doing the auth call, last message id will be the previous one just before the WebchatSettings in DB. So I'm closing this PR atm. It's a backend issue.

PR #1360, opened 2021-03-12
fix(core/react): retrieve agent messages after connection loss and HubtypeService refactor

vanbasten17:
## Description
* Solve issue regarding enduser not receiving some agent messages when connection was lost.

## Approach
* Moving onConnectionRegained logic within HubtypeService. Every time the event subcription_succeeded is received, we can asure that the auth call has been done. Hence, onConnectionRegained should be called at that point.
* Pusher headers are not refreshed automatically, only when `HubtypeService.init()` is called. So when the state of pusher is `connecting` we must be sure to update the corresponding headers.

Considerations:
* The first approach of destroying pusher and then initialize it again in `onConnectionRegained` breaks the capability of pusher to detect that the application has been reconnected.
* The process is the following:
Connecting --> update auth headers --> Connected --> Do auth call --> subscription:suceeded --> call logic when connection is regained

**References**:
* Authenticating users: https://pusher.com/docs/channels/server_api/authenticating-users
* Connection states: https://pusher.com/docs/channels/using_channels/connection#connection-states

Next steps:
* Randomly, one of the messages is duplicated in messagesComponents object, but not in messagesJSON (which is consistent). https://linear.app/hubtype/issue/BOT-295/investigate-duplicated-messages-randomly-when-connection-is-recovered

### DIFF
--- a/packages/botonic-core/src/hubtype-service.js
+++ b/packages/botonic-core/src/hubtype-service.js
@@ -71,9 +71,7 @@ export class HubtypeService {
       const cleanAndReject = msg => {
         // eslint-disable-next-line @typescript-eslint/no-use-before-define
         clearTimeout(connectTimeout)
-        this.pusher.connection.unbind()
-        this.channel.unbind()
-        this.pusher = null
+        this.destroyPusher()
         reject(msg)
       }
       const connectTimeout = setTimeout(
@@ -86,18 +84,17 @@ export class HubtypeService {
       })
       this.pusher.connection.bind('error', event => {
         if (event.type == 'WebSocketError') this.handleConnectionChange(false)
-        const errorMsg =
-          event.error && event.error.data
-            ? event.error.data.code || event.error.data.message
-            : 'Connection error'
-        cleanAndReject(`Pusher error (${errorMsg})`)
+        else {
+          const errorMsg =
+            event.error && event.error.data
+              ? event.error.data.code || event.error.data.message
+              : 'Connection error'
+          cleanAndReject(`Pusher error (${errorMsg})`)
+        }
       })
     })
     this.pusher.connection.bind('connected', () =>
       this.handleConnectionChange(true)
-    )
-    this.pusher.connection.bind('disconnected', () =>
-      this.handleConnectionChange(false)
     )
     this.pusher.connection.bind('unavailable', () =>
       this.handleConnectionChange(false)
@@ -149,7 +146,6 @@ export class HubtypeService {
       await this.init(user)
     } catch (e) {
       this.handleUnsentInput(message)
-      this.handleConnectionChange(false)
       return Promise.resolve()
     }
     try {
@@ -182,13 +178,14 @@ export class HubtypeService {
     this.pusher.disconnect()
     this.pusher.unsubscribe(this.pusherChannel)
     this.pusher.unbind_all()
+    this.pusher.channels = {}
     this.pusher = null
   }
 
   async onConnectionRegained() {
     this.destroyPusher()
-    this.init()
-    this.resendUnsentInputs()
+    await this.init()
+    await this.resendUnsentInputs()
   }
 
   async resendUnsentInputs() {
--- a/packages/botonic-core/src/hubtype-service.js
+++ b/packages/botonic-core/src/hubtype-service.js
@@ -151,28 +151,20 @@ export class HubtypeService {
     })
   }
 
-  // eslint-disable-next-line consistent-return
   async postMessage(user, message) {
     try {
       await this.init(user)
-    } catch (e) {
-      this.handleUnsentInput(message)
-      return Promise.resolve()
-    }
-    try {
-      return axios
-        .post(
-          `${_HUBTYPE_API_URL_}/v1/provider_accounts/webhooks/webchat/${this.appId}/`,
-          {
-            sender: this.user,
-            message: message,
-          }
-        )
-        .then(res => {
-          if (res && res.status === 200) this.handleSentInput(message)
-          return
-        })
-        .catch(e => this.handleUnsentInput(message))
+      await axios.post(
+        `${_HUBTYPE_API_URL_}/v1/provider_accounts/webhooks/webchat/${this.appId}/`,
+        {
+          sender: this.user,
+          message: message,
+        },
+        {
+          validateStatus: status => status === 200,
+        }
+      )
+      this.handleSentInput(message)
     } catch (e) {
       this.handleUnsentInput(message)
     }
--- a/packages/botonic-core/src/hubtype-service.js
+++ b/packages/botonic-core/src/hubtype-service.js
@@ -19,6 +19,10 @@ const _HUBTYPE_API_URL_ = getWebpackEnvVar(
 const ACTIVITY_TIMEOUT = 20 * 1000 // https://pusher.com/docs/channels/using_channels/connection#activitytimeout-integer-
 const PONG_TIMEOUT = 5 * 1000 // https://pusher.com/docs/channels/using_channels/connection#pongtimeout-integer-
 export class HubtypeService {
+  user
+  lastMessageId
+  lastMessageUpdateDate
+
   constructor({
     appId,
     user,
@@ -175,6 +179,20 @@ export class HubtypeService {
     return axios.get(
       `${_HUBTYPE_API_URL_}/v1/provider_accounts/${appId}/visibility/`
     )
+  }
+
+  destroyPusher() {
+    if (!this.pusher) return
+    this.pusher.disconnect()
+    this.pusher.unsubscribe(this.pusherChannel)
+    this.pusher.unbind_all()
+    this.pusher = null
+  }
+
+  async onConnectionRegained() {
+    this.destroyPusher()
+    this.init()
+    this.resendUnsentInputs()
   }
 
   async resendUnsentInputs() {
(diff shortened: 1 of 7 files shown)


packages/botonic-core/src/hubtype-service.js
```diff
@@ -146,6 +150,20 @@ export class HubtypeService {
...
 
+  destroyPusher() {
+    if (!this.pusher) return
+    this.pusher.disconnect()
+    this.pusher.unsubscribe(this.pusherChannel)
+    this.pusher.unbind_all()
+    this.pusher = null
+  }
+
+  async onConnectionRegained() {
```
dpinol, line 161: :100:

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -524,7 +524,7 @@ export const Webchat = forwardRef((props, ref) => {
     if (isMedia(input)) input.data = await readDataURL(input.data)
     sendUserInput(input)
     updateLatestInput(input)
-    updateLastMessageDate(new Date())
+    isOnline && updateLastMessageDate(new Date().toISOString())
```
dpinol, line 527: do we need to add updateLastMessageDate to somewhere in index.d.ts?

dpinol (changes requested):
can fix the type of  WebchatState.lastMessageUpdate in webchat/index.d.ts? now it's undefined

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -550,7 +550,7 @@ export const Webchat = forwardRef((props, ref) => {
         updateHandoff(handoff)
       }
       if (lastRoutePath) updateLastRoutePath(lastRoutePath)
-      updateLastMessageDate(new Date())
+      isOnline && updateLastMessageDate(new Date().toISOString())
```
dpinol, line 553: where is WebchatState.lastMessageUpdate? only by the bot?
Would it be possible to store the Date instead of the string? or would we have issues with browser state?

asastre (approved):
:clap: :clap:

packages/botonic-core/src/hubtype-service.js
```diff
@@ -19,6 +19,10 @@ const _HUBTYPE_API_URL_ = getWebpackEnvVar(
 const ACTIVITY_TIMEOUT = 20 * 1000 // https://pusher.com/docs/channels/using_channels/connection#activitytimeout-integer-
 const PONG_TIMEOUT = 5 * 1000 // https://pusher.com/docs/channels/using_channels/connection#pongtimeout-integer-
 export class HubtypeService {
+  user
```
dpinol, line 22: afaik this is not a common idiom in js. Like in python, IDEs (and developers) understand the idiom of using as class fields those assigned in the constructor. If we migrate ever migrate to typescript, we'll see then what's the best way to declare the fields here.

packages/botonic-core/src/hubtype-service.js
```diff
@@ -177,6 +181,20 @@ export class HubtypeService {
     )
   }
 
+  destroyPusher() {
+    if (!this.pusher) return
```
dpinol, line 185: shouldn't we call destroyPusher from cleanAndReject in line 75?

dpinol (changes requested):

packages/botonic-core/src/hubtype-service.js
vanbasten17: cool, removing it 👍

packages/botonic-react/src/webchat-app.jsx
```diff
@@ -109,24 +109,32 @@ export class WebchatApp {
...
+          unsentInputs: () =>
+            this.webchatRef.current
+              .getMessages()
+              .filter(msg => msg.ack === 0 && msg.unsentInput),
+        })
+      }
+    } else {
+      this.hubtypeService.lastMessageId = lastMessageId
+      this.hubtypeService.lastMessageUpdateDate = lastMessageUpdateDate
+      this.hubtypeService.user = user
```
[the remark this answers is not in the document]
vanbasten17: maybe not, as user should be preserved, removing this one 👍 , good catch

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: storing Date may cause issues in localStorage, it's better to store a stringified iso version of the timestamp

packages/botonic-core/src/hubtype-service.js
vanbasten17: calling it now

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: added 👍

packages/botonic-core/src/hubtype-service.js
```diff
@@ -149,7 +146,6 @@ export class HubtypeService {
       await this.init(user)
     } catch (e) {
       this.handleUnsentInput(message)
```
dpinol, line 159: * we discussed this in the past. any reason for postMessage function having 2 try/catch?
* Remove return Promise.resolve() and //eslint-disable-next-line consistent-return
* convert axios .then.catch to await to simplify code
* You will then only need 1 this.handleUnsentInput calls instead of 3

dpinol (changes requested):

packages/botonic-core/src/hubtype-service.js
vanbasten17: solved, I think you will like this refactor :)

packages/botonic-react/src/components/message.jsx
```diff
@@ -121,13 +121,10 @@ export const Message = props => {
...
-    if (isDev) ack = 1
-    else {
-      if (props.ack !== undefined) ack = props.ack
-      else ack = isFromUser ? 0 : 1
-    }
-    return ack
+    if (isDev) return 1
+    if (!isFromUser) return 1
+    if (props.ack !== undefined) return props.ack
+    return 0
```
ericmarcos, line 127: Thank you =)

packages/botonic-react/src/webchat/messages-reducer.js
```diff
@@ -14,7 +14,10 @@ export const messagesReducer = (state, action) => {
     case ADD_MESSAGE_COMPONENT:
       return {
         ...state,
-        messagesComponents: [...state.messagesComponents, action.payload],
+        messagesComponents: [
+          ...(state.messagesComponents || []),
+          action.payload,
+        ],
```
ericmarcos, line 20: Why the message ID duplicate check is not included in this PR?

ericmarcos (approved):
Nice work!

packages/botonic-react/src/webchat/messages-reducer.js
vanbasten17: I tried to added it here, but the logic we added was not working correctly. I prefer to close this PR which is getting large and keep investigating this issue aside in another PR (I added a task for it).

vanbasten17: added 👍

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -524,7 +521,7 @@ export const Webchat = forwardRef((props, ref) => {
     if (isMedia(input)) input.data = await readDataURL(input.data)
     sendUserInput(input)
     updateLatestInput(input)
-    updateLastMessageDate(new Date())
+    isOnline() && updateLastMessageDate(new Date().toISOString())
```
marcpdw, line 524: nit: `const currentDateString = () => new Date().toISOString()`

packages/botonic-react/src/webchat/messages-reducer.js
```diff
@@ -7,14 +7,30 @@ import {
...
+          m => getMsgComponentId(m) === action.payload.id
+        )
+      )
+        return state
       return {
         ...state,
-        messagesComponents: [...state.messagesComponents, action.payload],
+        messagesComponents: [
+          ...(state.messagesComponents || []),
+          action.payload,
```
marcpdw, line 32: Any risk for this to be `null | undefined` and have a parasite within this new array?

packages/botonic-react/src/webchat-app.jsx
```diff
@@ -111,25 +111,32 @@ export class WebchatApp {
...
-        unsentInputs: () =>
-          this.webchatRef.current
-            .getMessages()
-            .filter(msg => msg.ack === 0 && msg.unsentInput),
-        server: this.server,
-      })
+    const lastMessage = messagesJSON[messagesJSON.length - 1]
+    const lastMessageId = lastMessage && lastMessage.id
+    const lastMessageUpdateDate = this.getLastMessageUpdate()
+    if (!this.hubtypeService) {
```
marcpdw, line 122: nit: this might seem very picky, but it improves the readability. I'd start with the simple case and add the other condition altogether next:

```
if (this.hubtypeService) {
   ...
} else if (!this.hubtypeService && user) {
   ...
}
```

marcpdw (approved):
It looks good overall, but have a look to one of my comments to be sure we can't have a potential bug there

packages/botonic-react/src/webchat-app.jsx
vanbasten17: cool! 😎

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: added 👍

packages/botonic-react/src/webchat/messages-reducer.js
vanbasten17: not needed at all

PR #1370, opened 2021-03-15
feat(core): improvements in HubtypeService

dpinol:
Depends on #1360 
## Description
* Throw an exception when webchat has lost user or appId from session.
* Small HubtypeService refactor to simplify the code
* HubtypeService.init() to fail when user or appId are missing 
* Simple test for HubtypeService

## Context

So far we we were skipping silently.  and it took to long to diagnose it

Cleanup code, reject the Promise when there are issues
Added unit tests for happy and sad paths of init methos. 

## Testing

- has unit tests

### DIFF
--- a/packages/botonic-core/src/hubtype-service.js
+++ b/packages/botonic-core/src/hubtype-service.js
@@ -18,7 +18,12 @@ const _HUBTYPE_API_URL_ = getWebpackEnvVar(
 
 const ACTIVITY_TIMEOUT = 20 * 1000 // https://pusher.com/docs/channels/using_channels/connection#activitytimeout-integer-
 const PONG_TIMEOUT = 5 * 1000 // https://pusher.com/docs/channels/using_channels/connection#pongtimeout-integer-
+
+/**
+ * Calls Hubtype APIs from Webchat
+ */
 export class HubtypeService {
+  PUSHER_CONNECT_TIMEOUT_MS = 10000
   constructor({
     appId,
     user,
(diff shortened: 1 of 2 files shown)


vanbasten17 (approved):
LGTM, I think you should rebase with master in order to include the latest changes in HubtypeService

packages/botonic-core/src/hubtype-service.js
```diff
@@ -35,36 +40,34 @@ export class HubtypeService {
...
-  }
-
+  /**
+   * @returns {Promise<void>}
+   */
   init(user, lastMessageId, lastMessageUpdateDate) {
     if (user) this.user = user
     if (lastMessageId) this.lastMessageId = lastMessageId
     if (lastMessageUpdateDate)
       this.lastMessageUpdateDate = lastMessageUpdateDate
```
marcpdw, line 58: Question here: are you ensuring those properties are not overwritten in case they are passed `null` | `undefined`, so they keep the values set on the constructor if they were provided?

It looks weird passing mandatory values that may be set `null` on purpose and then it sends different ones.

packages/botonic-core/src/hubtype-service.test.js
```diff
@@ -0,0 +1,23 @@
...
+describe('HubtypeService', () => {
+  test.each([
+    [{ user: { id: 'userId' } }],
+    [{ appId: 'appId' }],
+    [{ appId: '', user: { id: 'userId' } }],
+    [{ appId: 'appId', user: { id: undefined } }],
+  ])('init fails if user or appid are empty: %s', async args => {
+    const sut = new HubtypeService(args)
+
+    await expect(sut.init()).rejects.toEqual(
```
marcpdw, line 12: Can you call init with no parameters? This is not TS but ES6, right? Getting confused here

marcpdw (commented):
Looks good overall, but I have a couple of questions regarding the syntax used here

packages/botonic-core/src/hubtype-service.js
dpinol: I agree it's not a great design, but I understand that the PR is not affecting this. In the future I'd also like to apply SRP and separate the Pusher part to its own class

packages/botonic-core/src/hubtype-service.test.js
dpinol: yes, it's ES6. We're migrating anyway to TS in #1419 (actually it will be moved to another PR)

marcpdw (approved):
Thank your for the clarifications. Let's roll it out and we can discuss the design eventually!

PR #1373, opened 2021-03-16
feat(core/react): detect disconnection (pusher-based) and add minor improvements

vanbasten17:
## Description
* Using pusher-js to detect disconnection instead of connection-checker. Removed connection-checker dependency
* Use [`activityTimeout`](https://pusher.com/docs/channels/using_channels/connection#activitytimeout-integer-) and [`pongTimeout`](https://pusher.com/docs/channels/using_channels/connection#pongtimeout-integer-) to accelerate pusher checks to detect disconnection.
* Rely on `webchatState.online` to handle connected/disconnected logic
* Let `server` object configurable:
* Added `isOnline` as part of Botonic's public API
* Added a way to control message to be displayed when there are connection issues
## Documentation
1. Configuring server options in order to handle connection/disconnections timings.

```
export const webchat = {
  server: {
    activityTimeout: 50 * 1000,
    pongTimeout: 20 * 1000,
  },
}

```
* `activityTimeout`: Time to wait (in ms) before pinging the server to know if the connection is still working.
* `pongTimeout`: After pinging the server, time to wait (in ms) before assuming the connection is dead and closing it.

2. Added a new `isOnline` in public API in order to know the state of Botonic (it will depend on how you have configured `server`).
e.g:
```
Botonic.isOnline() --> true/false
```

3. Added `errorMessage` (function or variable) inside server to change display text when user has connection issues.

```
export const webchat = {
  server: {
    activityTimeout: 20 * 1000,
    pongTimeout: 5 * 1000,
    errorMessage: 'Problemas de conexión',
  },
}

```
Screenshot 2021-03-16 at 13 57 04


dpinol (approved):

asastre (approved):
:clap: :clap:

dpinol (approved):

packages/botonic-react/src/webchat-app.jsx
```diff
@@ -126,13 +128,14 @@ export class WebchatApp {
...
             .filter(msg => msg.ack === 0 && msg.unsentInput),
+        server: this.server,
       })
     }
   }
 
   onServiceEvent(event) {
-    if (event.isError)
-      this.webchatRef.current.setError({ message: event.errorMessage })
+    if (event.action == 'connectionChange')
```
marcpdw, line 137: nit: `===` as good practise

marcpdw (approved):
Looks really good. The only improvements I might think of are related to constant/enum declaration to have all cases enumerated.
I know this will be introduced later on though, so good job here! 👍

packages/botonic-react/src/webchat-app.jsx
vanbasten17: nice observation! thanks, we missed it!

PR #1375, opened 2021-03-17
Plugin-contentful: add custom fields

AlbertGom:
## Description

- Added custom fields in order to get any field from contentful without having to add it to the plugin.
- Added integration tests for each top content to validate that when we define in contentful a new field, then we will be able to access it through customFields field.

## Context
In some cases, the bots need some custom fields from contentful that are not defined in the plugin. So wee needed a way to always be able to get these custom fields.

Added the addCustomFields function (delivery-utils.ts file) to all the top contents. This function gets all the fields that are not defined in the common fields of the plugin and adds them in the customFields field.

Sometimes the bots need some custom information related to a content. For example, for the image content model, you might want to add a custom field called description that contains the description of the image.
With customFields, the user only needs to add the custom field in the CMS and it will appear to the bot without needing extra configuration.

Example:

const image = cms.image("CONTENT_ID")
const imageDescription = image.customFields.description

## Testing

The pull request...

- has unit tests


packages/botonic-plugin-contentful/src/contentful/delivery-utils.ts
```diff
@@ -83,13 +83,16 @@ export class ContentfulEntryUtils {
...
+    const customFields: Record<string, unknown> = {}
+
     return new CommonFields(entry.sys.id, fields.name || '', {
       keywords: fields.keywords,
       shortText: fields.shortText,
       partitions: fields.partitions,
       searchableBy,
       dateRange,
       followUp,
+      customFields,
```
dpinol, line 95: remove, since it will be automatically set to {}.
Add here a comment saying "customFields cannot be easily diferenciated from standard fields until the content is created. see addCustomFields."

packages/botonic-plugin-contentful/src/cms/contents.ts
```diff
@@ -171,6 +172,7 @@ export class CommonFields implements Stringable {
       partitions?: string[]
       dateRange?: DateRangeContent
       followUp?: FollowUp
+      customFields?: Record<string, unknown>
```
dpinol, line 175: add a comment:
"customFields are any fields found in CMS but not expected in the corresponding model".
@vanbasten17 @guinii @asastre should we call it customFields or extraFields?

dpinol (changes requested):
go ahead adding it for all topContents!

packages/botonic-plugin-contentful/src/cms/contents.ts
asastre: I'm ok with `customFields`.

dpinol:
As this is a new feature:
* Add the "document" label to the PR
* Fill in the "To document / Usage example" section of the PR description so that it can be incorporated to botonic documentation

packages/botonic-plugin-contentful/src/contentful/contents/startup.ts
```diff
@@ -34,11 +34,15 @@ export class StartUpDelivery extends DeliveryWithFollowUp {
...
-      buttons
+    return addCustomFields(
+      new cms.StartUp(
+        await this.getFollowUp().commonFields(entry, context),
+        this.urlFromAssetOptional(fields.pic, context),
+        fields.text ?? '',
+        buttons
+      ),
+      fields,
+      ['pic', 'followup']
```
dpinol, line 45: [dry](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself): since followup is common for all types, add it within addCustomFields (with a comment on the different spelling with the contentful.com field name)

dpinol (changes requested):

AlbertGom:
@dpinol updated 'to document' section!

dpinol (approved):

packages/botonic-plugin-contentful/src/contentful/delivery-utils.ts
```diff
@@ -144,3 +145,22 @@ export function convertContentfulException(e: any, query: any): any {
...
+  content: T,
+  entryFields: CommonEntryFields,
+  ignoreFields?: string[]
+): T {
+  const customKeys = Object.keys(entryFields).filter(
+    f =>
+      !Object.keys(content).includes(f) &&
+      !Object.keys(content.common).includes(f) &&
+      !ignoreFields?.includes(f) &&
+      //contentful: followup, plugin: followUp
```
vanbasten17, line 159: remove console.log?

vanbasten17 (approved):

packages/botonic-plugin-contentful/src/contentful/delivery-utils.ts
AlbertGom: This is info to indicate that it has to ignore 'followup' content name. We are receiving 'followup' from contentful but in the plugin is named followUp

PR #1376, opened 2021-03-17
docs: fix example and add info in readme

annehubtype:
2 doc-related updates:
- In the Multichannel component doc, the syntax example has been fixed.
- In the Readme GitHub file, 2 titles have been enhanced for better seo.

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 <p align="center" style="text-align: center;"><a href="https://botonic.io">Website</a> |  <a href="https://botonic.io/blog">Blog</a>  |  <a href="https://botonic.io/docs/welcome">Docs</a> |  <a href="https://botonic.io/examples">Examples</a>  |  <a href="https://botonic.io/blog/2021/03/10/introducing-botonic-office-hours">Friday Office Hours</a></p>
 <hr>
 
-## <center>Build Chatbots and Conversational Apps Using React</center>
+# <center>Build Chatbots and Conversational Apps Using React</center>
 
 [![botonic](https://img.shields.io/badge/cli-botonic-brightgreen.svg)](https://botonic.io)
 [![Version](https://img.shields.io/npm/v/@botonic/cli.svg)](https://npmjs.org/package/@botonic/cli)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -15,7 +15,7 @@
 <a target="_blank" rel="noopener noreferrer" href="https://twitter.com/intent/follow?screen_name=botonic_"><img src="https://camo.githubusercontent.com/9e04e9647d574f9e2ad27f93d6eade8f5e9d6b9fed6c50ad6399742b7713ba1f/68747470733a2f2f696d672e736869656c64732e696f2f747769747465722f666f6c6c6f772f626f746f6e69635f3f7374796c653d736f6369616c" alt="Twitter Follow" data-canonical-src="https://img.shields.io/twitter/follow/botonic_?style=social" style="max-width:100%;"></a>
 
 
-# 🐣 What is Botonic?
+# 🐣 What is Botonic Open Source Framework?
```
asastre, line 18: Would it be better to add the open source inside the description rather than in the title?
Something like this:

> [Botonic](https://botonic.io) is an open source full-stack framework to create chatbots...

I understand that this change has been made for a better SEO position, but for me it looks a little odd to have part of the answer in the title itself. What do you think?

annehubtype: I agree. Other suggestion: instead of a question, what about "Introducing Botonic Open Source Framework"?

asastre: If it's good to have this "open source" concept in the title for SEO improvements, then yes, I agree with you and I think that it's better to put it as you suggest, without being a question.

asastre: Maybe I would also add the "open source" concept into the description, as I suggested in the comment before, what do you think? 
Right now I don't see anywhere in the README file that the framework is open source and I think we should have it in the description also.

annehubtype: @asastre Totally, I'm adding it right now :)

asastre (approved):
:+1:

docs/docs/components/components-multichannel.md
```diff
@@ -26,52 +39,28 @@ As an example, it can be used to adapt native representations of a Text with But
...
-
-- `boldIndex`: Applies a bold format to the index element. True or False.
-
-- `Carousel`: Adds a letter or number to the index for a carousel.
-
-- `Text` : Adds a letter or number to the index for a text.
-
-- `Indexseparator`: Adds a dot or dash after the letter or number.
-
-- `Message separator` : Adds a space between the various messages. `{'\n\n'}` adds a line break between the index message and the selection option.
```
vanbasten17, line 56: Why removing descriptions for attributes?

docs/docs/components/components-multichannel.md
```diff
@@ -26,52 +39,28 @@ As an example, it can be used to adapt native representations of a Text with But
...
         Button 2
-      </Button>,
+      </Button>
       <Button key={'3'} url='http://testurl.com'>
         Visit website
-      </Button>,
-    ]}
+      </Button>
   </Text>
 </Multichannel>
```
vanbasten17, line 65: this is not an example of the output, but what it generates. I would ask for an example where multichannel is used an adapt it in order to show a screenshot of how it looks in for example, Whatsapp.

vanbasten17 (changes requested):

docs/docs/components/components-multichannel.md
annehubtype: @vanbasten17 They are already listed in the Properties table, just the very same above the example. It's exactly the same content and see it as redundant.

docs/docs/components/components-multichannel.md
annehubtype: @vanbasten17 I've just added a visual example and modified the code.

vanbasten17 (approved):
Looking good now!

PR #1377, opened 2021-03-19
refactor(Preprocess): Engines use nlpjs package

elozano98:
## Description
Preprocess engines now use **nlpjs package**.

## Context
The BotonicPluginNer did not work with the natural package, so we decided to change it to the nlpjs one(also used in contentful). 

- Tokenizers and Stemmers now use the nlpjs package.
- Stopwords are now just an array containing all stopwords of the locale.
- Now, each type of engine has a dictionary where each locale is mapped with its corresponding engine.

## Testing

The pull request...

- [x] has unit tests
- [x] has integration tests


packages/botonic-nlp/src/preprocess/engines/en/tokenizer-en.ts
```diff
@@ -1,10 +1,8 @@
-import * as TreebankWordTokenizer from 'natural/lib/natural/tokenizers/treebank_word_tokenizer'
-
 import { Tokenizer } from '../../types'
 
-export default class TokenizerEn implements Tokenizer {
+export class TokenizerEn implements Tokenizer {
   readonly locale = 'en'
-  private tokenizer = new TreebankWordTokenizer()
+  private tokenizer = new (require('@nlpjs/lang-en-min/src/tokenizer-en'))()
```
dpinol, line 5: why do you do the require here?

dpinol (approved):

packages/botonic-nlp/src/preprocess/engines/en/tokenizer-en.ts
elozano98: It could be done before but I didn't know which name to use for it so I decided to just use it directly. 
Maybe I could have used something like `NlpjsTokenizerEn` (`Tokenizer` and `TokenizerEn` were already in use).

Do you think refactor is needed? @dpinol 

Maybe to something like:
`const NlpjsTokenizerEn = require('@nlpjs/lang-en-min/src/tokenizer-en')`
`private tokenizer = new NlpjsTokenizerEn()`

dpinol: ok thanks, leave. Let's discuss it when merging with botonic/contentful Tokenizers

packages/botonic-nlp/src/preprocess/preprocessor.ts
```diff
@@ -1,4 +1,8 @@
 import { Locale } from '../types'
+import { getNormalizer } from './engines/normalizer'
+// import { getStemmer } from './engines/stemmer'
```
vanbasten17, line 3: we should think of a way to activating stemmers depending on the use case (dev is using word embeddings or not). It's a discussion we need to have

vanbasten17 (approved):

packages/botonic-nlp/src/preprocess/preprocessor.ts
elozano98: I am agree, we need a way of choosing which engines are enabled. 
Maybe a class attribute for each engine, something like: `isTokenizerEnabled`, `areStopwordsEnabled`, etc.

dpinol: passing an interface with optional arguments is a flexible way to customize this.
btw, @marcrabat @elozano98  I'm not a fan of calling flags starting with is/are. Writing more is always worse than written less.
The convention to start function names with is/are is to make clear that the function is a predicate (it returns boolean). In this case, maybe enableTokenizer or tokenizeInput would be better. As usual, most important is consistent within the project

vanbasten17: I agree with Dani 👍

elozano98: Okay, I'll do it in another PR. Thank you both!

PR #1378, opened 2021-03-19
fix(processor): ner processor needs to pad sequence

elozano98:
**Depends on #1377.** Please, review it first. :warning: 

## Description
The `generateInput` method of `Processor` needs to pad the sequence so it matches the defined maxLength.

## Context
The sequence generated by the `generateInput` method of `Processor` was not padded. So it can result in a sequence which length exceeds `maxLength`. 

## Solution
Padding the sequence before it was returned.

## Testing

The pull request...

- [x] has unit tests
- [x] has integration tests

### DIFF
--- a/packages/botonic-nlp/src/tasks/ner/process/processor.ts
+++ b/packages/botonic-nlp/src/tasks/ner/process/processor.ts
@@ -25,7 +25,10 @@ export class Processor {
 
   // Generates the Input data and the unmasked sequence.
   generateInput(text: string): { sequence: string[]; input: InputData } {
-    const sequence = this.processText(text)
+    const sequence = this.preprocessor.pad(
+      this.processText(text),
+      PADDING_TOKEN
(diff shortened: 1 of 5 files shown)


packages/botonic-nlp/src/tasks/ner/process/processor.ts
```diff
@@ -25,7 +25,10 @@ export class Processor {
 
   // Generates the Input data and the unmasked sequence.
   generateInput(text: string): { sequence: string[]; input: InputData } {
-    const sequence = this.processText(text)
+    const sequence = this.preprocessor.pad(
```
dpinol, line 28: in tests/helpers/tasks/ner/helper.ts please rename MAX_LENGTH to st more descriptive . MAX_TOKEN_LENGTH?

vanbasten17 (approved):

PR #1379, opened 2021-03-19
feat(cli): task flag in botonic train command

elozano98:
## Description
Botonic now has two different NLP task:

1.  Text Classification
2. Named Entity Recognition

The `botonic train` command needs a way to let the user decide the desired task for training the model. Or whether train a model for each of the possible tasks.
 
## Context
This PR gives to the user the option of selecting the task(s) of the training process.

The `train` command now has a `task` flag where the user can define the desired task. If no flag is defined, a model for each of the tasks is trained. 

By now, the only task flag values allowed are:
 - `ner`
 - `text-classification`

Examples:
- `botonic train` : Trains a **Named Entity Recognition** model and also a **Text Classifier** model (all possible Botonic tasks).
- `botonic train --task=ner` : Only a **Named Entity Recognition** model is trained.

:warning:  **The Plugin of the selected task is required. If not installed, no model would be trained for this specific task.** :warning:

### DIFF
--- a/packages/botonic-cli/src/commands/train.ts
+++ b/packages/botonic-cli/src/commands/train.ts
@@ -29,22 +29,12 @@ class Task {
   }
 
   private logTaskPluginNotInstalled(): void {
-    console.error(
+    console.log(
       colors.red(
         `Training process has been stopped because you don't have @botonic/plugin-${this.name} installed.\nPlease, install it with the following command:`
       )
     )
     console.log(colors.bold(`$ npm install @botonic/plugin-${this.name}`))
-  }
-}
-
-class InvalidTaskError extends Error {
-  constructor(taskName: string, availableTasks: string[]) {
-    super(
-      `Unsupported task '${taskName}'. Available tasks: '${availableTasks.join(
-        "', '"
-      )}'.`
-    )
   }
 }
 
@@ -56,7 +46,11 @@ export class Tasks {
 
   static getByName(taskName: string): Task {
     if (!this.isValidTask(taskName)) {
-      throw new InvalidTaskError(taskName, Object.keys(this.tasks))
+      throw new Error(
+        `Unsupported task '${taskName}'. Available tasks: '${Object.keys(
+          this.tasks
+        ).join("', '")}'.`
+      )
     }
     return this.tasks[taskName]
   }
--- a/packages/botonic-cli/src/commands/train.ts
+++ b/packages/botonic-cli/src/commands/train.ts
@@ -6,16 +6,45 @@ import { join } from 'path'
 import { Telemetry } from '../analytics/telemetry'
 import { spawnNpmScript } from '../util/system'
 
-enum TASKS {
-  NER = 'ner',
-  TEXT_CLASSIFICATION = 'text-classification',
+class Task {
+  constructor(readonly name: string) {}
+
+  run(): void {
+    if (this.isTaskPluginInstalled()) {
+      track(`Trained with Botonic train:${this.name}`)
+      spawnNpmScript(`train:${this.name}`, () => `Finished training `)
+    } else {
+      this.logTaskPluginNotInstalled()
+    }
+  }
+
+  private isTaskPluginInstalled(): boolean {
+    const path = join(
+      process.cwd(),
+      'node_modules',
+      '@botonic',
+      `plugin-${this.name}`
+    )
+    return existsSync(path)
+  }
+
+  private logTaskPluginNotInstalled(): void {
+    console.error(
+      colors.red(
+        `Training process has been stopped because you don't have @botonic/plugin-${this.name} installed.\nPlease, install it with the following command:`
+      )
+    )
+    console.log(colors.bold(`$ npm install @botonic/plugin-${this.name}`))
+  }
 }
+
+const TASK_NAMES = ['ner', 'text-classification']
 
 export default class Run extends Command {
   static description = 'Train your bot with NLP'
 
   static examples = [
-    `$ botonic train [--task=<${Object.values(TASKS).join('|')}>]
+    `$ botonic train [--task=<${TASK_NAMES.join('|')}>]
     TRAINING MODEL...
     `,
   ]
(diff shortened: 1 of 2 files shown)


packages/botonic-cli/src/commands/train.ts
```diff
@@ -1,44 +1,104 @@
 import { Command, flags } from '@oclif/command'
 import colors from 'colors'
-import * as path from 'path'
+import { existsSync } from 'fs'
```
vanbasten17, line 3: use pathExists from `util/file-system` to not introduce more noise. If it's not present in your current code, maybe you will need to rebase with master

packages/botonic-cli/src/commands/train.ts
```diff
@@ -1,44 +1,104 @@
...
+  }
+
+  private isPackageInstalled(packageName: string): boolean {
+    const path = join(
       process.cwd(),
       'node_modules',
       '@botonic',
-      'nlu'
+      packageName,
+      'dist'
```
vanbasten17, line 90: remove `dist`

vanbasten17 (approved):
Looking good! Try to add some tests in order to get rid of codecov warnings. Maybe you will need to mock spawnNpmScript in order to obtain the needed outputs. Let us know if you need help. There are examples of how to mock specific functions in cli's tests.

packages/botonic-cli/src/commands/train.ts
```diff
@@ -1,44 +1,103 @@
...
-      await import(botonicNLUPath)
-    } catch (e) {
-      console.log(e)
-      console.log(
-        colors.red(
-          `You don't have @botonic/plugin-nlu installed.\nPlease, install it with the following command:`
-        )
+    const { flags } = this.parse(Run)
+    const task = flags.task
+    if (task) {
```
dpinol, line 31: minimize if/checks and remember SRP.
create Task classes Ner and classification and you'll see that everything is simpler.

```
const ALL_TASKS = {
NER : new NerTask()
classification: new ClassificationTask()
}

getTaskByName(taskName): Task {
 if (!ALL_TASKS.taskName) {
    move here the contents of logInvalidTask 
}
async run() {

  const tasks = flags.task ?  [getTaskByName(flags.task)] : ALL_TASKS.values()
 tasks.forEach(t-> t.run()) 
}
```

packages/botonic-cli/src/commands/train.ts
```diff
@@ -1,44 +1,103 @@
...
-      return
+    } else {
+      this.logPackageNotInstalled(packageName)
     }
-    track('Trained with Botonic train')
-    spawnNpmScript('train', () => 'Finished training')
+  }
+
+  private logInvalidTask(task: string): void {
+    console.log(
```
dpinol, line 81: why don't use console.error?

packages/botonic-cli/src/commands/train.ts
```diff
@@ -1,44 +1,103 @@
...
+      colors.red(
+        `Unsupported task '${task}'. Available tasks: '${Object.values(
+          TASKS
+        ).join("', '")}'.`
+      )
+    )
+  }
+
+  private isPackageInstalled(packageName: string): boolean {
+    const path = join(process.cwd(), 'node_modules', '@botonic', packageName)
```
dpinol, line 91: how do you know that cwd will be the top folder here?

dpinol (changes requested):

packages/botonic-cli/src/commands/train.ts
elozano98: You're right, the code assumes that the user runs the `botonic train` command from the bot root folder. 
I would need to look at the node_modules of the bot, but there is an easy way of doing this?

packages/botonic-cli/src/commands/train.ts
elozano98: Not sure if I did something similar to what you expected, I would thank you very much if you can take a look at it. :pray:

packages/botonic-cli/src/commands/train.ts
elozano98: You're right, changed!

packages/botonic-cli/src/commands/train.ts
dpinol: If you rebase to master, you'll see that now in https://github.com/hubtype/botonic/blob/master/packages/botonic-cli/src/constants.ts#L4 we have constant BOTONIC_PROJECT_PATH
please do it after all these PRs are merged

dpinol (changes requested):
Try to implement Tasks class

packages/botonic-cli/src/commands/train.ts
```diff
@@ -38,6 +38,16 @@ class Task {
   }
 }
 
+class InvalidTaskError extends Error {
```
elozano98, line 41: What do you think about creating a specific error for this? :thinking: @dpinol @vanbasten17

packages/botonic-cli/src/commands/train.ts
```diff
@@ -46,20 +56,14 @@ class Tasks {
 
   static getByName(taskName: string): Task {
     if (this.isInvalidTask(taskName)) {
-      throw new Error(this.getInvalidTaskMessage(taskName))
+      throw new InvalidTaskError(taskName, Object.keys(this.tasks))
     }
     return this.tasks[taskName]
   }
 
   private static isInvalidTask(taskName: string): boolean {
```
dpinol, line 64: change to isValidTask. try to avoid negative functions/constants/variables. Double negations (!isInvalid) make the code more difficult to understand

packages/botonic-cli/src/commands/train.ts
dpinol: Whatever you decide, be consistent. For instance, it doesn't look to consistent that logTaskPluginNotInstalled generates the error message in a function, but InvalidTaskError does it in an exception

dpinol (changes requested):

packages/botonic-cli/src/commands/train.ts
```diff
@@ -1,49 +1,100 @@
...
-      )
-      console.log(`$ npm install @botonic/plugin-nlu`)
-      return
+      console.error(e)
     }
-    spawnNpmScript('train', () => 'Finished training')
+  }
+
+  private getTasks(taskName: string | undefined): Task[] {
+    return taskName ? [Tasks.getByName(taskName)] : Object.values(Tasks.tasks)
```
dpinol, line 98: move getTasks within Tasks.
A hint that it should be there is that it only gets data from Tasks class :-)
Tasks.tasks might can it be made private

dpinol (changes requested):

packages/botonic-cli/src/commands/train.ts
```diff
@@ -29,7 +29,7 @@ class Task {
   }
 
   private logTaskPluginNotInstalled(): void {
-    console.error(
+    console.log(
```
dpinol, line 32: in a different PR, please create a log_error function in some utils file which does:
msg => console.log(colors.red(...)

and maybe a log_msg with 
msg => console.log(colors.bold(...)

dpinol (approved):

PR #1391, opened 2021-03-22
feat(utils): Environment and locale utils implemented

elozano98:
**Depends on #1392 .** Please, review it first. :warning: 

## Description
Environment and Locale utils for Botonic NER Plugin implemented.

## Context
The implemented util functions are:
- `detectLocale`: detects the locale of a given text.
- `getModelInfo`: returns the configuration and the model of the given locale.

## Testing

The pull request...

- [x] has unit tests


vanbasten17 (approved):

packages/botonic-plugin-ner/babel.config.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
```
dpinol, line 1: try removing it. I'm not sure it's necesary for TS, and we have already one at the packages folder

packages/botonic-plugin-ner/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "prepare": "node ../../preinstall.js",
+    "lint": "npm run lint_core -- --fix",
+    "lint_ci": "npm run lint_core -- -c ../.eslintrc_slow.js",
+    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.ts*'"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.5.5",
+    "@botonic/core": "~0.18.0",
+    "@types/webgl2": "0.0.5",
+    "axios": "latest",
```
dpinol, line 21: never use latest for non-dev dependencies. At most, use ^

packages/botonic-plugin-ner/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "node-fetch": "^2.6.1",
+    "ts-node": "^9.0.0",
```
dpinol, line 23: move @types/* to devdependencies as they are not necessary for runtime
remove typescript and ts-node because they're already in parent folder's package.json

packages/botonic-plugin-ner/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "typescript": "^4.0.3",
+    "@tensorflow/tfjs": "^2.7.0",
+    "franc": "^5.0.0",
+    "langs": "^2.0.0"
+  },
+  "devDependencies": {
+    "@babel/cli": "^7.5.5",
```
dpinol, line 30: try to remove all the babel dependencies. You don't need babel nor core-js with typescript

packages/botonic-plugin-ner/src/utils/environment-utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  ASSETS_DIR,
+  CONFIG_FILENAME,
+  MODEL_FILENAME,
+  MODELS_DIR,
+  NER_DIR,
+} from '../constants'
+import { ModelInfo } from '../types'
+
+export async function getModelInfo(locale: Locale): Promise<ModelInfo> {
+  const isProd = process.env.STATIC_URL !== undefined
```
dpinol, line 15: I guess isProd is not used to differenciate production & staging, but local/dev or deployed, right?
if so, I would have a temporary named st like isBrowser, or localModel

dpinol, line 15: Env vars are shared for all applications on a system. Use a name as distinct as possible to give it as much semantincs as possible and to avoid collisions (eg. use BOTONIC_NLP_MODEL_URL instead of STATIC_URL)

packages/botonic-plugin-ner/tsconfig.json
```diff
@@ -0,0 +1,34 @@
...
+  "include": ["src/**/*.ts", "src/**/*.tsx"],
+  "compilerOptions": {
+    "allowUnreachableCode": false,
+    "baseUrl": "src",
+    "paths": {
+      "*": ["src/*", "dist/src/*", "types/*"]
+    },
+    "moduleResolution": "node",
+    "rootDir": "src",
+    "outDir": "dist",
```
dpinol, line 11: we use the convention in hubtype of using lib for libraries and dist for final applications like bots botonic-nlu was not following it, but now it's time to do it here. please apply it in other nlp packages (except the legacy ones) in other PRs

dpinol (changes requested):

packages/botonic-plugin-ner/src/utils/environment-utils.ts
elozano98: We would have to modify backend to change this env variable name. @marcrabat

elozano98:
Checks are not successful because there is no @botonic/nlp package published yet.

dpinol (changes requested):

dpinol:
> succes

you can publish it as an alpha when you finish these PRs

dpinol (approved):

PR #1392, opened 2021-03-22
feat(ner): model stored by locale

elozano98:
**Depends on #1379.** Please, review it first. :warning: 

## Description
Models need to be stored by their locale.

## Context
Since we want Botonic NLP to allow having different models for different locales, storing them by their locale is needed.


vanbasten17 (approved):

dpinol (approved):
this PR is too small, but it might make sense on the context

PR #1393, opened 2021-03-22
feat(ner-plugin): Preprocess implemented

elozano98:
**Depends on #1391.** Please, review it first. :warning: 

## Description
An `InputGenerator` has been implemented to generate the input tensor needed by the model.

## Context
This `InputGenerator` is needed for creating the input tensor needed by the model and also the sequence that will be used to map the prediction of the model with each token.

It uses the `Preprocessor` and `Codifier` classes of the **botonic/nlp package** to generate the input tensor and the sequence.

## Testing

The pull request...

- [x] has unit tests


vanbasten17 (approved):

packages/botonic-plugin-ner/src/preprocess/input-generator.ts
```diff
@@ -0,0 +1,45 @@
+import { Codifier } from '@botonic/nlp/dist/preprocess/codifier'
+import {
+  PADDING_TOKEN,
+  UNKNOWN_TOKEN,
+} from '@botonic/nlp/dist/preprocess/constants'
+import { Preprocessor } from '@botonic/nlp/dist/preprocess/preprocessor'
+import { tensor } from '@tensorflow/tfjs-core/dist/ops/tensor'
+import { Tensor2D } from '@tensorflow/tfjs-core/dist/tensor'
+
+export class InputGenerator {
```
dpinol, line 10: this is SRP :100: 
However, for me a "generator" sounds like the initial source of data (normally used for test). Please correct me if in NLP people use this terminology.
Please review NLP literature and libraries, but what about this?:
* rename Codifier to Encoder (google finds 1M search results for "nlp encoder" and 33k for "nlp codifier"
* rename InputGenerator to NerEncoder
as usual appy the agreed refactors in a separate PR at the end of the PR chain

packages/botonic-plugin-ner/src/preprocess/input-generator.ts
```diff
@@ -0,0 +1,45 @@
...
+      this.preprocessor.stem(
+        this.preprocessor.removeStopwords(
+          this.preprocessor.tokenize(this.preprocessor.normalize(text))
+        )
+      ),
+      PADDING_TOKEN
+    )
+  }
+
+  private maskUnknownTokens(tokens: string[]): string[] {
```
dpinol, line 34: Within a class, order the functions from top/public to down/private. maskUnknownTokens should go after processTokens

packages/botonic-plugin-ner/src/preprocess/input-generator.ts
```diff
@@ -0,0 +1,45 @@
...
+} from '@botonic/nlp/dist/preprocess/constants'
+import { Preprocessor } from '@botonic/nlp/dist/preprocess/preprocessor'
+import { tensor } from '@tensorflow/tfjs-core/dist/ops/tensor'
+import { Tensor2D } from '@tensorflow/tfjs-core/dist/tensor'
+
+export class InputGenerator {
+  constructor(
+    readonly preprocessor: Preprocessor,
+    readonly tokenCodifier: Codifier,
+    readonly entityCodifier: Codifier
```
dpinol, line 14: remove entityCodifier if not used

dpinol (changes requested):

packages/botonic-plugin-ner/src/preprocess/input-generator.ts
elozano98: I named the class `Codifier` because it has two methods. One for encoding and the other for decoding. I thought it might be helpful to split this class into two: `Encoder` and `Decoder`. Maybe Even 4: `Encoder`, `CategoricalEncoder`, `Decoder`, `CategoricalDecoder`. What do you tink about this? @dpinol

dpinol (approved):

PR #1396, opened 2021-03-22
feat(ner-plugin): BotonicPluginNer implemented

elozano98:
**Depends on #1395.** Please, review it first. :warning: 

## Description
Implementeation of the BotonicPluginNER. 

## Context
This plugin runs the model assign to the detected locale and inserts the recognized entities.


elozano98:
I don't know if `ModelsRouter` is the best name for the implemented class. The main use of this class is to detect the language of the given text and use the according NER model to recognize entities.
Also, it loads the NER models in parallel.

packages/botonic-plugin-ner/src/model/ner.ts
```diff
@@ -2,38 +2,41 @@ import { Codifier } from '@botonic/nlp/dist/preprocess/codifier'
...
 export class NamedEntityRecognizer {
   model: LayersModel
   config: NerConfig
   inputGenerator: InputGenerator
   predictionProcessor: PredictionProcessor
 
-  constructor(readonly locale: Locale) {
+  constructor(modelInfoPromise: ModelInfoPromise) {
     // @ts-ignore
     return (async () => {
```
dpinol, line 25: @elozano98 @vanbasten17 
don't use async constructor. it's a hack (that's why you need to use ts-ignore). I haven't found a single web page recommending using it and that's why we don't use it in any TS project and why we avoided in botonic-nlp. Pages recommending not to use it and alternatives:
* https://rostacik.net/2017/04/23/how-to-handle-async-functions-inside-constructors-in-typescript/
* https://medium.com/@guillaume.viguierjust/dealing-with-asynchronous-constructors-in-typescript-c13c14c80954
* https://stackoverflow.com/a/47315507/145289

iife is also not recommended to use. See:
* https://github.com/thornjad/eslint-plugin-no-iife 
* https://levelup.gitconnected.com/why-its-time-to-stop-using-javascript-iifes-b62602f25bfc

dpinol (changes requested):

packages/botonic-plugin-ner/src/types.ts
```diff
@@ -1,10 +1,11 @@
 import { NerConfig } from '@botonic/nlp/dist/tasks/ner/storage/types'
 import { Locale } from '@botonic/nlp/dist/types'
 import { LayersModel } from '@tensorflow/tfjs'
+import { AxiosPromise } from 'axios'
 
-export type ModelInfo = {
-  config: NerConfig
-  model: LayersModel
+export type ModelInfoPromise = {
```
dpinol, line 6: since it's not actually a Promise, I think it makes sense to call it st like ModelInfoPromises to avoid that developers do await on the whole object.
I'm not sure if we should change the function that creates it to return instead of `Promise<{config:NerConfig, model: LayersModel}`. But we can do it in the future.

dpinol (changes requested):

vanbasten17 (approved):
Cool!! 👏 👏 👏

elozano98:
Tests failed because the `@botonic/nlp` package is not released.

packages/botonic-plugin-ner/src/model/model-info.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export class ModelInfo {
+  private config: AxiosPromise<NerConfig>
+  private model: Promise<LayersModel>
+
+  constructor(readonly locale: Locale, readonly uri: string) {
+    this.init()
+  }
+
+  private init() {
```
dpinol, line 16: what is the purpose of init? drop it 
actually it causes a problem, since in strict mode the compiler will complain that config & model are not initialized in the constructor

packages/botonic-plugin-ner/src/utils/environment-utils.ts
```diff
@@ -1,24 +1,33 @@
...
-  const isLocalModel = process.env.STATIC_URL === undefined
-  const domain = isLocalModel ? window.location.href : process.env.STATIC_URL
-  const uri = isLocalModel
-    ? `${domain}/${NER_DIR}/${MODELS_DIR}/${locale}`
-    : `${domain}/${ASSETS_DIR}/${MODELS_DIR}/${NER_DIR}/${locale}`
-  return {
-    config: (await axios({ url: `${uri}/${CONFIG_FILENAME}` })).data,
-    model: await loadLayersModel(`${uri}/${MODEL_FILENAME}`),
+export enum ENVIRONMENT {
+  PRODUCTION,
```
dpinol, line 6: maybe call it DEPLOYED?

dpinol (changes requested):

packages/botonic-plugin-ner/src/model/model-info.ts
```diff
@@ -10,10 +10,6 @@ export class ModelInfo {
   private model: Promise<LayersModel>
```
dpinol, line 10: please make these 2 fields private readonly

dpinol (approved):

PR #1397, opened 2021-03-22
Chore/publish abort

dpinol:
Two improvements in our script to publish all packages:
###     install common dependencies once during publish
    The script to publish all packages called "npm i -D" on each package, which causes to reinstall the common dependencies at the root folder multiple times.
    Now there's a flag to disable this feature, and the common dependencies are installed only once

###     publish to abort when st fails
Before, when "npm run build" failed, there was a generic error message, and the script continued. Now, the script logs the stderror of the command that failed and the process abort. I added NOP build scripts for JS packages to avoid that "npm run build" fails for them.

### DIFF
--- a/packages/botonic-plugin-dashbot/package.json
+++ b/packages/botonic-plugin-dashbot/package.json
@@ -4,6 +4,7 @@
   "license": "MIT",
   "main": "src/index.js",
   "scripts": {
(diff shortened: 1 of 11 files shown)


vanbasten17 (approved):

manuelfidalgo (approved):
:clap:

PR #1398, opened 2021-03-24
Fix/cli logout command

vanbasten17:
## Description
* Logout should unlink file inside credentials folder, not credentials folder.
* Bumping cli to 0.18.5


elozano98 (approved):

manuelfidalgo (approved):
:clap: thanks!

PR #1401, opened 2021-03-25
refactor(plugin-watson): revive watson plugin with assistantV2

vanbasten17:
## Description
Botonic `plugin-watson` stopped working and a refactor has been needed to revive it.

## Documentation
We need to document that this plugin **does not support V1 Assistant anymore**. **Instead, it supports V2 Assistant.**

Now, the plugin options needed are:
- `apikey`
- `url`
- `assistant_id`
- `version` (Optional: if not provided, '2020-04-01' will be used)

## Test
If you want to try the new plugin:
1. Ask @elozano98 for the Watson Assistant options used.
2. Create a bot and link this `plugin-watson` version to it.
3. Log the data assigned to `input` in the `pre` method.
4. Deploy your chatbot.
5. Text to your chatbot with something like: "I want to return this leather jacket".
6. Feel free to experiment with it!


elozano98:
@vanbasten17 I can not add you to reviewers, but can you take a look at this PR please? :pray:

vanbasten17 (commented):
LGTM, but we must be sure to return the results in the Botonic standard format.  Remove `@types/axios` devDep as the plugin is not in TS yet. Also, too many things happening in `index.js`. Encapsulate it into a class `WatsonOutputParser`. And call sth like:
```
const parser = new WatsonOutputParser()
const nluResults = await parser.toBotonicOutput(input.data)
Object.assing(input, nluResults)
```

packages/botonic-plugin-watson/src/index.js
```diff
@@ -1,36 +1,106 @@
...
-        workspaceId: this.options.workspaceId,
-      })
-      intent = res.result.intents[0].intent
-      confidence = res.result.intents[0].intent.confidence
-      intents = res.result.intents
-      entities = res.result.entities
-    } catch (error) {
-      console.log(error)
+      const output = await this.getWatsonOutput(input.data)
+      Object.assign(input, WatsonOutputParser.parse(output))
```
vanbasten17, line 83: much better now! I would call it `parseToBotonic` just to be explicit that we are converting it to the Botonic format, o.w, go for it

vanbasten17: also, we should modify the README with the new keys to be given.

elozano98: I agree, lets change it! :+1:

packages/botonic-plugin-watson/src/index.js
```diff
@@ -1,36 +1,99 @@
...
-        input: { text: input.data },
-        workspaceId: this.options.workspaceId,
-      })
-      intent = res.result.intents[0].intent
-      confidence = res.result.intents[0].intent.confidence
-      intents = res.result.intents
-      entities = res.result.entities
-    } catch (error) {
-      console.log(error)
+      const output = await this.getWatsonOutput(input.data)
```
dpinol, line 75: if initialization fails, here you will get strange errors.
I suggest that here you check eg sessionId and if it's undefined, you log st like "Not executing Watson because initialization failed"

packages/botonic-plugin-watson/src/index.js
```diff
@@ -1,36 +1,99 @@
...
+  EXPECTED_SESSION_ID_REQUEST_STATUS = 201
+  EXPECTED_OUTPUT_REQUEST_STATUS = 200
+
   constructor(options) {
-    this.options = options
+    try {
+      this.validateOptions(options)
+      this.init(options)
+    } catch (e) {
+      console.error(String(e))
```
dpinol, line 16: if you cast e to String, you loose the callstack.

packages/botonic-plugin-watson/src/index.js
```diff
@@ -1,36 +1,99 @@
...
+    }
+  }
+
+  validateOptions(options) {
+    if (!options) throw new Error('No options provided.')
+    if (!this.areRequiredOptionsProvided(options))
+      throw new Error('Missing one or more fields.')
```
dpinol, line 23: nice that you check the options are provided.
But think on the user (developer). If you don't provide which one is missing, it will be difficult to find which one is wrong (maybe there's a misspell, maybe a bug lead to a null value...)
I suggest this approach
* here you call "Object.keys(o => checkRequiredOption(o))"
```
function  checkRequiredOption(key) {
  if key does not exist {
   throw "missing option ${key}
  }
  if option[key] == undefined {
   throw "undefined value for option ${key}"
  }
}
```

packages/botonic-plugin-watson/src/index.js
```diff
@@ -1,36 +1,99 @@
...
+    const requiredOptionsFieldsProvided = Object.keys(options).filter(field =>
+      this.REQUIRED_OPTION_FIELDS.includes(field)
+    )
+    return (
+      requiredOptionsFieldsProvided.length == this.REQUIRED_OPTION_FIELDS.length
+    )
+  }
+
+  areRequiredOptionsDefined(options) {
+    const requiredOptionsProvided = Object.entries(
```
dpinol, line 38: read clean code book chapter on names .
"the length of a name should be correspond to the size of its scope." https://dzone.com/articles/naming-conventions-from-uncle-bobs-clean-code-phil

dpinol (changes requested):

packages/botonic-plugin-watson/src/watson-output-parser.js
```diff
@@ -0,0 +1,30 @@
...
+    output.intents.sort((i1, i2) => (i1.confidence > i2.confidence ? -1 : 1))
+    return {
+      intent: this.getIntentWithMaxConfidence(output.intents),
+      confidence: this.getMaxConfidence(output.intents),
+      intents: output.intents,
+      entities: this.parseEntities(output.entities),
+    }
+  }
+
+  static getIntentWithMaxConfidence(intents) {
```
MarcosCA, line 15: the intent with maxconfidence is the first one?

elozano98: Yes, because I first sorted them

packages/botonic-plugin-watson/package.json
```diff
@@ -11,9 +11,12 @@
     "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.js'"
   },
   "dependencies": {
-    "ibm-watson": "^6.0.0"
+    "@babel/runtime": "^7.13.10",
+    "axios": "^0.21.1"
+  },
+  "devDependencies": {
+    "@types/axios": "^0.14.0"
```
asastre, line 18: As @vanbasten17 said, this plugin is not using typescript so you can remove this package

asastre (approved):
Just 2 small observations (the documentation one I think it should be fixed), the rest LGTM!

MarcosCA (approved):

PR #1418, opened 2021-03-31
feat(nlp): Split different types of Codifiers

elozano98:
## Description
Two different codifiers are implemented:
- `Codifier`.
- `OneHotCodifier`

## Context
The old version of the `Codifier` has the `isCategorial` flag implemented to decide if the `encoding`/`decoding` would generate a _one hot vector_ or not.

Since this flag conditions the behaviour of the class methods, a refactor to split this `Codifier`  into two different classes is needed.

## Testing

The pull request...
- [x] has unit tests


packages/botonic-nlp/src/codify/codifier.ts
```diff
@@ -0,0 +1,39 @@
+import { unique } from '../utils/array-utils'
+
+export class Codifier {
```
dpinol, line 3: split into Encoder & Decoder classes. Rationale:
* I've been skimming other liibraries and it looks like they always go independently.
* encoder & decoder share very little state
* it will allow extending each Encoder & Decoder independently

You could create a shared class Vocabulary (but don't inherit from it, compose it) where you call unique and held the isValidToken and isValidTokenId methods

packages/botonic-nlp/src/codify/codifier.ts
```diff
@@ -0,0 +1,39 @@
...
+    this.vocabulary = unique(vocabulary)
+  }
+
+  encode(sequence: string[]): number[] {
+    return sequence.map(token => this.encodeToken(token))
+  }
+
+  private encodeToken(token: string): number {
+    if (!this.isValidToken(token)) {
+      throw new Error(`Invalid Token '${token}'.`)
```
dpinol, line 16: I think a more useful message is "Token ${token} not found in vocabulary"

dpinol (changes requested):

packages/botonic-nlp/src/codify/codifier.ts
dpinol: I suggest using the vocabulary from https://www.analyticsvidhya.com/blog/2020/03/one-hot-encoding-vs-label-encoding-using-scikit-learn/ https://towardsdatascience.com/categorical-encoding-using-label-encoding-and-one-hot-encoder-911ef77fb5bd (label encoder vs one-hot encoder), which is what sklearn uses

packages/botonic-nlp/src/codify/one-hot-codifier.ts
```diff
@@ -0,0 +1,51 @@
...
+  private decodeCategorical(categorical: number[]): string {
+    const id = this.categoricalToTokenId(categorical)
+    return this.vocabulary[id]
+  }
+
+  private categoricalToTokenId(categorical: number[]): number {
+    if (!this.isValidCategorical(categorical)) {
+      throw new Error(`Invalid categorical length '${categorical.length}'.`)
+    }
+    return categorical.indexOf(Math.max(...categorical))
```
dpinol, line 45: would it be the same to do categorial.indexOf(1) ?

packages/botonic-nlp/src/codify/codifier.ts
```diff
@@ -0,0 +1,39 @@
+import { unique } from '../utils/array-utils'
```
dpinol, line 1: I suggest renaming the codify directory to encoding or codecs

packages/botonic-nlp/src/codify/codifier.ts
```diff
@@ -0,0 +1,39 @@
...
+  constructor(vocabulary: string[]) {
+    this.vocabulary = unique(vocabulary)
+  }
+
+  encode(sequence: string[]): number[] {
+    return sequence.map(token => this.encodeToken(token))
+  }
+
+  private encodeToken(token: string): number {
+    if (!this.isValidToken(token)) {
```
dpinol, line 15: this has 2 issues:
1) performance: you're finding the token in the vocabulary twice
2) simplicity: isValidToken function would make sense if you needed it as public function or if you called it from more than 1 place. Otherwise, you're just introducing complexity by separating the check from the exception

packages/botonic-nlp/src/codify/codifier.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  private decodeTokenId(id: number): string {
+    if (!this.isValidTokenId(id)) {
+      throw new Error(`Invalid Token id '${id}'.`)
+    }
+    return this.vocabulary[id]
+  }
+
+  private isValidTokenId(id: number) {
+    return this.vocabulary.length > id && id >= 0
```
dpinol, line 37: readibility: in a comparison, you should try to put the checked variable first  
id >=0 && id <=this.voca....
as this is how people would speak: you wouldn't say that you're checking that the vocabulary length is larger than the specified number, but that "you're checking that a number is smaller than the voc length"
another option would be 
 0<=id && id <=this.voca....
since it's simmetrical and it's easier to read left to right.

dpinol (changes requested):

packages/botonic-nlp/tests/helpers/tasks/ner/helper.ts
```diff
@@ -1,12 +1,13 @@
 import { join } from 'path'
 
+import { Codifier } from '../../../../src/codify/codifier'
```
dpinol, line 3: please rebase so that we don't get the spam from codecov

dpinol (changes requested):

packages/botonic-nlp/src/codify/one-hot-codifier.ts
elozano98: No, because the maximum number of a One Hot Vector is not always 1.
Normally the One Hot Vector you want to decode is the output prediction of a model:
`[0.81, 0.04, 0.1, 0.05] `

That's why the max function is needed.

packages/botonic-nlp/src/codify/codifier.ts
elozano98: LGTM, I will implement the Vocabulary class in another PR. 👍🏽

dpinol (approved):

vanbasten17 (approved):

PR #1435, opened 2021-04-07
feat(nlp): Text Classification - Storage

elozano98:
## Description
Implementation of the class that `loads` and `stores` the configuration of a trained model.

## Context
The configuration needs to be:
- **Loaded** to configure the model that will be used.
- **Stored** so the trained model can be used later.  

## Testing

The pull request...

- [x] has unit tests

### DIFF
--- a/packages/botonic-nlp/src/tasks/text-classification/storage/config-storage.ts
+++ b/packages/botonic-nlp/src/tasks/text-classification/storage/config-storage.ts
@@ -1,0 +1,17 @@
+import { mkdirSync, writeFileSync } from 'fs'
+import { join } from 'path'
+
+import { CONFIG_FILENAME } from '../../../storage/constants'
+import { readJSON } from '../../../utils/file-utils'
+import { TextClassificationConfig } from './types'
+
+export class TextClassificationConfigStorage {
+  static load(path: string): TextClassificationConfig {
+    return readJSON(join(path, CONFIG_FILENAME)) as TextClassificationConfig
+  }
+
+  static save(path: string, config: TextClassificationConfig): void {
+    mkdirSync(path, { recursive: true })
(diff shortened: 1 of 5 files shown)


packages/botonic-nlp/src/tasks/text-classification/storage/config-storage.ts
```diff
@@ -0,0 +1,17 @@
+import { mkdirSync, writeFileSync } from 'fs'
+import { join } from 'path'
+
+import { CONFIG_FILENAME } from '../../../storage/constants'
```
dpinol, line 4: personally I prefer to put the constants as close as possible to where they're used (so here it would be within TextClassificationConfigStorage (and in private const), since no other class will ever need to know about it)

dpinol (changes requested):

vanbasten17 (approved):

packages/botonic-nlp/src/tasks/text-classification/storage/config-storage.ts
```diff
@@ -0,0 +1,20 @@
+import { mkdirSync, writeFileSync } from 'fs'
+import { join } from 'path'
+
+import { readJSON } from '../../../utils/file-utils'
+import { TextClassificationConfig } from './types'
+
+export class TextClassificationConfigStorage {
+  public readonly CONFIG_FILENAME = 'config.json'
+
+  load(path: string): TextClassificationConfig {
```
dpinol, line 10: when we see how the class is used, we'll decide if we want to move the path to the constructor

dpinol (approved):

packages/botonic-nlp/src/tasks/text-classification/storage/config-storage.ts
elozano98: makes sense! 👍🏽

PR #1437, opened 2021-04-07
refactor(plugin-ner): Codifier replaced by LabelEncoder

elozano98:
## Description
Due to the Codifier refactor previously done, all Codifier instances should be replaced by their corresponding encoder in `plugin-ner`.


elozano98:
Tests fail because the alpha version of `@botonic/nlp` does not include the Codifier refactor.

dpinol (approved):

vanbasten17 (approved):

PR #1441, opened 2021-04-09
feat(nlp): Text Classification - Processor

elozano98:
## Description
Implementation of the class needed to process data and generate the input and output tensors.

## Context
The Text Classifier model works with tensors.
We need to process the data to:
- Create Input tensors for training the model.
- Create the expected output tensors to evaluate the trained models. 

## Testing

The pull request...

- [x] has unit tests


packages/botonic-nlp/src/tasks/text-classification/process/processor.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  private generateSampleInput(text: string): number[] {
+    const sequence = this.generateSequence(text)
+    return this.tokensEncoder.encode(sequence)
+  }
+
+  private generateSequence(text: string): string[] {
+    const normalizedText = this.preprocessor.normalize(text)
```
dpinol, line 38: DRY: Duplicated in VocabularyGenerator.generate i a TextProcessor.generateSequence. Create Preprocessor.full(string): string[]?

packages/botonic-nlp/tests/tasks/text-classification/botonic-text-classifier.test.ts
```diff
@@ -30,6 +30,29 @@ describe('Botonic Text Classifier', () => {
...
   })
 
+  test('Train and Evaluate Model', async () => {
+    const dataset = classifier.loadDataset(generalHelper.DATA_DIR_PATH)
+    const { trainSet, testSet } = classifier.splitDataset(dataset, 0.5, false)
+    classifier.generateVocabulary(trainSet)
+    try {
+      classifier.compile()
+    } catch (e) {
+      console.log(e)
```
dpinol, line 40: why? remove catch if possible

packages/botonic-nlp/tests/tasks/text-classification/botonic-text-classifier.test.ts
```diff
@@ -30,6 +30,29 @@ describe('Botonic Text Classifier', () => {
...
+      TEXT_CLASSIFIER_TEMPLATE.SIMPLE_NN,
+      await DatabaseStorage.with(
+        helper.LOCALE,
+        helper.EMBEDDINGS_TYPE,
+        helper.EMBEDDINGS_DIMENSION
+      )
+    )
+    await classifier.train(trainSet, 4, 8)
+    const sut = await classifier.evaluate(testSet)
+    expect(sut.accuracy).toBeDefined()
```
dpinol, line 52: try to put some check even result is not deterministic. eg expect(..).toBeGreaterThan(...)

packages/botonic-nlp/tests/tasks/text-classification/process/processor.test.ts
```diff
@@ -0,0 +1,39 @@
...
+import * as helper from '../../../helpers/tasks/text-classification/tools-helper'
+
+describe('Text Classification Processor', () => {
+  const processor = new Processor(
+    helper.preprocessor,
+    helper.tokensEncoder,
+    helper.classEncoder
+  )
+  test('Sample Processing', () => {
+    const sut = processor.process([
```
dpinol, line 11: sut is Subject Under Test (eg. the class instance you're testing). In this case the processor.

packages/botonic-nlp/src/tasks/text-classification/process/processor.ts
```diff
@@ -0,0 +1,61 @@
...
+      this.tokensEncoder.vocabulary.includes(token) ? token : UNKNOWN_TOKEN
+    )
+  }
+
+  //TODO: modify encoders to just encode one token (not sequences).
+  private generateSampleOutput(className: string): number[] {
+    return this.classEncoder.encode([className])[0]
+  }
+
+  generateInput(text: string): InputData {
```
dpinol, line 58: public methods should be at the top of the class.

I don't like the word "generate" ( unless it's used in NLP community). 
be consistent with names. If the other public method is called "process", I suggest renaminf it to processSamples, and this one to processInput.

dpinol (changes requested):

packages/botonic-nlp/tests/tasks/text-classification/botonic-text-classifier.test.ts
elozano98: The catch is a way to avoid that the compile method throws an Error because the PredictionProcessor is not implemented. Maybe it is better to remove this test until the PredictionProcessor is implemented and no errors are thrown.

packages/botonic-nlp/src/tasks/text-classification/process/processor.ts
elozano98: You are right, maybe something like `preprocessor.preprocess()` would be nice to have. I've opened a [ticket](https://linear.app/hubtype/issue/BOT-327/preprocessor-generic-method) for this.

packages/botonic-nlp/src/tasks/text-classification/process/processor.ts
elozano98: The reason why this public method is not at the top of the class is that I think that methods should be defined in the same order that are used to make the code easy to understand. That's why there are some private methods between both public methods. Is that a wrong approach? @dpinol

packages/botonic-nlp/src/tasks/text-classification/process/processor.ts
```diff
@@ -0,0 +1,53 @@
...
+import { InputData, OutputData } from './types'
+
+export class Processor {
+  constructor(
+    readonly preprocessor: Preprocessor,
+    readonly tokensEncoder: LabelEncoder,
+    readonly classEncoder: OneHotEncoder
+  ) {}
+
+  process(samples: Sample[]): { x: InputData; y: OutputData } {
```
dpinol, line 17: rename process to processSamples and generateInput to processTexts (to be consistent with processText). Consistency above all! :-)

dpinol (changes requested):

packages/botonic-nlp/src/tasks/text-classification/process/processor.ts
vanbasten17: agree :)

vanbasten17 (approved):

dpinol (approved):

PR #1453, opened 2021-04-12
refactor(nlp): dataset processes refactored

elozano98:
## Description
Implementation and integration of a simpler way of loading and using a Dataset.

## Context
We want to simplify main NLP classes like `BotonicNer` or `BotonicTextClassifier`. In order to achieve this, Dataset processes like `load` or `split` should be removed from them.  

- A new Dataset class includes its own split method that generates two Datasets (train and test).

## Testing

The pull request...

- [x] has unit tests
- [x] has integration tests


packages/botonic-nlp/src/utils/array-utils.ts
```diff
@@ -1,3 +1,7 @@
 export function unique<T>(array: T[]): T[] {
   return Array.from(new Set(array))
 }
+
+export function randomSort<T>(array: T[]): T[] {
```
dpinol, line 5: rename to shuffle
apply this algorithm https://stackoverflow.com/a/12646864/145289
here explains why the solution in the PR is not optimal https://stackoverflow.com/a/18650169/145289

dpinol (changes requested):

packages/botonic-nlp/src/dataset/types.ts
```diff
@@ -6,7 +6,7 @@ export type FoundEntity = {
   end: number
 }
 export type Sample = { text: string; entities: FoundEntity[]; class: string }
-export type Dataset = {
+export type DatasetInfo = {
```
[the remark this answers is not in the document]
elozano98: Totally agree

dpinol (approved):

packages/botonic-nlp/tests/dataset/dataset.test.ts
```diff
@@ -0,0 +1,30 @@
...
+      { text: 'I want to buy this', class: 'BuyProduct', entities: [] },
+      { text: 'I want to return this', class: 'ReturnProduct', entities: [] },
+      {
+        text: 'I would like to return it',
+        class: 'ReturnProduct',
+        entities: [],
+      },
+      { text: 'I would like to buy it', class: 'BuyProduct', entities: [] },
+    ]
+  )
```
asastre, line 17: Just a comment of how I like to make/see tests (do we have any "best practices" documentation about how to structure tests @dpinol?)

I think it's a good practice to have all the tests at the beginning of the test file and at the bottom all the datasets or initialization functions. This way you can see right away what are being tested and if all the details will be at the bottom.

In this case, you could create a `getDataset()` function to return this `Dataset` and the describe block would be much cleaner. What do you think @elozano98?

asastre: Also, why not using the `dataset` of `tools-helper`? You're already loading a dataset there. Is it too big?

asastre (approved):
Nice refactor :clap:!

packages/botonic-nlp/tests/dataset/dataset.test.ts
dpinol: @asastre see https://www.notion.so/Unit-tests-8515f1d36352406ea461d06c08b2ee36#5b3d48716fac4ef6b76912f4e135a9e4

> If you need to repeat a given complex function call or data structure many times, create a helper function or constant as close and customized as possible to the test. 

I think the test follows the best practice which makes it readable because the data is close to the test

asastre: Thanks for the link and the explanation @dpinol !

vanbasten17 (approved):

PR #1455, opened 2021-04-13
style(core): remove RouteRequest type replace it with BotRequest

asastre:
## Description
Remove `RouteRequest` type in favour of using `BotRequest` instead in `Routes` type.

## Context
As you can see [here](https://github.com/hubtype/botonic/blob/master/packages/botonic-core/src/core-bot.js#L62), `CoreBot` is passing all 3 attributes declared in `BotRequest` type (`input`, `session` and `lastRoutePath`) in the `routes` function so having `RouteRequest` as the argument's type for the `routes` function was no correct.

## Testing

### DIFF
--- a/packages/botonic-core/src/types/index.d.ts
+++ b/packages/botonic-core/src/types/index.d.ts
@@ -170,8 +170,7 @@ export interface Route {
   type?: StringMatcher
 }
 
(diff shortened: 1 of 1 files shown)


dpinol (approved):
:clap:

AinaVendrell (approved):

PR #1459, opened 2021-04-14
feat(nlp): Vocabulary class

elozano98:
## Description
Implementation of the `Vocabulary` class and the way the vocabulary is extracted from a `Dataset`.

## Context
We need an easy way of creating a vocabulary and using it.

The `Dataset` class can extract a `Vocabulary` by preprocessing its samples.

Also, the `Vocabulary` has some methods that will be used later by the Encoders.

How to create a `Vocabulary`:
```typescript
const vocabulary = new Vocabulary([ 'product', 'material', 'color' ])
``` 

Extract the `Vocabulary` from a `Dataset`:
```typescript
const vocabulary = dataset.extractVocabulary(preprocessor)
``` 

## Testing

The pull request...

- [x] has unit tests
- [x] has integration tests

## Next steps
- Integrate `Vocabulary` class into `BotonicNER` and `BotonicTextClassifier`.
- Refactor `LabelEncoder` and `OneHotEncoder` to use `Vocabulary` class.

### DIFF
--- a/packages/botonic-nlp/src/preprocess/vocabulary.ts
+++ b/packages/botonic-nlp/src/preprocess/vocabulary.ts
@@ -14,10 +14,11 @@ export class Vocabulary {
   }
 
   getTokenId(token: string): number {
-    if (!this.includes(token)) {
+    const id = this.tokens.indexOf(token)
+    if (id === -1) {
       throw new Error(`Token '${token}' not found in vocabulary.`)
     }
-    return this.tokens.indexOf(token)
+    return id
   }
 
   getToken(id: number): string {
--- a/packages/botonic-nlp/src/preprocess/vocabulary.ts
+++ b/packages/botonic-nlp/src/preprocess/vocabulary.ts
@@ -1,0 +1,46 @@
+import { Dataset } from '../dataset/dataset'
+import { flatten, unique } from '../utils/array-utils'
+import { PADDING_TOKEN } from './constants'
+import { Preprocessor } from './preprocessor'
+
+export class Vocabulary {
+  readonly tokens: string[]
+  readonly length: number
+
+  constructor(tokens: string[]) {
+    this.tokens = unique(tokens)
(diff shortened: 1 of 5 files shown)


packages/botonic-nlp/src/preprocess/vocabulary.ts
```diff
@@ -0,0 +1,46 @@
...
+export class Vocabulary {
+  readonly tokens: string[]
+  readonly length: number
+
+  constructor(tokens: string[]) {
+    this.tokens = unique(tokens)
+    this.length = this.tokens.length
+  }
+
+  static fit(dataset: Dataset, preprocessor: Preprocessor): Vocabulary {
```
dpinol, line 15: pardon my nlp ignoreance, but "fit" sounds like a ML concept. is it?
A typical factory method name is fromXXX.What about fromDataset?
however, what about moving the method to Dataset.extractVocabulary?
In case of doubt in wondering where to place a function:
* Each class should only have 1 responsibility. To which one does the function semantics fit the best?
* put the function in the class which accesses data from a class instance
* Avoid low level classes to depend on high level classes (to avoid circular dependencies)

packages/botonic-nlp/src/preprocess/vocabulary.ts
```diff
@@ -0,0 +1,46 @@
...
+    const filteredTokens = uniqueTokens.filter(token => token !== PADDING_TOKEN)
+    return new Vocabulary(filteredTokens)
+  }
+
+  includes(token: string): boolean {
+    return this.tokens.includes(token)
+  }
+
+  getTokenId(token: string): number {
+    if (!this.includes(token)) {
```
dpinol, line 30: you're performing the search twice

dpinol (changes requested):

dpinol (approved):

MarcosCA (approved):

PR #1464, opened 2021-04-19
refactor(nlp): IndexedItems Integration

elozano98:
**Depends on #1459.** Please, review it first. ⚠️ 

## Description
- `Vocabulary` class renamed to `IndexedItems`.
- Dependency injection of `IndexedItems` to both Encoders.
- Refactor `LabelEncoder` and `OneHotEncoder` to use `IndexedItems` class.
- Refactor of `BotonicNer`, `BotonicTextClassifier`.
- VocabularyGenerator removed.

## Context
- Now, we inject the dependency of `IndexedItems` to Encoders to make it easier.
- `BotonicNer` and `BotonicTextClassifier` simplified.

## Testing

The pull request...
- [x] has integration tests


packages/botonic-nlp/src/tasks/ner/process/processor.ts
```diff
@@ -81,20 +81,20 @@ export class Processor {
     )
   }
 
-  private maskUnknownTokens(tokens: string[]): string[] {
-    return tokens.map(t =>
-      this.tokenCodifier.vocabulary.includes(t) ? t : UNKNOWN_TOKEN
+  private processTokens(sequence: string[]): number[] {
```
dpinol, line 84: 1) maskUnknownTokens is implemented 3 times (ner, text classification & botonic-plugin-ner). I suggest moving it to to a new IndexedItems.maskUnknownItems
2) Nested calls are difficult to read. I suggest to follow the "pipeline" model that you have done in other functions:
```
const unknownsMasked = this.tokenEncoder.items.maskUnknownItems(tokens, UNKNOWN_TOKEN)
const padded = this.preprocessor.pad(unknownsMasked, PADDING_TOKEN)
return this.tokenEncoder(padded)
```

packages/botonic-nlp/tests/tasks/text-classification/botonic-text-classifier.test.ts
```diff
@@ -2,31 +2,22 @@ import { existsSync, rmdirSync } from 'fs'
...
-    sut.generateVocabulary(trainSet)
-    expect(sut.vocabulary.length).toBeGreaterThan(2)
-    expect(sut.vocabulary.includes(PADDING_TOKEN)).toBeTruthy()
-    expect(sut.vocabulary.includes(UNKNOWN_TOKEN)).toBeTruthy()
-  })
-
   test('Train and Evaluate model', async () => {
     const { trainSet, testSet } = toolsHelper.dataset.split()
-    sut.generateVocabulary(trainSet)
+    const vocabulary = trainSet.extractVocabulary(toolsHelper.preprocessor)
```
dpinol, line 13: https://github.com/hubtype/botonic/pull/1464#commitcomment-49707848

dpinol (changes requested):

packages/botonic-nlp/src/tasks/ner/process/processor.ts
elozano98: I agree nested calls are difficult to read and that this part should be refactored, but I don't think `IndexedItems` is the best place where to move the `maskedUnknownTokens` method. 🤔

elozano98: I've implemented the `maskUnknownItems` method into `IndexedItems` on this PR, but the refactor of the three `Processors` will be done in a new PR. @dpinol

dpinol: I don't see maskUnknownItems method into IndexedItems

packages/botonic-nlp/src/encode/indexed-items.ts
```diff
@@ -26,6 +27,12 @@ export class IndexedItems {
     return this.items[idx]
   }
 
+  maskUnknownItems(sequence: string[], unknownToken = UNKNOWN_TOKEN): string[] {
```
dpinol, line 30: I prefer not to define any default value for unknownToken to avoid dependencies from encode to preprocess folder (to try to avoid circular dependencies)
rename unknownToken to unknownItem

dpinol (changes requested):

packages/botonic-nlp/src/encode/indexed-items.ts
elozano98: makes sense! 👍🏽

packages/botonic-nlp/src/encode/indexed-items.ts
```diff
@@ -1,3 +1,4 @@
+import { UNKNOWN_TOKEN } from '../preprocess/constants'
```
dpinol, line 1: remove import

dpinol (changes requested):

dpinol (approved):

vanbasten17 (approved):
Looking good! 👍

PR #1466, opened 2021-04-20
refactor(nlp): Remove Getters & Setters #BOT-330

elozano98:
## Description
Remove getters and setters of preprocessing engines.

## Context
No need for getters and setters for preprocessing engines needed. If an engine modification is required, then it should be implemented directly into the `Preprocess` class.


dpinol (approved):

marcpdw (approved):
Love it! 🤪  🔪

PR #1467, opened 2021-04-20
feat(nlp): importModel method #BOT-336

elozano98:
## Description
The way the model is created has been split into two steps:
- `createModel()` method refactored to return the created model.
- `setModel()` method needs to be called to set the model we want to use for training, evaluation, and predictions.

## Context
In this way, a user can decide to directly import a custom model. Also, **in the future**, if Botonic's community develops models for NLP tasks, this would be easily imported this way.  

If a user wants to use a template model from Botonic, the code would look something like:
```typescript

const embeddingsStorage = await DatabaseStorage.with(
  constantsHelper.LOCALE,
  constantsHelper.EMBEDDINGS_TYPE,
  constantsHelper.EMBEDDINGS_DIMENSION
)

const model = await classifier.createModel(
  TEXT_CLASSIFIER_TEMPLATE.SIMPLE_NN,
  embeddingsStorage
)
classifier.setModel(model)
```

## Testing

The pull request...

- [x] has unit tests
- [x] has integration tests

### DIFF
--- a/packages/botonic-nlp/src/tasks/ner/botonic-ner.ts
+++ b/packages/botonic-nlp/src/tasks/ner/botonic-ner.ts
@@ -108,7 +108,7 @@ export class BotonicNer {
     }
   }
 
-  importModel(model: LayersModel): void {
+  setModel(model: LayersModel): void {
     this.modelManager = new ModelManager(model)
   }
 
--- a/packages/botonic-nlp/src/tasks/ner/botonic-ner.ts
+++ b/packages/botonic-nlp/src/tasks/ner/botonic-ner.ts
@@ -1,4 +1,4 @@
-import { Tensor3D } from '@tensorflow/tfjs-node'
+import { LayersModel, Tensor3D } from '@tensorflow/tfjs-node'
 import { join } from 'path'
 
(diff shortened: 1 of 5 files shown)


dpinol (approved):

ericmarcos (approved):
Nice! Totally optional: I'd prefer renaming to `setModel` as the word "import" has some connotations (at least in my head) that imply a slow/dangerous/magic process. For example dynamically loading code from external sources at runtime, which is not the case.

elozano98:
I understand what you have said in the previous [comment](https://github.com/hubtype/botonic/pull/1467#pullrequestreview-641112488), but if we finally decide to change the method name, I think that it would be better to implement a setter instead:

```typescript
class BotonicNer {
    ....
    set model(model: LayersModel): void {
        this.modelManager = new ModelManager(model)
    }
}
``` 

Usage example:
```typescript
const ner = new BotonicNer(...)
....
ner.model = customModel
```

What do you think? Should we make any changes? @dpinol @ericmarcos

ericmarcos:
@elozano98 LGTM!

PR #1456, opened 2021-04-13
refactor(nlp): Preprocessor

elozano98:
## Description
The `Preprocessor` class needs the following modifications:
- Create a new `preprocess` method that preprocesses a given text.
- Remove `paddingPosition` from class constructor.
- Split `pad` method into: `pad` and `truncate` methods.

## Context
These changes are required to improve the code.

## Testing
The pull request...
- [x] has unit tests
- [x] has integration tests

PR #1468, opened 2021-04-20
fix(nlp, plugin-ner): Processors fixed #BOT-340

elozano98:
## Description
In a previous PR (#1456), the `pad` method from Preprocessor was split into two methods:
- `pad`
- `truncate`

By applying this change, we have to also add the `truncate` method wherever the `pad` method was used. But it was not done in `Processor` classes.

Also, we can use the `preprocess` method implemented in this previous PR to simplify the code.

## Context
A bug was introduced when the `pad` method was split in a previous PR.

## Testing

The pull request...

- [x] has unit tests
- [x] has integration tests


elozano98:
Checks fail because we need to deploy a new alpha for `nlp` package with the new `Preprocess`.

dpinol (changes requested):
when you fix a bug, you should always create a regression test which detects that we have actually fixed the bug

packages/botonic-nlp/tests/tasks/ner/process/processor.test.ts
```diff
@@ -1,31 +1,83 @@
...
   )
-  test('Process sample', () => {
-    expect(
-      sut
-        .process([
-          { class: '', entities: [], text: 'I love this leather jacket' },
-        ])
-        .x.arraySync()
-    ).toEqual([[6, 11, 1, 27, 0, 0, 0, 0, 0, 0, 0, 0]])
+  // The samples text has been chosen so we have a short and a long sentence.
```
dpinol, line 12: always use code if possible instead of comments.
eg.
shortTextSample = {class: ,...}
longTextSample = {class: ,...}
sut.process([shortTextSample, longTextSample])

in the comments you should always write WHY (if necessary).
in this why is it important to test both short & long text? or did you mean single vs multiple sentences?

dpinol (changes requested):

dpinol (approved):

dpinol:
fix test

AlbertGom (approved):

PR #1476, opened 2021-04-23
feat(plugin-text-classification): TextClassifier implemented #BOT-349

elozano98:
**Depends on #1475 .** Please, review it first. ⚠️ 

## Description
The `TextClassifier` class is responsible for classifying a given text. 

This class takes the given text and:
1. Processes it to get the input tensor.
2. Uses the model to predict the output.
3. Processes the output so it returns a list with all possible intents and their corresponding confidences.

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-plugin-text-classification/src/model/model-info.ts
+++ b/packages/botonic-plugin-text-classification/src/model/model-info.ts
@@ -2,6 +2,7 @@ import { TextClassificationConfig } from '@botonic/nlp/lib/tasks/text-classifica
 import { Locale } from '@botonic/nlp/lib/types'
 import { LayersModel, loadLayersModel } from '@tensorflow/tfjs'
 import axios, { AxiosPromise } from 'axios'
+import { fetch } from 'cross-fetch'
 
 import { CONFIG_FILENAME, MODEL_FILENAME } from '../constants'
 
@@ -11,7 +12,10 @@ export class ModelInfo {
 
   constructor(readonly locale: Locale, readonly uri: string) {
     this.config = axios({ url: `${this.uri}/${CONFIG_FILENAME}` })
-    this.model = loadLayersModel(`${this.uri}/${MODEL_FILENAME}`)
+    // fetchFunc is defined to replace window.fetch to work in node environment
+    this.model = loadLayersModel(`${this.uri}/${MODEL_FILENAME}`, {
+      fetchFunc: fetch,
+    })
   }
 
   async getConfig(): Promise<TextClassificationConfig> {
--- a/packages/botonic-plugin-text-classification/src/model/text-classifier.ts
+++ b/packages/botonic-plugin-text-classification/src/model/text-classifier.ts
@@ -7,9 +7,6 @@ import { LayersModel } from '@tensorflow/tfjs'
 import { PredictionProcessor } from '../process/prediction-processor'
 import { TextProcessor } from '../process/text-processor'
 
-// @ts-ignore
-global.fetch = fetch
-
 export class TextClassifier {
   private readonly textProcessor: TextProcessor
   private readonly predictionProcessor: PredictionProcessor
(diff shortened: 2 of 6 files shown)


packages/botonic-plugin-text-classification/src/model/text-classifier.ts
```diff
@@ -1,14 +1,34 @@
...
-import { TextClassificationConfig } from '@botonic/nlp/lib/tasks/text-classification/storage/types'
+import type { TextClassificationConfig } from '@botonic/nlp/lib/tasks/text-classification/storage/types'
+import type { Tensor3D } from '@tensorflow/tfjs'
 import { LayersModel } from '@tensorflow/tfjs'
 
+import { PredictionProcessor } from '../process/prediction-processor'
+import { TextProcessor } from '../process/text-processor'
+
+// @ts-ignore
+global.fetch = fetch
```
dpinol, line 11: where do you use fetch?
Your link says it's a dirty hack we shoulnt use in production https://stackoverflow.com/a/51466112/145289
remove both usages if not required. if required, try to encapsulate in  a function or use a 3rd party package as suggested in  stackoverflow

dpinol (changes requested):

packages/botonic-plugin-text-classification/src/model/text-classifier.ts
vanbasten17: @dpinol , the reason of this is that `tfjs` uses fetch internally to load a model.json given a url, so the package will work correctly in browser envs but not in lambda (node). In the same thread there are some solutions like `node-fetch` or `cross-fetch`. We should give them a try :)

vanbasten17 (changes requested):
Let's solve the non-usage in production of `global.fetch` with a more suitable solution, do some tests by deploying the bot and then we can go for it.

packages/botonic-plugin-text-classification/src/model/text-classifier.ts
dpinol: @elozano98 can you create a ticket to solve this in a more elegant way?
Please add st like this:

> Your link says it's a dirty hack we shoulnt use in production https://stackoverflow.com/a/51466112/145289
> remove both usages if not required. if required, try to encapsulate in a function or use a 3rd party package as suggested in stackoverflow

elozano98: I have created [a ticket](https://linear.app/hubtype/issue/BOT-357/remove-fetch-from-nlp-plugins) for this (will be implemented in another PR).

elozano98: Fixed, cross-fetch used! @dpinol

vanbasten17 (approved):

packages/botonic-plugin-text-classification/src/model/model-info.ts
```diff
@@ -11,7 +12,10 @@ export class ModelInfo {
 
   constructor(readonly locale: Locale, readonly uri: string) {
     this.config = axios({ url: `${this.uri}/${CONFIG_FILENAME}` })
-    this.model = loadLayersModel(`${this.uri}/${MODEL_FILENAME}`)
+    // fetchFunc is defined to replace window.fetch to work in node environment
```
dpinol, line 15: please create a ticket to unify both ModelInfo classes.
 
* create a single class template at botonic-nlp.  Just modify this
```
 export class GenericModelInfo<T> {
  private readonly config: AxiosPromise<T>
async getConfig(): Promise<T> {
```
* Move the constants CONFIG_FILENAME & MODEL_FILENAME to be static fields of GenericModelInfo. constants should be as close as possible to the class that use them. They only need to be in a centralized placed when they need to be configured in a centralized way through env vars. even in that case, there should be a class which encapsulates them duplicating them and encapsulating them
* at each ner & textclassifer, 
export class NerModelInfo extends GenericModelInfo<NerConfig> {}

dpinol (approved):

PR #1477, opened 2021-04-23
feat(plugin-text-classification): TextProcessor implemented #BOT-345

elozano98:
**Depends on #1476.** Please, review it first. ⚠️ 

## Description
The `TextProcessor` class is responsible for process the given text and generate the input tensor and its corresponding sequence.

## Context
We need the input tensor for the model prediction and the sequence for unmasking the decoded input tensor.

## Testing

The pull request...

- [ ] has unit tests
- [ ] has integration tests

### DIFF
--- a/packages/botonic-plugin-text-classification/src/process/text-processor.ts
+++ b/packages/botonic-plugin-text-classification/src/process/text-processor.ts
@@ -4,8 +4,7 @@ import {
   UNKNOWN_TOKEN,
 } from '@botonic/nlp/lib/preprocess/constants'
 import { Preprocessor } from '@botonic/nlp/lib/preprocess/preprocessor'
-import type { Tensor2D } from '@tensorflow/tfjs'
-import { tensor } from '@tensorflow/tfjs'
+import { Tensor2D, tensor2d } from '@tensorflow/tfjs'
 
 export class TextProcessor {
   private readonly encoder: LabelEncoder
@@ -21,7 +20,7 @@ export class TextProcessor {
     const sequence = this.preprocessor.preprocess(text, PADDING_TOKEN)
     const maskedSequence = this.maskUnknownTokens(sequence)
     const encodedSequence = this.encoder.encode(maskedSequence)
-    const input = tensor([encodedSequence]) as Tensor2D
+    const input = tensor2d([encodedSequence])
     return { sequence, input }
   }
 
--- a/packages/botonic-plugin-text-classification/src/process/text-processor.ts
+++ b/packages/botonic-plugin-text-classification/src/process/text-processor.ts
@@ -8,15 +8,19 @@ import type { Tensor2D } from '@tensorflow/tfjs'
 import { tensor } from '@tensorflow/tfjs'
 
 export class TextProcessor {
+  private readonly encoder: LabelEncoder
+
   constructor(
     readonly vocabulary: string[],
     private readonly preprocessor: Preprocessor
-  ) {}
+  ) {
+    this.encoder = new LabelEncoder(this.vocabulary)
+  }
 
   process(text: string): { sequence: string[]; input: Tensor2D } {
     const sequence = this.preprocessor.preprocess(text, PADDING_TOKEN)
     const maskedSequence = this.maskUnknownTokens(sequence)
-    const encodedSequence = this.encodeSequence(maskedSequence)
+    const encodedSequence = this.encoder.encode(maskedSequence)
     const input = tensor([encodedSequence]) as Tensor2D
     return { sequence, input }
   }
@@ -25,10 +29,5 @@ export class TextProcessor {
     return sequence.map(token =>
       this.vocabulary.includes(token) ? token : UNKNOWN_TOKEN
     )
-  }
-
-  private encodeSequence(sequence: string[]): number[] {
-    const encoder = new LabelEncoder(this.vocabulary)
-    return encoder.encode(sequence)
   }
 }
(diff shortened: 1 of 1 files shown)


packages/botonic-plugin-text-classification/src/process/text-processor.ts
```diff
@@ -8,7 +14,21 @@ export class TextProcessor {
...
+  }
+
+  private maskUnknownTokens(sequence: string[]): string[] {
+    return sequence.map(token =>
+      this.vocabulary.includes(token) ? token : UNKNOWN_TOKEN
+    )
+  }
+
+  private encodeSequence(sequence: string[]): number[] {
+    const encoder = new LabelEncoder(this.vocabulary)
```
dpinol, line 31: create it in constructor

packages/botonic-plugin-text-classification/src/process/text-processor.ts
```diff
@@ -8,7 +14,21 @@ export class TextProcessor {
...
-    //TODO: pending to be implemented
-    return { sequence: null, input: null }
+    const sequence = this.preprocessor.preprocess(text, PADDING_TOKEN)
+    const maskedSequence = this.maskUnknownTokens(sequence)
+    const encodedSequence = this.encodeSequence(maskedSequence)
+    const input = tensor([encodedSequence]) as Tensor2D
+    return { sequence, input }
+  }
+
+  private maskUnknownTokens(sequence: string[]): string[] {
```
dpinol, line 24: here we can't use IndexItems.maskUnknownTokens right?
what about creating a Vocabulary class which just contains an indexedTokens field of type IndexedItem and add a  maskUnknownTokens method?
I guess it's better to do it after you merge this & the indexedItems PR

packages/botonic-plugin-text-classification/src/process/text-processor.ts
```diff
@@ -1,5 +1,11 @@
+import { LabelEncoder } from '@botonic/nlp/lib/encode/label-encoder'
+import {
+  PADDING_TOKEN,
+  UNKNOWN_TOKEN,
+} from '@botonic/nlp/lib/preprocess/constants'
 import { Preprocessor } from '@botonic/nlp/lib/preprocess/preprocessor'
 import type { Tensor2D } from '@tensorflow/tfjs'
+import { tensor } from '@tensorflow/tfjs'
 
 export class TextProcessor {
```
dpinol, line 10: this class is very similar to ner's TextProcessor. does it make sense to move it to botonic-nlp and adapt it to each task with a decorator or strategy design pattern?

dpinol (changes requested):

packages/botonic-plugin-text-classification/src/process/text-processor.ts
elozano98: In the future we can use `IndexedItems.maskUnknownItems` method. But first, we need to merge the PR where it is implemented. This would be implemented after both PR are merged. 👍🏽

packages/botonic-plugin-text-classification/src/process/text-processor.ts
elozano98: Both tasks of the nlp package contain a `Processor` class that is responsible for creating the Input and expected Output tensors.

Also, in the same way, each plugin of both NLP tasks (text-classification and ner) contains a `TextProcessor` class which is responsible for creating **only** the Input tensor.

One important fact is that the creation process of the Input tensor is the same for all implemented NLP tasks for the moment, but the expected Output creation is different for each one.

I think if maybe we should split `Processor` into two different classes:
1. Input tensor creation (common class for all NLP tasks).
2. Expected Output creation (each NLP task has its own implementation).

This way, we could reuse the class that creates the Input tensor in the plugins.

> **Important:** I think that this Processor class was not reused in previous implementations because of an error with Tensor types.

Anyways, I think it would be nice to discuss and define it in a call when possible.

@dpinol @ericmarcos @vanbasten17

dpinol: The 2 TextProcessor classes are almost identical (the only difference is in the call to preprocess which is now only refactored in 1 of the 2 tasks). 
I'm not sure why you mention the 2 Processor classes look more diverse and they could be kept separate (though their maskUnknownTokens & preprocessing call could be factored out)
 

image

dpinol (changes requested):
review my comment

packages/botonic-plugin-text-classification/src/process/text-processor.ts
elozano98: You are right, the two TextProcessor are almost identical. When I said that the `Processor` classes look more diverse I referred to the `Processor` class of [`ner`](https://github.com/hubtype/botonic/blob/feat/text-processor/packages/botonic-nlp/src/tasks/ner/process/processor.ts) and [`text-classification`](https://github.com/hubtype/botonic/blob/feat/text-processor/packages/botonic-nlp/src/tasks/text-classification/process/processor.ts) tasks in `botonic-nlp-package`.

dpinol: So, what's the plan to unify them? Is there a PR or a ticket?

elozano98: Here is the [ticket](https://linear.app/hubtype/issue/BOT-358/split-preprocessors-into-2-classes). The plan is to split the `Processor` classes into 2 different classes:
1. For creating the input tensor and sequence.
2. For creating the expected output tensor.

This way we can use the input tensor generator class in all NLP tasks and plugins. 
Also, each NLP task will have its own class responsible for generating the expected output tensor.

dpinol (approved):

vanbasten17 (approved):

PR #1498, opened 2021-04-29
Feat/create and delete entry

AinaVendrell:
## Description
Add `deleteEntry` and `createEntryWithId` in ManageContentful

`createEntryWithId` will not define the fields so the entry will not be published (draft). To define the fields and publish the entry you can use the function `updateFields` that already exists.

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
@@ -47,6 +47,34 @@ export class ManageContentful implements ManageCms {
     return this.environment
   }
 
+  async deleteEntry(contentId: ContentId): Promise<void> {
+    const environment = await this.getEnvironment()
+    const getEntry = async () => {
+      try {
+        return await environment.getEntry(contentId.id)
+      } catch (e) {
+        throw new ResourceNotFoundCmsException(contentId, e)
+      }
+    }
+    const oldEntry = await getEntry()
+    if (oldEntry) {
+      await oldEntry.unpublish()
+      await oldEntry.archive()
+      await oldEntry.delete()
+    }
+  }
+
+  async createEntryWithId(contentId: ContentId) {
+    const environment = await this.getEnvironment()
+    try {
+      await environment.createEntryWithId(contentId.model, contentId.id, {
+        fields: {},
+      })
+    } catch (e) {
+      console.error(`ERROR while creating entry ${contentId}:`, e)
+    }
+  }
+
   async updateFields(
     context: ManageContext,
     contentId: ContentId,
--- a/packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
@@ -8,7 +8,7 @@ import { Entry } from 'contentful-management/dist/typings/entities/entry'
 // eslint-disable-next-line node/no-missing-import
 import { Environment } from 'contentful-management/dist/typings/entities/environment'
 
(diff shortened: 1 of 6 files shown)


packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
```diff
@@ -47,6 +47,34 @@ export class ManageContentful implements ManageCms {
...
+      try {
+        return await environment.getEntry(contentId.id)
+      } catch (e) {
+        throw new ResourceNotFoundCmsException(contentId, e)
+      }
+    }
+    const oldEntry = await getEntry()
+    if (oldEntry) {
+      await oldEntry.unpublish()
+      await oldEntry.archive()
```
dpinol, line 62: I think that archive is not necessary here

dpinol, line 62: I think that archive is not necessary here

packages/botonic-plugin-contentful/src/manage-cms/manage-cms-error.ts
```diff
@@ -26,6 +26,20 @@ export class ErrorReportingManageCms implements ManageCms {
       .catch(this.handleError('updateFields', context, contentId, fields))
   }
 
+  deleteEntry<T extends cms.Content>(contentId: ContentId): Promise<void> {
+    return this.manageCms
+      .deleteEntry(contentId)
+      .catch(this.handleError('deleteEntry', undefined, contentId, {}))
+  }
+
+  createEntryWithId<T extends cms.Content>(
```
dpinol, line 35: I think that often we will need to create entries with autogenerated id. To avoid having 2 different calls, I suggest changin to :
`createEntry(readonly model: ContentType, id: string): Promise<ContentId>`

and in the future if we want to support autogenerate id, we add "?" to id and call contentful createEntryWithId (without breaking compatibility)

packages/botonic-plugin-contentful/src/manage-cms/manage-cms-error.ts
```diff
@@ -26,6 +26,20 @@ export class ErrorReportingManageCms implements ManageCms {
...
+    contentId: ContentId
+  ): Promise<void> {
+    return this.manageCms
+      .createEntryWithId(contentId)
+      .catch(this.handleError('createEntryWithId', undefined, contentId, {}))
+  }
+
   copyField<T extends cms.Content>(
```
dpinol, line 43: please remove all "<T extends cms.Content>" in manage classes (even previous ones)

dpinol (changes requested):

packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
```diff
@@ -47,6 +47,34 @@ export class ManageContentful implements ManageCms {
...
+  }
+
+  async createEntryWithId(contentId: ContentId) {
+    const environment = await this.getEnvironment()
+    try {
+      await environment.createEntryWithId(contentId.model, contentId.id, {
+        fields: {},
+      })
+    } catch (e) {
+      console.error(`ERROR while creating entry ${contentId}:`, e)
```
dpinol, line 74: if you catch the exception, the client application will never know that the creation failed.
in libraries, it's usually better to let it fail. In this case, you can wrap it so that any CMS implementation will throw the same type of exception
```
} catch (e) {
        throw new CmsException("ERROR while creating entry",e)
      }
```

dpinol (changes requested):

dpinol (approved):

elozano98 (approved):
Good job!! 👍🏽

PR #1523, opened 2021-05-05
refactor(nlp): Ner and Text Classification model layers renamed #BOT-381

elozano98:
## Description
When creating a bot with plugins for Text Classification and NER tasks we get an error.
This error is caused because models used for both tasks have the same names for some layers.

**Solution**: Renaming the model layers to have different names for each task.

### DIFF
--- a/packages/botonic-nlp/src/tasks/text-classification/models/simple-nn.ts
+++ b/packages/botonic-nlp/src/tasks/text-classification/models/simple-nn.ts
@@ -8,12 +8,13 @@ import {
   train,
 } from '@tensorflow/tfjs-node'
 
-import {
-  DEFAULT_DROPOUT,
-  DEFAULT_LEARNING_RATE,
-  DEFAULT_UNITS,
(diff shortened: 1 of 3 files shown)


packages/botonic-nlp/src/tasks/text-classification/models/simple-nn.ts
```diff
@@ -8,13 +8,14 @@ import {
...
-import {
-  DEFAULT_DROPOUT,
-  DEFAULT_LEARNING_RATE,
-  DEFAULT_UNITS,
-} from './constants'
 import { TextClassifierParameters } from './types'
 
+const MODEL_NAME = 'SimpleTextClassifier'
+
+const DEFAULT_DROPOUT = 0.3
```
dpinol, line 15: :clap: now it's clearer where these constants belong to

dpinol (approved):

vanbasten17 (approved):

PR #1525, opened 2021-05-06
feat(nlp): group imports #BOT-326

elozano98:
## Description
Group exports of modules in their own module's `index.ts`.

## Context
Grouping exports simplifies imports of the NLP package modules.

## TODO:
**- Simplify imports of `botonic-nlp` in both plugins.**

### DIFF
--- a/packages/botonic-nlp/src/dataset/index.ts
+++ b/packages/botonic-nlp/src/dataset/index.ts
@@ -1,2 +1,5 @@
+export * from './data-augmenter'
 export * from './dataset'
-export * from './types'
+export * from './entities-parser'
+export * from './input-data-parser'
+export * from './input-data-reader'
--- a/packages/botonic-nlp/src/dataset/index.ts
+++ b/packages/botonic-nlp/src/dataset/index.ts
@@ -1,0 +1,4 @@
+import { Dataset } from './dataset'
+import { Sample } from './types'
+
+export { Dataset, Sample }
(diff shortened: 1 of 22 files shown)


packages/botonic-nlp/src/tasks/ner/botonic-ner.ts
```diff
@@ -1,31 +1,25 @@
...
-import { NerConfigStorage } from './storage/config-storage'
+import { unique } from '../../utils'
+import { createBiLstmModel, NER_TEMPLATE, NerModelParameters } from './models'
+import {
+  Entity,
+  NEUTRAL_ENTITY,
+  PredictionProcessor,
+  Processor,
+} from './process'
+import { NerConfig, NerConfigStorage } from './storage'
```
vanbasten17, line 22: very much cleaner! 💯

packages/botonic-nlp/src/dataset/index.ts
```diff
@@ -0,0 +1,4 @@
+import { Dataset } from './dataset'
+import { Sample } from './types'
+
+export { Dataset, Sample }
```
vanbasten17, line 4: if you want to expose all the exported members within `dataset` and `types`, I think you can also do:

```
export * from './dataset'
export * from './types'
```

vanbasten17: same for all the other `index.ts`

vanbasten17 (approved):

packages/botonic-nlp/src/dataset/index.ts
dpinol: yes, if you want to restrict what to expose, you could do export  { X, Y } from '...' but it's fine to export everything

dpinol (changes requested):

dpinol (approved):

PR #1526, opened 2021-05-06
refactor(plugin-ner, plugin-text-classification): Update Plugins Imports #BOT-382

elozano98:
**Depends on #1525.** Please, review it first. ⚠️ 

## Description
Imports from both NLP plugins (plugin-ner & plugin-text-classification) have been updated. 

## Testing

The pull request...
- [x] has unit tests
- [x] has integration tests

### DIFF
--- a/packages/botonic-nlp/package-lock.json
+++ b/packages/botonic-nlp/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/nlp",
-  "version": "0.19.0-alpha.3",
+  "version": "0.19.0-alpha.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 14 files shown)


elozano98:
Tests fail because the alpha version of the NLP package does not include the changes of #1525. An alpha version that includes them will be published after that PR is approved.

vanbasten17 (changes requested):
@dpinol , should this be a good moment to leave @botonic/nlp's main pointing to `./lib/index.js` instead of importing from `lib`? Two facts here:
* Not importing from lib (importing directly from ts src) would make the plugins to do an extra compilation step.
* Importing directly from lib would make this plugin to import already minified code and to have a faster compilation process
What do you think?

vanbasten17:
Also @dpinol , what can happen is that when trying to import directly from this new main pointing to lib/index.js in plugin's code, could be that it loads Node's exclusive code that when used cause to not compile. What will be the best way to discretize between browser compatible functions/classes and Node ones?

> @dpinol , should this be a good moment to leave @botonic/nlp's main pointing to `./lib/index.js` instead of importing from `lib`? Two facts here:
> 
> * Not importing from lib (importing directly from ts src) would make the plugins to do an extra compilation step.
> * Importing directly from lib would make this plugin to import already minified code and to have a faster compilation process
>   What do you think?

dpinol:
> @dpinol , should this be a good moment to leave @botonic/nlp's main pointing to `./lib/index.js` instead of importing from `lib`? Two facts here:
> 
>     * Not importing from lib (importing directly from ts src) would make the plugins to do an extra compilation step.
> 
>     * Importing directly from lib would make this plugin to import already minified code and to have a faster compilation process
>       What do you think?

do you mean src vs lib, or lib/index.js vs lib?
as far as I remember, in last meeting we decided to keep "the status quo" which is that TS projects publish the transpiled code to js (and set main), and JS publish the source code only. If you

dpinol (approved):

vanbasten17 (approved):

PR #1529, opened 2021-05-07
refactor(nlp): Dataset module refactor #BOT-385

elozano98:
## Description
The Dataset module has been simplified.

- `DatasetLoader` class has been removed.
- `InputParser` is the class responsible for reading YAML files and parsing the content.
- `EntitiesParser` is not static anymore and processes a `string`, not a `Sample`.
- Entities defined inside a `Sample` object do not include the text anymore, they only include `start`, `end`, `label`.

## Testing

The pull request...

- [x] has unit tests
- [x] has integration tests


vanbasten17 (approved):

packages/botonic-nlp/src/dataset/input-parser.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    let classes: string[] = []
+    let entities: string[] = []
+    let samples: Sample[] = []
+
+    files.forEach(filePath => {
+      const parsedData = this.parseInputFile(filePath)
+      classes = classes.concat(parsedData.classes)
+      entities = entities.concat(parsedData.entities)
+      samples = samples.concat(parsedData.samples)
```
dpinol, line 36: Since unique will create a Set, It's more efficient to directly use a Set (and to Array.from in the return).
efficiently is specially important here if we have many samples to avoid using too much memory

packages/botonic-nlp/src/dataset/input-parser.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    return readdirSync(this.path)
+      .filter(fileName => this.ALLOWED_EXTENSIONS.includes(extname(fileName)))
+      .map(fileName => join(this.path, fileName))
+  }
+
+  private parseInputFile(path: string): ParsedData {
+    const inputData: InputData = loadYaml(readFileSync(path))
+    const className: string = inputData.class ?? ''
```
dpinol, line 60: this class is doing 2 things (not SRP):
* load files
* parse the files.
This will make :
* testing the parsing difficult
* the 3rd party developers won't be able to configure/extend the parsing algorithms
* the 3rd party developers won't be able to read the data from somewhere else
* you can't mock the data parsing when testing the main class

Maybe we should have a main class DataParser which loads gets the InputData and calls InputParser.parseData¡, which contains all the code of this function except the first line
 the constructor of DataParser :
```
constructor(private readonly path: string, readonly parser = new DataParser()) {
}
```

dpinol (changes requested):

packages/botonic-nlp/src/dataset/input-parser.ts
elozano98: Since Set doesn't have a `concat` method, it would look something like:

```typescript
parse(): ParsedData {
  const files = this.getInputFiles()

  const classes: Set<string> = new Set()
  const entities: Set<string> = new Set()
  const samples: Set<Sample> = new Set()

  files.forEach(filePath => {
    const parsedData = this.parseInputFile(filePath)
    parsedData.classes.forEach(c => classes.add(c))
    parsedData.entities.forEach(e => entities.add(e))
    parsedData.samples.forEach(s => samples.add(s))
  })

  return {
    classes: Array.from(classes),
    entities: Array.from(entities),
    samples: Array.from(samples),
  }
}
```

Do you think this is better? We are also itterating over each element of the parsedData. @dpinol

dpinol: yes, that's better in terms of memory efficiency (which is only important when dealing with very big data, such as samples).
in this case, 
* we avoid keeping in the array in memory multiple copies of repeated samples
* we avoid having the data in 3 places (array with repeated, set and unique array)

packages/botonic-nlp/src/dataset/entities-parser.ts
```diff
@@ -10,7 +10,8 @@ export class EntitiesParser {
 
     let text = sentence
     let m
-    let accumulation = 0
+
+    let charactersRemoved = 0 //This value represents the amount of characters removed because of entity definition removal: [shirt](product) -> shirt.
```
dpinol, line 14: Nobody likes reading long text/emails/comments. Be short
"This value represents the" adds no meaning. Always start comments directly with st like "amount of characters..."

dpinol (approved):

PR #1567, opened 2021-05-18
Fix(plugin-ner, plugin-intent-classification): fix imports to work with serve mode

vanbasten17:
## Description
* Remove imports pointing to modules (this makes the bundle to include Node deps). This is an implementation detail that has to be taken into account when developing nlp plugins.
     * Done for plugin-intent-classification
     * Done for plugin-ner 
* Safe check for process env (not included anymore in Webpack5 unless it is defined with external plugins)

## Context
Plugin intent classification and ner were not working when run in serve mode.


elozano98 (approved):
LGTM

ericmarcos (approved):

PR #1568, opened 2021-05-18
Refactor(nlp): NLP Tasks constructors refactored #BOT-329

elozano98:
## Description
`BotonicIntentClassifier` and `BotonicNer` constructors have been updated. Now they need a config interface to be initialized.

## Context
We want to simplify the main NLP classes by using a configuration interface.  

## Testing

The pull request...

- [x] has unit tests
- [x] has integration tests

### DIFF
--- a/packages/botonic-nlp/src/tasks/intent-classification/botonic-intent-classifier.ts
+++ b/packages/botonic-nlp/src/tasks/intent-classification/botonic-intent-classifier.ts
@@ -12,6 +12,7 @@ import { Preprocessor } from '../../preprocess'
 import { ModelStorage } from '../../storage'
 import { Locale } from '../../types'
 import { unique } from '../../utils'
+import { NlpTaskConfig } from '../nlp-task-config'
 import {
   createSimpleNN,
   INTENT_CLASSIFIER_TEMPLATE,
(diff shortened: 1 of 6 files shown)


packages/botonic-nlp/src/tasks/intent-classification/botonic-intent-classifier.ts
```diff
@@ -20,40 +21,44 @@ import {
...
     )
-    this.predictionProcessor = new PredictionProcessor(this.intents)
+    this.predictionProcessor = new PredictionProcessor(config.intents)
   }
 
   static async load(
     path: string,
     preprocessor: Preprocessor
   ): Promise<BotonicIntentClassifier> {
     const config = new IntentClassificationConfigStorage().load(path)
```
elozano98, line 54: I am aware that I have a similar implementation of configuration in `BotonicIntentClassification` and `BotonicNer` for when the NLP task is initialized by loading a previously trained model. **This would be solved in another PR.** ⚠️

packages/botonic-nlp/src/tasks/intent-classification/botonic-intent-classifier.ts
```diff
@@ -20,40 +21,44 @@ import {
...
+export interface IntentClassifierConfig extends NlpTaskConfig {
+  intents: string[]
+}
+
 export class BotonicIntentClassifier {
+  readonly locale: Locale
+  readonly maxLength: number
+  readonly vocabulary: string[]
   readonly intents: string[]
   private processor: Processor
```
dpinol, line 33: if possible add readonly to processor &  predictionProcessor

packages/botonic-nlp/src/tasks/ner/botonic-ner.ts
```diff
@@ -21,40 +22,44 @@ import {
...
-  constructor(
-    readonly locale: Locale,
-    readonly maxLength: number,
-    entities: string[],
-    readonly vocabulary: string[],
-    private readonly preprocessor: Preprocessor
-  ) {
-    this.entities = unique([NEUTRAL_ENTITY].concat(entities))
+  constructor(config: EntityRecognizerConfig) {
+    this.locale = config.locale
```
dpinol, line 39: I don't that BotonicNer and BotonicIntentClassifier have so many fields. They are difficult to maintain & document, and since you use them in 2 different places, encapsulating them you would be worth.

Can you try this?
```
export class BotonicNer {
   readonly (private if possible) config: Readonly<EntityRecognizerConfig>
   constructor(config: EntityRecognizerConfig) {
       this.config = {...config, entities : unique([NEUTRAL_ENTITY].concat(config.entities))}
   }
}
```

dpinol (changes requested):

packages/botonic-nlp/src/tasks/ner/botonic-ner.ts
elozano98: I totally agree with you, I was going to apply something similar to this in the next PR. 🤔

dpinol (approved):

vanbasten17 (approved):

PR #1465, opened 2021-04-19
Chore/botonic Developer Experience (DX)

dpinol:
## Description

Package with baseline Developer Experience (linters, tests, node scripts, ...) configuration for typescript or javascript projects.
It does not have any botonic dependency (not even react) 

## Context

We want an easy to maintain common configuration not only for botonic packages, but also for projects that use botonic (eg. bots)

## Approach taken / Explain the design

The package contains 3 parts:
* eslint, jest, typescript dependencies
* Baseline typescript, jest & eslint configuration (through [botonic-eslint-config](https://github.com/hubtype/botonic/tree/master/packages/botonic-eslint-config)). This configuration will be maintained in this botonic-ci project.
* Sample configuration to be copied into your project. 

## To document / Usage example

> @botonic/ci is a package with baseline CI (linters, tests, node scripts, ...) for typescript or javascript projects.
> It provide a common configuration not only for botonic packages, but also for projects that use botonic (eg. bots)
> It does not have any botonic dependency (not even react) 
>
> * Install this package
> ```
> npm run -D @botonic/ci
> ```
> * Copy the contents of the `sample-config` folder to the root of your project 
>   (Merge this package.json's scripts into your project ones)
> * These files just import the configuration maintained within this project. 
>   To adapt them to your project needs, you just need to patch the specific options after importing the baseline.
>   See instructions on each of these files.
>

PR #1570, opened 2021-05-20
botonic Developer Experience (DX) 2/3

dpinol:
#1465 created a @botonic/dx & @botonic/eslint-config PoC. After testing it in new projects, many issues were detected.
This PR solves these issues. There were many, but the main ones were:
* run prettier con config .js, .json, .md and css files
* avoid "node/no-unpublished-import" eslint errors when importing dev dependencies from tests
* allow requires on .js files 
More issues fixes at #1602

### DIFF
--- a/packages/botonic-dx/sample-config/package.json
+++ b/packages/botonic-dx/sample-config/package.json
@@ -5,9 +5,10 @@
     "build-ci": "tsc -b tsconfig.tests.json",
 
     "##########  LINT": "",
-    "lint": "npm run lint-core -- --fix",
-    "lint-ci": "npm run lint-core -- -c ../.eslintrc-ci.js",
-    "lint-core": "eslint_d --cache --quiet '*.js' '{src,tests}/**/*.[jt]{s,sx}'",
+    "lint": "npm run lint-core -- --fix && npm run prettier",
+    "lint-ci": "npm run lint-core -- -c .eslintrc-ci.js && npm run prettier",
+    "lint-core": "eslint_d --cache --quiet  '*.md' './*.js' '*.js{,on}' '{src,tests}/**/*.[jt]{s,sx}'",
+    "prettier": "prettier --no-error-on-unmatched-pattern --write '**/*.md' '**/*.css' ",
 
     "########## DEBUG": "",
     "test": "jest",
--- a/packages/botonic-dx/sample-config/package.json
+++ b/packages/botonic-dx/sample-config/package.json
@@ -20,5 +20,15 @@
     "preversion": "npm run lint",
     "version": "git add -A src tests",
     "postversion": "git push && git push --tags"
-  }
+    },
+    "engines": {
+      "node": ">=12.22.1"
+    },
+    "files": [
+      "bin/**",
+      "doc/**",
+      "lib/**",
+      "src/**",
+      "README.md"
+    ]
 }
--- a/packages/botonic-dx/sample-config/package.json
+++ b/packages/botonic-dx/sample-config/package.json
@@ -17,8 +17,9 @@
 
     "########## PUBLISH": "",
     "prepublishOnly": "npm run build && npm run lint-ci && npm test",
-    "preversion": "npm run lint",
-    "version": "git add -A src tests",
+    "preversion": "npm run lint-ci",
+    "# version issue": "See https://github.com/npm/cli/issues/2010. Temporarily checking all files have been git-added",
+    "version": "git add package.json && git diff --exit-code && git commit",
     "postversion": "git push && git push --tags"
     },
     "engines": {
--- a/packages/botonic-dx/.gitignore
+++ b/packages/botonic-dx/.gitignore
@@ -1,0 +1,3 @@
+#not adding lock to avoid noise in repository
+#since it only contains dev deps, upgrading too far will not breaking in runtime
+package-lock.json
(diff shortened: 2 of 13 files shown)


packages/botonic-dx/sample-config/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "scripts": {
+    "##########  BUILD": "",
+    "build": "tsc",
+    "build-ci": "tsc -b tsconfig.tests.json",
+
+    "##########  LINT": "",
+    "lint": "npm run lint-core -- --fix && npm run prettier",
+    "lint-ci": "npm run lint-core -- -c ../.eslintrc-ci.js && npm run prettier",
```
asastre, line 9: Should this be `.eslintrc-ci.js` instead of `../.eslintrc-ci.js`?

packages/botonic-dx/sample-config/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "lint-core": "eslint_d --cache --quiet  '*.md' './*.js' '*.js{,on}' '{src,tests}/**/*.[jt]{s,sx}'",
+    "prettier": "prettier --write '**/*.md' '**/*.css' ",
```
asastre, line 11: I would add here the `--no-error-on-unmatched-pattern` flag in order to not get an error if no `css` or `md` files are found.

packages/botonic-dx/sample-config/package.json
dpinol: Yes! And I hadn't realized because this [issue](https://github.com/mantoni/eslint_d.js/pull/169) was ignoring the -c flag.
Fixed in eslint_d 10.1.3

packages/botonic-dx/sample-config/package.json
dpinol: good point. added!

packages/botonic-dx/sample-config/package.json
```diff
@@ -16,8 +17,19 @@
 
     "########## PUBLISH": "",
     "prepublishOnly": "npm run build && npm run lint-ci && npm test",
-    "preversion": "npm run lint",
-    "version": "git add -A src tests",
+    "preversion": "npm run lint-ci",
+    "# version issue": "See https://github.com/npm/cli/issues/2010. Temporarily checking all files have been git-added",
+    "version": "git add package.json && git diff --exit-code && git commit",
     "postversion": "git push && git push --tags"
```
asastre, line 23: As I commented [here](https://github.com/hubtype/botonic/pull/1465#discussion_r638544484), why not adding the `package-lock.json` in the `git add` command? If you don't do it, it will remain changed after the commit is done.

Also, added all the commands in the `postversion` script because doing it in the `version` script wasn't adding the files correctly.

This is my suggestion:
```
"preversion": "npm run lint-ci",
"postversion": "git add package.json package-lock.json && git diff --exit-code && git commit && git push && git push --tags"
```

dpinol: My idea was not even adding package-lock.json to git because:
* these packages only incorporate dev dependencies. So, it's not critical to freeze to a specific version because they won't reach production
* lock file is long and add noise when visually inspecting diffs between commits
* when there are merge conflicts, it's a pain to merge lock files 

@marcpdw do you think we should add package-lock.json to git for DX packages?

dpinol: do we still need to move the commands to postversion if we remove package-lock.json?

asastre: If the idea is to not add `package-lock.json` file to git, we should add it to `.gitignore` right? I don't see it there, am I missing something?. If we don't git ignore the file, it will be modified anyway when doing `npm version` and it will be uncommited with the actual scripts configuration.

If we ignore `package-lock.json` in git, the scripts are fine as they are now, but if we want to include `package-lock.json`, we have to do `git add` of both files in the `postversion` script because otherwise the commit will include only the `package.json` file.

dpinol: thanks, good point, I'll add package-lock.json to .gitignore if @marcrabat & @marcpdw agree on my [rationale](https://github.com/hubtype/botonic/pull/1570#discussion_r644832062) for not adding the lock file to git

vanbasten17: it sounds reasonable to me @dpinol . Beware adding package-lock in root's gitignore, as it would ignore it for all packages. I understand that adding inside this folder would override the one in the root, is it like this?

marcpdw: My opinion is that we either add `package-lock` or we set the major versions of those packages with the caret character `^` in front.
We should have it fixed for the compatibility of the plugins. I tried to use the same config as _@botonic/dx_ on `@hubtype/eslint-config` and one of the plugins was already deprecated with the version of _eslint_ I installed there.

marcpdw (approved):
Looks good, we just need to clarify the package-lock subject.

packages/botonic-dx/sample-config/package.json
dpinol: @asastre I added package-lock.json to .gitignore in dx & eslint packages.
@marcpdw we user ^ for most dependencies, except:
 * 2 dependencies didn't have it . fixed  in #1602
 * I don't use it for @botonic/eslint-config because, since we're still in pre v1 version, we break sometimes break compatibility when upgrading the minor cunk of the version

asastre (approved):
:rocket:

PR #1571, opened 2021-05-20
Feat/create and remove asset

AinaVendrell:
#1574 does some modifications on new createAsset function

## Description
Create `createAsset` and `removeAsset` functions

`createAsset` will return the id of the new asset and its url

## Testing

The pull request...

- has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/contentful/manage/manage-asset.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/manage/manage-asset.ts
@@ -5,7 +5,7 @@ import { Asset } from 'contentful-management/dist/typings/entities/asset'
 // eslint-disable-next-line node/no-missing-import
 import { Environment } from 'contentful-management/dist/typings/entities/environment'
 
-import { AssetId } from '../../cms'
+import { AssetId, CmsException } from '../../cms'
 import { ManageContext } from '../../manage-cms/manage-context'
 import * as nlp from '../../nlp'
 import { ContentfulOptions } from '../../plugin'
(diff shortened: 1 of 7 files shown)


packages/botonic-plugin-contentful/src/contentful/manage/manage-asset.ts
```diff
@@ -51,6 +51,51 @@ export class ManageContentfulAsset {
...
   }
 
+  async removeAsset(context: ManageContext, assetId: AssetId): Promise<void> {
+    const environment = await this.environment
+    try {
+      const asset = await environment.getAsset(assetId.id)
+      if (asset.isPublished()) await asset.unpublish()
+      await asset.delete()
+    } catch (e) {
+      throw new CmsException('ERROR while deleting asset', e)
```
vanbasten17, line 61: when giving information about an error, it's important to have a quick way to identify the resource in question. I would return `ERROR while deleting asset with ID: '${assetId}' .`

packages/botonic-plugin-contentful/src/contentful/manage/manage-asset.ts
```diff
@@ -51,6 +51,51 @@ export class ManageContentfulAsset {
...
+    }
+  }
+
+  async createAsset(
+    context: ManageContext,
+    title: string,
+    fileName: string,
+    contentType: string,
+    file: any,
+    description?: string
```
vanbasten17, line 71: having lots of parameters in a function can be a sign of having something hard to extend and maintain. I would do the following here:
* create an interface representing the asset you want to add
```
inferface NewAsset(AssetToCreate, ...) {
title: string,
fileName: string,
contentType: string,
file: any,
description?: string
}
```
* calling the function `createAsset` in this way `createAsset(context, {title: 'newTitle', fileName: 'fileName', ...}`

what do you think? :)

packages/botonic-plugin-contentful/src/contentful/manage/manage-contentful.ts
```diff
@@ -102,4 +102,26 @@ export class ManageContentful implements ManageCms {
...
     return this.manageAsset.copyAssetFile(context, assetId, fromLocale)
   }
+
+  async createAsset(
+    context: ManageContext,
+    title: string,
+    fileName: string,
+    contentType: string,
+    file: any,
+    description?: string
```
vanbasten17, line 112: take a look at previous comment

vanbasten17 (changes requested):
I think that you can rethink what I was suggesting to improve this, after that, you can go for it! Nice job!! 💯

vanbasten17:
now I see that @dpinol has opened a new branch to solve this!

dpinol (approved):

packages/botonic-plugin-contentful/src/contentful/manage/manage-asset.ts
dpinol: @AinaVendrell here you can do new CmsException('ERROR while deleting asset', e, assetId)

PR #1586, opened 2021-05-25
fix(cli): pass -c flag to build command

vanbasten17:
## Description
`-c` flag of deploy command was not working: https://linear.app/hubtype/issue/BOT-413/botoniccli-0185-not-using-c-flag
Just pass it as an optional argument to `build`

Deployed version to test: https://www.npmjs.com/package/@botonic/cli/v/0.18.6-alpha.0


asastre (approved):
:raised_hands:

packages/botonic-cli/README.md
```diff
@@ -21,7 +21,7 @@ $ npm install -g @botonic/cli
 $ botonic COMMAND
 running command...
 $ botonic (-v|--version|version)
-@botonic/cli/0.18.5 darwin-x64 node-v12.21.0
+@botonic/cli/0.18.6-alpha.0 darwin-x64 node-v14.16.1
```
asastre, line 24: Why the node version is changing here?

vanbasten17: nice catch! this is something that is generated automatically by oclif. In practical terms it means I have ran the publish command with node 14 in my local machine. I would downgrade to node 12 when publishing the final package.

elozano98 (approved):

PR #1589, opened 2021-05-27
Global/chores

dpinol:
Several chores improvements. Check commit by commit.

### upgrade eslint_d to enable --fix

New version 10.1.2 incorporates this PR from us so that pre-commit, which calls eslint_d --fix actually fixes the detected issues
    https://github.com/mantoni/eslint_d.js/pull/165

### improvements in prepublish package.json scripts
botonic-nlp: Run QA scripts before publishing 
dynamo/contentful: run lint_ci instead of lint, and do it first step as it's faster than other steps


elozano98 (approved):

package.json
```diff
@@ -26,7 +26,7 @@
     "jest-each": "^26.6.2",
     "jest-expect-message": "^1.0.2",
     "jest-junit": "^12.0.0",
-    "lerna": "^3.20.2",
+    "lerna": "3.22.1",
```
vanbasten17, line 29: maybe we can get rid of `lerna`, as we are not using it at all, it seems to me that yarn or npm workspaces are trendier to manage monorepos now.

vanbasten17 (approved):

package.json
dpinol: yes, we'll consider this in the near future

PR #1590, opened 2021-05-28
Contentful/Make acronyms agnostic to dots #CUS-271

dpinol:
## Description

Now Normalizer will remove the dots only when an acronym is detected.

## Context

Tokenizer splits acronyms in several words and it makes difficult to match "ok" with "o.k"

 An AcronymPreprocessor is now executed as first step of our Normalizer so that we do this conversion. It will also be applied to stopwords, keywords,...
 ```
 //acronyms
  ['o.k.', 'ok'],
  ['o.k', 'ok'],
  ['o.k ', 'ok '],
  ['o.k ', 'ok '],
  ['U.S.A', 'USA'],
  ['o.k, vale', 'ok, vale'],
  //not acronyms
  ['.a.', '.a.'],
  ['U..A', 'U..A'],
  ['a. opcion', 'a. opcion'],
  ['no separators', 'no separators'],
  ['final dot.', 'final dot.'],

```

Now Normalizer will remove the dots only when an acronym is detected. Works like "o.k.", "o.k" and "ok" will be normalized to the same stem.

## Testing

- has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/nlp/normalizer.ts
+++ b/packages/botonic-plugin-contentful/src/nlp/normalizer.ts
@@ -70,6 +70,7 @@ export class NormalizedUtterance {
    * @param onlyStopWords: true iff all tokens are stop words
    */
   constructor(
+    /** raw is actually lowercased and trimmed*/
     readonly raw: string,
     readonly words: Word[],
     private readonly onlyStopWords = false
--- a/packages/botonic-plugin-contentful/src/nlp/normalizer.ts
+++ b/packages/botonic-plugin-contentful/src/nlp/normalizer.ts
@@ -3,10 +3,12 @@ import { equalArrays } from '../util/arrays'
 import { Locale } from './locales'
 import { stemmerFor } from './stemmer'
 import {
+  DEFAULT_SEPARATORS,
   DEFAULT_SEPARATORS_REGEX,
   stopWordsFor,
   tokenizerPerLocale,
 } from './tokens'
+import { replaceAll } from './util/strings'
 
 /**
  * Both tokens and stem will be converted to the <code>stem</code>
@@ -35,7 +37,7 @@ export class StemmingBlackList {
 
 export class Word {
   /**
-   * @param token lowercase, with i18n characters converted to ascii
+   * @param token lowercase, with i18n characters converted to ascii and after executing Preprocessor
    * @param stem lowercase, stemmed. Same as token for stopwords
    */
   constructor(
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by vanbasten17 — APPROVED** [review]

File: packages/botonic-plugin-contentful/src/nlp/normalizer.ts [hunk 1]
```diff
@@ -90,17 +93,94 @@ export class NormalizedUtterance {
...
+      if (!isSeparator) {
+        txt = txt + this.preprocessWord(wOrSep)
+      } else {
+        txt = txt + wOrSep
+      }
+    }
+    return txt
+  }
+
+  private splitWordsAndSeparators(txt: string): string[] {
```
asastre, line 132: The same way we have the `DEFAULT_SEPARATORS_REGEX` constant, maybe it would be better to create a `RegExp` with the separators without the dot and use a `split` instead of using this function? The result should be the same, right?

packages/botonic-plugin-contentful/src/nlp/normalizer.ts
```diff
@@ -145,13 +226,12 @@ export class Normalizer {
       const tokenStems = stemmer.stem([token])
       words = words.concat(tokenStems.map(stem => new Word(token, stem)))
     }
-    return new NormalizedUtterance(txt, words, numStopWords == tokens.length)
+    return new NormalizedUtterance(raw, words, numStopWords == tokens.length)
```
asastre, line 229: Why passing `raw` instead of `txt` here?

packages/botonic-plugin-contentful/tests/nlp/normalizer.test.ts
```diff
@@ -9,6 +11,28 @@ import {
...
+  ['o.k ', 'ok '],
+  ['o.k ', 'ok '],
+  ['U.S.A', 'USA'],
+  ['o.k, vale', 'ok, vale'],
+  //not acronyms
+  ['.a.', '.a.'],
+  ['U..A', 'U..A'],
+  ['a. opcion', 'a. opcion'],
+  ['no separators', 'no separators'],
+  ['final dot.', 'final dot.'],
```
asastre, line 27: What about receiving a text like this: `'textwith.dotattached'`? My feeling is that the preprocess will detect it as an acronym when it shouldn't. 
If this is detected as an acronym, if one of the 2 words is a keyword, it won't be detected as such because the dot will be removed and both words will be unified, right?

asastre (commented):
Some comments/doubts about the implementation

packages/botonic-plugin-contentful/src/nlp/normalizer.ts
dpinol: We cannot use split because this function does not "by the separators", it splits "the words AND the separators"
eg. splitWordsAndSeparators("aa.bb") returns ["aa", ".", "bb"]
Not loosing the separator makes easier the reconstruction of the full input in AcronymPreprocessor.preprocess

packages/botonic-plugin-contentful/src/nlp/normalizer.ts
dpinol: because the first argument of NormalizedUtterance expected a raw :-)

packages/botonic-plugin-contentful/tests/nlp/normalizer.test.ts
dpinol: this is a good testcase, thanks!
I added it and it doesn't recognize as an acronym because we discard the word when more than 1 letter appear at any chunk

packages/botonic-plugin-contentful/src/nlp/normalizer.ts
asastre: You're totally right!

asastre (approved):
:clap: :clap:

PR #1591, opened 2021-05-28
Feat/manage entry convert field values

AinaVendrell:
## Description
Convert the values of the field when updating the contents of contentful

## Context
In contentful the link between contents is an object that contains the id of the target content, if the given value is an string id, `convertValueType` converts the id into a link.
If the value is a `buttonStyle` it converts it to 'QuickReplies' or 'Buttons' string.

## Testing

The pull request...

- has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/contentful/manage/manage-entry.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/manage/manage-entry.ts
@@ -5,7 +5,7 @@ import { Entry } from 'contentful-management/dist/typings/entities/entry'
 // eslint-disable-next-line node/no-missing-import
 import { Environment } from 'contentful-management/dist/typings/entities/environment'
 
-import { CmsException, ContentId, ContentType } from '../../cms'
+import { ButtonStyle, CmsException, ContentId, ContentType } from '../../cms'
 import { ResourceNotFoundCmsException } from '../../cms/exceptions'
 import {
   CONTENT_FIELDS,
@@ -16,11 +16,7 @@ import { ManageContext } from '../../manage-cms/manage-context'
 import * as nlp from '../../nlp'
 import { ContentfulOptions } from '../../plugin'
 import { isOfType } from '../../util/enums'
-
-class EntryLink {
-  public readonly type = 'Link'
-  constructor(public id: string, public linkType: string) {}
-}
+import { BUTTON, EntryLink, LinkType, QUICK_REPLY } from './contentful-api'
 
 export class ManageContentfulEntry {
   constructor(
@@ -46,10 +42,7 @@ export class ManageContentfulEntry {
         continue
       }
       needUpdate = true
-      const valueType = CONTENT_FIELDS.get(key)?.valueType
-      if (fields[key] !== undefined && valueType !== typeof fields[key]) {
-        fields[key] = this.convertValueType(key, fields[key])
-      }
+      fields[key] = this.convertValueType(key, fields[key])
       oldEntry.fields[field.cmsName][context.locale] = fields[key]
     }
     if (!needUpdate) {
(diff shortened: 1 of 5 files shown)


packages/botonic-plugin-contentful/src/contentful/manage/manage-entry.ts
```diff
@@ -170,4 +179,27 @@ export class ManageContentfulEntry {
       throw new ResourceNotFoundCmsException(contentId, e)
     }
   }
+
+  private convertValueType(key: ContentFieldType, field: any): any {
+    if (key === ContentFieldType.BUTTONS_STYLE) {
+      return field ? 'QuickReplies' : 'Buttons'
```
vanbasten17, line 185: are 'QuickReplies' or 'Buttons' stored in some constants to be used instead of literals?
Same for 'Asset', in line 191.

vanbasten17 (approved):
Looking good! 👏

packages/botonic-plugin-contentful/src/contentful/manage/manage-entry.ts
```diff
@@ -17,6 +17,11 @@ import * as nlp from '../../nlp'
 import { ContentfulOptions } from '../../plugin'
 import { isOfType } from '../../util/enums'
 
+class EntryLink {
```
dpinol, line 20: move to its own file. Maybe contentful-api.ts

packages/botonic-plugin-contentful/tests/contentful/manage/manage-contentful-entry.test.ts
```diff
@@ -75,4 +80,44 @@ describe('ManageContentful entries', () => {
       sut.deleteContent(context, TEST_NEW_CONTENT_ID)
     ).rejects.toThrow(CmsException)
   })
+
+  test('TEST: convertValueType converts the ButtonStyle and the FollowUp', async () => {
```
dpinol, line 84: test name is not accurate. it should not refer to how it's implemented, but about the feature.
change to st like updateFields is able to manage fields of type ButtonStyle and FollowUps

packages/botonic-plugin-contentful/tests/contentful/manage/manage-contentful-entry.test.ts
```diff
@@ -75,4 +80,44 @@ describe('ManageContentful entries', () => {
...
+    const followUpName = rndStr()
+    try {
+      // ACT
+      await sut.createContent(context, NEW_CONTENT.model, NEW_CONTENT.id)
+      await sut.createContent(
+        context,
+        FOLLOW_UP_CONTENT.model,
+        FOLLOW_UP_CONTENT.id
+      )
+      await sut.updateFields(context, NEW_CONTENT, {
```
dpinol, line 100: I'm afraid that this might fail because the the newly created contents have not yet been able to publish to the CDN.
Try to execute the test at least 10 times (you can do it from terminal with a bash script copy pasting the ts-jest command), and if it fails, wrap the 2 updateFields calls within a repeatWithBackoff. If it fails in tests, it will might also fail from flowbuilder. In this case,  I suggest creating a PR in which createContent has an optional argument to pass the fields so that a single call to contentful API is enough and then you can't have a timeout. 

Another option would be the following:
 create a function at manage-contentful.helper which, within repeatWithBackoff, executes manageCms and later checks that the 
```
asyn cfunction createContent(
    manageCms: ManageCms,
    context: ManageContext,
    model: ContentType,
    id: string
  ): Promise<void> 
```

dpinol (changes requested):

packages/botonic-plugin-contentful/src/contentful/manage/manage-entry.ts
dpinol, line 185: this have 2 issues: 
1) it relies on the implicit fact that ButtonStyle.QUICK_REPLY has value 1. It's difficult to understand and might break if we change the enum values.
2) as you can see at https://github.com/hubtype/botonic/blob/master/packages/botonic-plugin-contentful/src/contentful/contents/text.ts#L52 we support that the field is undefined (when user has not yet selected it). This is useful because some bots want to select the style dynamically when it has not been set.
so you'll need to treat the 2+1 cases individually

dpinol (changes requested):

packages/botonic-plugin-contentful/tests/contentful/manage/manage-contentful-entry.test.ts
AinaVendrell: I've executed it 20 times and it didn't fail

packages/botonic-plugin-contentful/src/contentful/manage/manage-entry.ts
```diff
@@ -181,25 +174,27 @@ export class ManageContentfulEntry {
   }
 
   private convertValueType(key: ContentFieldType, field: any): any {
+    const valueType = CONTENT_FIELDS.get(key)?.valueType
+    if (field === undefined || valueType === typeof field) return field
     if (key === ContentFieldType.BUTTONS_STYLE) {
-      return field ? 'QuickReplies' : 'Buttons'
+      return field === ButtonStyle.QUICK_REPLY ? QUICK_REPLY : BUTTON
```
dpinol, line 180: review my latest comment. you need to store "null" when this field is undefined.
Since you you're now supporting yet this case in TextFields, please create a ticket to support in a different PR.

dpinol (approved):

PR #1602, opened 2021-06-02
botonic Developer Experience (DX)  3/3

dpinol:
Continuation of  #1570. Please :warning: review it first
* eslint-ci rules were never executed
* eslint on .js files had several issues. Now temporarilly we never eslint 
* hidden files were not published


packages/botonic-dx/README.md
```diff
@@ -18,7 +18,7 @@ npm install -D @botonic/dx
 ```
 * Copy the contents of the `sample-config` folder to the root of your project 
   (Merge this package.json's scripts into your project ones)
-* These files just import the configuration maintained within this project. 
+* The files at `sample-config` just import the configuration maintained within this project.
```
asastre, line 21: I don't understand this sentence, what are you trying to say here? Are any files in `sample-config` that are not maintained within this project?

asastre (commented):
Just one comment before approval.

packages/botonic-dx/README.md
dpinol: What I mean is that we want to avoid copying from a template config file to the host project, because then maintaining them would require modifying each project which uses botonic-dx.
So the only files which are copied (the sample-config) are just stubs (imports) to the baseline config

asastre: Thanks for the explanation. 
I read it in a wrong way before and after reading it again it make sense.

asastre (approved):

marcpdw (approved):

PR #1604, opened 2021-06-04
feat(nlp): main points to lib/index.js

elozano98:
## Description
NLP package main points to lib/index.js so we can use it like:

```typescript
import { Preprocessor, Dataset, BotonicNer } from '@botonic/nlp'
```

### DIFF
--- a/packages/botonic-nlp/package.json
+++ b/packages/botonic-nlp/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@botonic/nlp",
   "version": "0.19.0-alpha.9",
+  "main": "lib/index.js",
   "license": "MIT",
   "scripts": {
     "build": "rm -rf lib && ../../node_modules/.bin/tsc",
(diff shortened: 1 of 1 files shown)


packages/botonic-nlp/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@botonic/nlp",
   "version": "0.19.0-alpha.9",
+  "main": "lib/index.js",
```
dpinol, line 4: add also 
`"types": "lib/index.d.ts"`
and the following (adapting to your case)
```
"files": [
    "bin/**",
    "lib/**",
    "src/**",
    "doc/**",
    "README.md"
  ],
```

dpinol (changes requested):

vanbasten17 (approved):

dpinol (approved):

PR #1624, opened 2021-06-10
feat(plugin-google-translate): Translation feature added

elozano98:
## Description
- Translation feature added to the plugin so there is the option to choose the target languages we want to translate the input text.
- Plugin options adapted to both features as optional.

## Approach taken
Only if we specify `translation` options in the plugin options, translation will be applied.
Only if we specify `languageDetection` options in the plugin options, language detection will be applied.

Plugin options example when both features are used:
```typescript
export const plugins = [
  {
    id: 'google-translate',
    resolve: require('@botonic/plugin-google-translate'),
    options: {
      credentials: {
        projectId: '',
        privateKeyId: '',
        privateKey: '',
        clientEmail: '',
      },
      translation: {
        targets: ['it', 'ro', 'ca'],
      },
      languageDetection: {
        whitelist: ['en'],
      },
    },
  },
]

```


packages/botonic-plugin-google-translate/src/google-translate-api-service.ts
```diff
@@ -7,34 +7,75 @@ export interface LanguageDetection {
   confidence: number
 }
 
+export interface Translation {
+  translatedText: string
+  detectedLanguageCode: string
```
dpinol, line 12: isn't this redundant if we're setting language in Input?

packages/botonic-plugin-google-translate/src/index.ts
```diff
@@ -5,27 +5,37 @@ import { AccessToken } from './access-token'
...
   async pre(request: PluginPreRequest): Promise<void> {
     try {
       if (request.input.type == INPUT.TEXT && !request.input.payload) {
         const text = request.input.data
-        request.input.language = await this.languageDetector.detect(text)
+        const translations = await this.translator.translate(
+          text,
+          this.options.translateTo
+        )
+        const detectedLanguage = await this.languageDetector.detect(text)
```
dpinol, line 34: why do we want to call detect if translation already guesses the language.
In case you need both, use Promise.all to parallelize

dpinol (changes requested):

packages/botonic-plugin-google-translate/src/translator.ts
```diff
@@ -0,0 +1,16 @@
...
+export type Translations = { [languageCode: string]: string }
+
+export class Translator {
+  constructor(private readonly service: GoogleTranslateApiService) {}
+
+  async translate(text: string, targets: string[]): Promise<Translations> {
+    const responses = await Promise.all(
+      targets.map(t => this.service.translateText(text, t))
+    )
+    const translations = {}
```
dpinol, line 12: Use map

dpinol (changes requested):

packages/botonic-plugin-google-translate/src/index.ts
elozano98: We don't reuse the detected language of the translation because the language detection can return more than one language. This is useful when a whitelist is defined, and the first one is not included, but the second is.

packages/botonic-plugin-google-translate/src/google-translate-api-service.ts
elozano98: I know we don't use it but I wanted the API Service to return the same information that receives keeping the same structure.

packages/botonic-plugin-google-translate/src/translator.ts
```diff
@@ -9,8 +9,9 @@ export class Translator {
     const responses = await Promise.all(
       targets.map(t => this.service.translateText(text, t))
     )
-    const translations = {}
-    responses.forEach((r, i) => (translations[targets[i]] = r.translatedText))
-    return translations
+    return Object.assign(
```
dpinol, line 12: I hadn't realized you returned an object instead of a list,otherwise your initial solution was ok.
But your new solution is also cool :+1:

packages/botonic-plugin-google-translate/src/google-translate-api-service.ts
```diff
@@ -7,34 +7,76 @@ export interface LanguageDetection {
   confidence: number
 }
 
+export interface Translation {
```
dpinol, line 10: this should be added to [core](https://github.com/hubtype/botonic/blob/master/packages/botonic-core/src/types/index.d.ts#L71)
@marcrabat one option to avoid polluting core with the types of plugins is using [TS declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) so that Input.translations is added in the plugin, not in core. But then if we add a plugin for another translator engine we should move it again to core, so it's up to you

packages/botonic-plugin-google-translate/src/index.ts
```diff
@@ -5,27 +5,37 @@ import { AccessToken } from './access-token'
...
     try {
       if (request.input.type == INPUT.TEXT && !request.input.payload) {
         const text = request.input.data
-        request.input.language = await this.languageDetector.detect(text)
+        if (this.options.translateTo) {
+          const translations = await this.translator.translate(
+            text,
+            this.options.translateTo
+          )
+          request.input['translations'] = translations
```
dpinol, line 35: Input type should have a translations field

dpinol (changes requested):

packages/botonic-plugin-google-translate/src/google-translate-api-service.ts
vanbasten17: yes, I think it would be better to import the intial declaration from core and extending it directly in the plugin, in favour of not pollution core with plugin definitions

packages/botonic-plugin-google-translate/src/index.ts
vanbasten17: can the same strategy as before be applied here? import the basic input declaration and merge it with the fields that are appended here

packages/botonic-plugin-google-translate/src/index.ts
```diff
@@ -5,27 +5,37 @@ import { AccessToken } from './access-token'
...
+        }
+        const detectedLanguage = await this.languageDetector.detect(text)
+        request.input['language'] = detectedLanguage || request.session.__locale
```
vanbasten17, line 38: same for language

vanbasten17 (approved):
Looking good! 👍

packages/botonic-plugin-google-translate/src/google-translate-api-service.ts
dpinol: personally I prefer to add it to core because:
* it works as "advertising" for the botonic-core users that we have a gtranslate plugin
* we (or 3rd party plugins)can reuse it for future translation plugins
but ymmv :-)

packages/botonic-plugin-google-translate/src/index.ts
dpinol: both comments refer to the same

packages/botonic-plugin-google-translate/src/index.ts
dpinol: :+1:

packages/botonic-plugin-google-translate/src/google-translate-api-service.ts
vanbasten17: pretty convincing, go for it 😜

dpinol (approved):

PR #1625, opened 2021-06-10
Contenful/handoff

dpinol:
## Description
Create `handoff` content

## Context
There wasn't any content to define the fields of the handoff an its target. With the handoff content we can avoid breaking the flow when there is an agent handoff.

Create a new component with some of the fields of [HandOffBuilder](https://github.com/hubtype/botonic/blob/HEAD/packages/botonic-core/src/types/handoff.d.ts#:~:text=export%20declare%20class-,HandOffBuilder,-%7B):
- `onFinish`: can be a payload or a target (Text, Image, Carousel)
- `queue`: reference to a desk queue
- `agent`: can be an agentEmail or an agentId
- `shadowing`: boolean 

`message`: to define the message that will be shown before the handoff
`failMessage`: to define the message that will be shown if the handoff can't be done

## Testing

The pull request...

- has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/cms/contents.ts
+++ b/packages/botonic-plugin-contentful/src/cms/contents.ts
@@ -428,6 +428,55 @@ export class ScheduleContent extends TopContent {
   }
 }
 
+export type OnFinish = Callback
+
+export class HandoffAgentEmail {
+  readonly type = 'AGENT_EMAIL'
+  constructor(readonly agentEmail: string) {}
+}
+
+export class HandoffAgentId {
+  readonly type = 'AGENT_ID'
+  constructor(readonly agentId: string) {}
+}
+
+export class HandoffQueue {
+  readonly type = 'QUEUE'
+  /**
+   * @param queue the queue id or name
+   */
+  constructor(readonly queue: Queue) {}
+}
+
+export type HandoffDestination =
+  | HandoffAgentEmail
+  | HandoffAgentId
+  | HandoffQueue
+
+/**
+ * Most CommonFields make no sense for Handoff.
+ * However, we decided to make it a TopContent since it does not depend on other content.
+ * Also CommonFields might be potentially useful.
+ */
+export class Handoff extends TopContent {
+  constructor(
+    readonly common: CommonFields,
+    readonly text: string,
+    readonly onFinish?: OnFinish,
+    //destination is optional because often it is set dynamically by the bot
+    readonly destination?: HandoffDestination,
+    readonly shadowing?: boolean
+  ) {
+    super(common, ContentType.HANDOFF)
+  }
+
+  cloneWithDestination(newDestination: HandoffDestination): this {
+    const clone = shallowClone(this)
+    ;(clone as any).destination = newDestination
+    return clone
+  }
+}
+
 /**
  * A {@link Content} which is automatically displayed after another one
  */
--- a/packages/botonic-plugin-contentful/src/cms/contents.ts
+++ b/packages/botonic-plugin-contentful/src/cms/contents.ts
@@ -440,18 +440,7 @@ export class HandoffAgentId {
   constructor(readonly agentId: string) {}
 }
 
-export class HandoffQueue {
-  readonly type = 'QUEUE'
-  /**
-   * @param queue the queue id or name
-   */
-  constructor(readonly queue: Queue) {}
-}
-
-export type HandoffDestination =
-  | HandoffAgentEmail
-  | HandoffAgentId
-  | HandoffQueue
+export type HandoffAgent = HandoffAgentEmail | HandoffAgentId
 
 /**
  * Most CommonFields make no sense for Handoff.
@@ -461,18 +450,26 @@ export type HandoffDestination =
 export class Handoff extends TopContent {
   constructor(
     readonly common: CommonFields,
-    readonly text: string,
-    readonly onFinish?: OnFinish,
-    //destination is optional because often it is set dynamically by the bot
-    readonly destination?: HandoffDestination,
+    readonly onFinish: OnFinish,
+    readonly message?: string,
+    readonly failMessage?: string,
+    //agent and queue are optional because often they are set dynamically by the bot
+    readonly queue?: Queue,
+    readonly agent?: HandoffAgent,
     readonly shadowing?: boolean
   ) {
     super(common, ContentType.HANDOFF)
   }
 
-  cloneWithDestination(newDestination: HandoffDestination): this {
+  cloneWithQueue(newQueue: Queue): this {
     const clone = shallowClone(this)
-    ;(clone as any).destination = newDestination
+    ;(clone as any).queue = newQueue
+    return clone
+  }
+
+  cloneWithAgent(newAgent: HandoffAgent): this {
+    const clone = shallowClone(this)
+    ;(clone as any).agent = newAgent
     return clone
   }
 }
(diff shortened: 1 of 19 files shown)


packages/botonic-plugin-contentful/src/cms/contents.ts
```diff
@@ -428,6 +428,58 @@ export class ScheduleContent extends TopContent {
...
+/**
+ * Most CommonFields make no sense for Handoff.
+ * However, we decided to make it a TopContent since it does not depend on other content.
+ * Also CommonFields might be potentially useful.
+ */
+export class Handoff extends TopContent {
+  constructor(
+    readonly common: CommonFields,
+    readonly text: string,
+    readonly onFinish?: OnFinish,
```
dpinol, line 475: onFinish shouldn't be optional, right?

packages/botonic-plugin-contentful/src/cms/contents.ts
```diff
@@ -428,6 +428,58 @@ export class ScheduleContent extends TopContent {
...
+    readonly destination?: HandoffDestination,
```
dpinol, line 476: Add a comment in the code explaining that destination is optional because often it is set dynamically by the bot.
Add a cloneWithDestination method so that the bot can set it dynamically

packages/botonic-plugin-contentful/src/cms/factories/content-factories.ts
```diff
@@ -219,3 +222,37 @@ export class DocumentBuilder extends MessageContentBuilder {
...
+
+export class HandoffBuilder extends MessageContentBuilder {
+  onFinish?: OnFinish
+  destination?: HandoffDestination
+  shadowing?: boolean
+  constructor(id: string, name: string, public text: string) {
+    super(id, name)
+  }
+
+  withOnFinish(onFinish?: OnFinish): this {
```
dpinol, line 234: if we don't allow OnFinish to be undefined
* remove "?"
* add argument to constructor

packages/botonic-plugin-contentful/src/cms/factories/content-factories.ts
```diff
@@ -219,3 +222,37 @@ export class DocumentBuilder extends MessageContentBuilder {
...
+    this.onFinish = onFinish
+    return this
+  }
+
+  withDestination(destination: HandoffDestination): this {
```
dpinol, line 239: add "?" since it's optional in HandOff

packages/botonic-plugin-contentful/src/contentful/contents/button.ts
```diff
@@ -161,6 +131,7 @@ type ButtonTarget = contentful.Entry<
   | QueueFields
   | HourRangeFields
   | ScheduleFields
+  | HandoffFields
```
dpinol, line 134: DRY: ButtonTarget & CallbackTarget are too similar.
Maybe ButtonTarget can be set to initially equal CallbackTarget and in the future we can improve it?

packages/botonic-plugin-contentful/src/contentful/contents/handoff.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  private onFinish(
+    entry: contentful.Entry<HandoffFields>,
+    context: cms.Context
+  ): OnFinish | undefined {
+    if (entry.fields.onFinish) {
+      return getTargetCallback(entry.fields.onFinish, context)
+    }
+    return undefined
```
dpinol, line 42: raise exception if we decide that OnFInish must be compulsory

dpinol (commented):
see my comments

packages/botonic-plugin-contentful/tests/contentful/contents/handoff.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  SPANISH,
+} from '../../../src'
+import { testContentful, testContext } from '../contentful.helper'
+
+const TEST_HANDOFF_AGENT_ID_MAIN_ID = '6I5SudoItjGmxrVZPr0qeG'
+const TEST_HANDOFF_QUEUE_MAIN_ID = '6A7D4ssRYuLufjvl1pnOzV'
+const TEST_QUEUE_ID = '62ILnVxLHOEp7aVvPMpCO8'
+
+describe('Contentful Handoff', () => {
+  test('TEST: contentful handoff with queue as destination', async () => {
```
AinaVendrell, line 15: When I link the handoff with a Queue that contains an Scheduler it fails: 
```
Error calling CMS.handoff with locale 'es' on 'handoff' with id '6A7D4ssRYuLufjvl1pnOzV'. 
Error loading Scheduler '71twiV4wcaFwhK6tSYuIKy'. Broken reference? Not published?
```
I've checked and the reference is okey, @dpinol any idea on why isn't it woking?

packages/botonic-plugin-contentful/src/contentful/contents/contents.ts
```diff
@@ -56,6 +57,7 @@ export class ContentsDelivery extends ResourceDelivery {
 
   private maxReferencesInclude() {
     return Math.max(
+      HandoffDelivery.REFERENCES_INCLUDE,
```
dpinol, line 60: :clap:

AinaVendrell:
Should onFinish be mandatory? @asastre @guinii @AlbertGom

asastre:
> Should onFinish be mandatory? @asastre @guinii @AlbertGom

I would set it as mandatory since the most common thing to do after a handoff is to show a message or a rating.
If no content has to be shown after a handoff, you can always set a payload that leads to an empty action.

packages/botonic-plugin-contentful/src/cms/contents.ts
```diff
@@ -428,6 +428,55 @@ export class ScheduleContent extends TopContent {
...
+
+/**
+ * Most CommonFields make no sense for Handoff.
+ * However, we decided to make it a TopContent since it does not depend on other content.
+ * Also CommonFields might be potentially useful.
+ */
+export class Handoff extends TopContent {
+  constructor(
+    readonly common: CommonFields,
+    readonly text: string,
```
asastre, line 464: After taking a look at the test, I think that this `text` is the one to be displayed when doing the handoff, am I right?

Maybe we should consider adding here a new attribute to store the text to be displayed if the handoff can't be done (if a queue is closed or if an agent is not available for example), what do you think?

packages/botonic-plugin-contentful/src/contentful/contents/handoff.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private destination(
+    entry: contentful.Entry<HandoffFields>
+  ): HandoffDestination | undefined {
+    if (!entry.fields.destination) return undefined
+    const AGENT_EMAIL_TYPE = 'agentEmail'
+    const AGENT_ID_TYPE = 'agentId'
```
asastre, line 56: I've seen the string literals 'agentEmail' and 'agentId' being used somewhere else in this code. Will it make sense to use them always as constants?

asastre:
The `destination` attribute in `Handoff` can be one of these 3: Queue, Agent Email or Agent ID. 

Setting the destination to an Agent Email, would it represent an auto-assign? If so, there's no way to do an auto-assign into a specific Queue to a specific agent, right? 
Right now I have to choose between setting a queue or an agent as the destination of the Handoff but doing the handoffs directly in the bots I can determine both in the same handoff.

I think this configuration can't be done with the actual changes in this PR, am I right? If I'm right, I think we should consider adding this type of configuration to `Handoff`.

asastre (approved):
:clap: Nice!! :rocket:

vanbasten17 (approved):

PR #1637, opened 2021-06-14
Feat/on-connection-change-listener

vanbasten17:
## Description
* Add feature to detect when connection has changed in order to do extra logic on websocket disconnection/connection.

## Context
* It's a priority needed for customers.

* Executing onConnectionChange everytime that websocket detects a change in the connection.

```
export const webchat = {
  onConnectionChange: (app, isOnline) => {
    if (isOnline) {
      // has been connected logic
    } else {
      // has been disconnected logic
    }
  },
}
```
**note**: listener is also executed the first time that webchat is connected after page load

### DIFF
--- a/packages/botonic-react/package-lock.json
+++ b/packages/botonic-react/package-lock.json
@@ -1,16 +1,16 @@
 {
   "name": "@botonic/react",
-  "version": "0.18.4-alpha.2",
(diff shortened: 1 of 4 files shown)


dpinol (approved):

asastre (approved):
:clap:

PR #1639, opened 2021-06-14
feat(contentful): memoizer #BOT-467

dpinol:
Depends on #1640. :warning:  **Please review it first**

## Description
Use memoization to remember forever the last successful invocation for each combination of arguments, and use its result whenever Contentful fails.

## Context

Contentful failing during 1h on 2021/6/8 and 2021/6/9.

* Create a generic Memoizer class with several strategies. 
* Create a decorator (FallbackCachedClientApi) around the Contentful API class which uses the Memoizer fallbackStrategy.
* Compose the new decorator within the existing decorator (CachedClientApi) which avoids recent calls.

FallbackCachedClientApi does not use the memoization library used by CachedClientApi (memoizee) because this library does not allow directly accessing the cache. 

The contentful plugin by default uses memoization to remember forever the last successful invocation for each combination of arguments, and use its result whenever Contentful fails. To disable this feature, use to `true` the new plugin option beloew

```
  /**
   * By default, the result of the last delivery invocation will be cached
   * forever and will only be used when a delivery call with the same arguments
   * fail.
   */
  disableFallbackCache?: boolean

```

## Testing

The pull request...

- has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/util/memoizer.ts
+++ b/packages/botonic-plugin-contentful/src/util/memoizer.ts
@@ -49,13 +49,12 @@ export class Memoizer {
     const cache: Cache<Return> = this.cacheFactory()
     const f = (...args: Args) =>
       this.strategy<Args, Return>(cache, this.normalizer, func, ...args)
-    // f.cache = cache
     return f as F
   }
 }
 
 /***
- * Only reinvoke if not in cache
+ * Only re-invoke if not in cache
  */
 export const cacheForeverStrategy: MemoizerStrategy = async <
   Args extends any[],
@@ -87,7 +86,6 @@ export function fallbackStrategy(
     func: (...args: Args) => Promise<Return>,
     ...args: Args
   ) => {
-    // return func(...args)
     const id = normalizer(...args)
     const oldVal = cache.get(id)
 
--- a/packages/botonic-plugin-contentful/src/util/memoizer.ts
+++ b/packages/botonic-plugin-contentful/src/util/memoizer.ts
@@ -1,0 +1,106 @@
+export class Cache<V> {
+  static readonly NOT_FOUND = Symbol('NOT_FOUND')
+  cache: Record<string, V> = {}
+  set(id: string, val: V): void {
+    this.cache[id] = val
+  }
+  get(id: string): V | typeof Cache.NOT_FOUND {
+    // Cache.has is only checked when undefined to avoid always searching twice in the object
+    const val = this.cache[id]
+    if (val === undefined && !this.has(id)) {
+      return Cache.NOT_FOUND
+    }
+    return val
+  }
+  has(id: string): boolean {
+    return id in this.cache
+  }
+}
+
+export type MemoizerNormalizer = (...args: any) => string
+
+export const jsonNormalizer: MemoizerNormalizer = (...args: any) => {
+  return JSON.stringify(args)
+}
+
+export type MemoizedFunction<Args extends any[], Return> = (
+  ...args: Args
+) => Promise<Return> //& { cache: Cache<Return> }
+
+export type MemoizerStrategy = <Args extends any[], Return>(
+  cache: Cache<Return>,
+  normalizer: typeof jsonNormalizer,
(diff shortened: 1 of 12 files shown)


packages/botonic-plugin-contentful/src/util/memoizer.ts
```diff
@@ -0,0 +1,106 @@
...
+  return val
+}
+
+/**
+ * Always invokes the function, but fallbacks to last invocation result if available
+ */
+export function fallbackStrategy(
+  usingFallback: (functName: string, args: any[], error: any) => Promise<void>
+): MemoizerStrategy {
+  return async <Args extends any[], Return>(
```
vanbasten17, line 84: do you no need to put again `extends any[]` in this type? I understand that Args will extend it as you have typed it as MemoizerStrategy, could be?

packages/botonic-plugin-contentful/src/util/memoizer.ts
```diff
@@ -0,0 +1,106 @@
...
+    cache: Cache<Return>,
+    normalizer = jsonNormalizer,
+    func: (...args: Args) => Promise<Return>,
+    ...args: Args
+  ) => {
+    // return func(...args)
```
vanbasten17, line 90: delete?

packages/botonic-plugin-contentful/src/util/memoizer.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  memoize<
+    Args extends any[],
+    Return,
+    F extends (...args: Args) => Promise<Return>
+  >(func: F): F {
+    const cache: Cache<Return> = this.cacheFactory()
+    const f = (...args: Args) =>
+      this.strategy<Args, Return>(cache, this.normalizer, func, ...args)
+    // f.cache = cache
```
vanbasten17, line 52: delete?

vanbasten17 (approved):

packages/botonic-plugin-contentful/src/util/memoizer.ts
dpinol: I typed fallbackStrategy as MemoizerStrategy for the compiler to verify that the lambda I retry complies with this type, but the lambda is not defined to be of this type (generics arguments must always be explicit, otherwise changing their names in the definition would break all usages)

packages/botonic-plugin-contentful/src/util/memoizer.ts
dpinol: thanks,  done in ee9efdb

packages/botonic-plugin-contentful/src/util/memoizer.ts
dpinol: thanks,  done in ee9efdb

dpinol:
https://user-images.githubusercontent.com/1954955/122032832-489be500-cdd0-11eb-840f-2a1f5f9829cd.mp4

packages/botonic-plugin-contentful/tests/util/memoizer.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    mock.salt = 'pepper'
+    await memoized(2, 'b')
+
+    mock.error = new Error('forced failure')
+    await memoized(2, 'b')
+
+    // act
+    mock.error = undefined
+    await expect(memoized(2, 'b')).resolves.toBe('A2bpepper')
+    expect(mock.callsFA).toBe(4)
```
asastre, line 94: Why `mock.callsFa` has to be called 4 times, wouldn't the last 2 times return the cached value instead of calling the `fA` function?

asastre (approved):
:clap: :clap: Just one doubt in the tests

packages/botonic-plugin-contentful/tests/util/memoizer.test.ts
dpinol: This cache is not a "typical cache", since it *always* perform the call. It only returns a cached value when the last invocation failed.

PR #1642, opened 2021-06-15
Contentful/cache size limit

dpinol:
Depends on #1639 . :warning:  **Please review it first**

## Description

#1639 created fallback cache decorator around Contentful API which remembers forever the last successful invocation for each combination of arguments, and use its result whenever Contentful fails. 
This PR adds a limit in the in memory cache size to avoid memory overflows.

## Context

In case any bot adds arguments with non repeated values, the cache could grow infinitely.

* When Cache reaches 50%, 75% or 90% size, a warning is reported to the logger configured in Contentful options
*  When Cache surpass the maximum size, existing cache items are deleted until the new item fits. 
* In the future, we might use an external library to have a LRU cache. Not critical because typically data in CMS is small (we're not caching media, only their URLs)

The contentful plugin by default uses memoization to remember forever the last successful invocation for each combination of arguments, and use its result whenever Contentful fails. To limit the amount of memory

```
```

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/util/cache.ts
+++ b/packages/botonic-plugin-contentful/src/util/cache.ts
@@ -98,7 +98,7 @@ export class LimitedCacheDecorator<V> implements Cache<V> {
       }
       if (!itFits) {
         this.logger(
-          `Cannot add object in cache because it's larger than max capacity(${this.limitKB})`
+          `Cannot add entry in cache because IT ALONE is larger than max capacity(${this.limitKB})`
         )
         return
       }
--- a/packages/botonic-plugin-contentful/src/util/cache.ts
+++ b/packages/botonic-plugin-contentful/src/util/cache.ts
@@ -1,0 +1,134 @@
+import { roughSizeOfObject } from './objects'
+
+export const NOT_FOUND_IN_CACHE = Symbol('NOT_FOUND')
+
+export interface Cache<V> {
+  set(id: string, val: V): void
+  get(id: string): V | typeof NOT_FOUND_IN_CACHE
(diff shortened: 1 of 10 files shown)


packages/botonic-plugin-contentful/src/util/cache.ts
```diff
@@ -0,0 +1,138 @@
...
+        }
+        this.cache.del(k)
+        itFits = checkFit()
+      }
+      if (!itFits) {
+        this.logger(
+          `Cannot add object in cache because it's larger than max capacity(${this.limitKB})`
+        )
+        return
+      }
```
asastre, line 104: Will this code ever be executed? 
In the previous loop, if adding the new key doesn't fit in the cache, it will keep removing cached values until it fits, so at the very end the cache may not have any value so it will be able to add new values because it will be empty, right?

asastre (approved):
Nice one!

packages/botonic-plugin-contentful/src/util/cache.ts
dpinol: since checkFit takes into account the incremental size of the new item, it's theoretically possible that the single item size is larger that the full cache capacity.
I improved the error message to be more descriptive
```
const checkFit = () =>
      this.cache.size() * 1024 + incBytes <= this.limitKB * 1024
    
```

asastre: Thanks for the clarification @dpinol!

ericmarcos (approved):

PR #1644, opened 2021-06-17
chore(ner/intent): support for botonic full-stack

vanbasten17:
## Description
* Support for nlp plugins in botonic full-stack project.

## Context
MODELS_BASE_URL will be defined when the lambda of the bot is deployed under an env variable `process.env.MODELS_BASE_URL` and defined by webpack when it's run in development mode under `MODELS_BASE_URL`.

### DIFF
--- a/packages/botonic-plugin-intent-classification/src/model/model-selector.ts
+++ b/packages/botonic-plugin-intent-classification/src/model/model-selector.ts
@@ -9,11 +9,13 @@ export class IntentModelSelector extends ModelSelector<
   IntentClassifier,
   IntentClassifierConfig
 > {
-  static async build(
-    locales: Locale[],
-    modelsBaseUrl: string
-  ): Promise<IntentModelSelector> {
-    const selector = new IntentModelSelector(locales, modelsBaseUrl)
+  static async build(locales: Locale[]): Promise<IntentModelSelector> {
+    const baseUrl =
+      // @ts-ignore
+      (typeof MODELS_BASE_URL !== 'undefined' && MODELS_BASE_URL) ||
+      process.env.MODELS_BASE_URL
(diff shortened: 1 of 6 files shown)


packages/botonic-plugin-intent-classification/src/model/model-selector.ts
```diff
@@ -9,11 +9,13 @@ export class IntentModelSelector extends ModelSelector<
...
-  static async build(
-    locales: Locale[],
-    modelsBaseUrl: string
-  ): Promise<IntentModelSelector> {
-    const selector = new IntentModelSelector(locales, modelsBaseUrl)
+  static async build(locales: Locale[]): Promise<IntentModelSelector> {
+    const baseUrl: string =
+      // @ts-ignore
+      (typeof MODELS_BASE_URL !== 'undefined' && MODELS_BASE_URL) ||
+      process.env.MODELS_BASE_URL
```
ericmarcos, line 16: why not just use `const baseUrl: string = process.env.MODELS_BASE_URL` (simpler, cleaner code) and in our local configuration use the EnvironmentPlugin:
```
new webpack.EnvironmentPlugin({
  MODELS_BASE_URL: '...',
});

```

ericmarcos (changes requested):

packages/botonic-plugin-intent-classification/src/model/model-selector.ts
vanbasten17: definitely is not working in this way. We have tried.
```
new webpack.EnvironmentPlugin({
  MODELS_BASE_URL: '...',
});
```
and also
```
new webpack.DefinePlugin({
  'process.env.MODELS_BASE_URL': '...',
});
```
we should think carelessly what should be the best way to handle this kind of variables with newer webpack versions

PR #1646, opened 2021-06-17
Full stack/chore(core): add data models for botonic 1.0

vanbasten17:
## Description
* Adding data models proposal into `@botonic/core` (note that some definitions may be susceptible to be changed) without conflicting with current `@botonic/core` functionality.

## Context
* Defining data models for botonic full stack.

* We defined the most simple data models for botonic full-stack. Each model it's being composed by extending from more basics interfaces (MessageEvent, ConnectionEvent, ...) and defining attributes more specifically in the very bottom interfaces (TextMessageEvent, ConnectionEvent, ...)  For these models it's important to:
1. Be able to have all the event details in separated fields to have the data as plain as possible in the database (will be useful for analytics concerns).
2. Be able to have dynamic resolution of typings.

To achieve both objectives, the purposed model is the following:
Screenshot 2021-06-17 at 13 40 30

A specific instance of a `TextMessageEvent` for example would contain all this information, that will be stored in the database:
Screenshot 2021-06-17 at 13 53 09

With this, we would be able to store the data as plain as possible in a DynamoDB table (with single table design approach, which will allow to save infrastructure costs): 
Screenshot 2021-06-11 at 18 03 18

And having dynamic type resolution:
```
async saveEvent(event: BotonicEvent): Promise<BotonicEvent> {
    if (event.eventType === 'message') {
      if (event.type === 'text') {
        await this.textMessageEventEntity.put(event)
      }
    }
    ...
}
```

**Note**: `MessageEvent` is named as `BotonicMessageEvent` in the source code due to conflicts with typescript dom definitions.

Scheme: 
[models.zip](https://github.com/hubtype/botonic/files/6670234/models.zip)

### DIFF
--- a/packages/botonic-core/src/models/events/index.ts
+++ b/packages/botonic-core/src/models/events/index.ts
@@ -19,6 +19,7 @@ export enum EventTypes {
 
 export interface BaseEvent {
   id: string
+  userId: string
   eventType: typeof EventTypes[keyof typeof EventTypes]
   createdAt: string
   modifiedAt?: string
--- a/packages/botonic-core/src/models/events/index.ts
+++ b/packages/botonic-core/src/models/events/index.ts
@@ -1,20 +1,20 @@
-import { TextMessageEvent } from './message/text'
-import { MediaMessageEvent } from './message/media'
-import { LocationMessageEvent } from './message/location'
+import { ConnectionEvent } from './connections/index'
 import { CarouselMessageEvent } from './message/carousel'
 import { CustomMessageEvent } from './message/custom'
-import { ConnectionEvent } from './connections/index'
+import { LocationMessageEvent } from './message/location'
+import { MediaMessageEvent } from './message/media'
+import { TextMessageEvent } from './message/text'
 
 export enum EventTypes {
-  connection = 'connection',
-  message = 'message',
-  ack = 'ack',
-  track = 'track',
+  CONNECTION = 'connection',
+  MESSAGE = 'message',
+  ACK = 'ack',
+  TRACK = 'track',
 }
 
-export interface BaseEvent<EventType extends EventTypes> {
+export interface BaseEvent {
   id: string
-  eventType: EventType
+  eventType: typeof EventTypes[keyof typeof EventTypes]
   createdAt: string
   modifiedAt?: string
 }
--- a/packages/botonic-core/src/models/events/index.ts
+++ b/packages/botonic-core/src/models/events/index.ts
@@ -1,0 +1,28 @@
+import { TextMessageEvent } from './message/text'
+import { MediaMessageEvent } from './message/media'
+import { LocationMessageEvent } from './message/location'
+import { CarouselMessageEvent } from './message/carousel'
+import { CustomMessageEvent } from './message/custom'
+import { ConnectionEvent } from './connections/index'
+
+export enum EventTypes {
+  connection = 'connection',
+  message = 'message',
+  ack = 'ack',
+  track = 'track',
+}
+
+export interface BaseEvent<EventType extends EventTypes> {
+  id: string
+  eventType: EventType
+  createdAt: string
+  modifiedAt?: string
+}
+
+export type BotonicEvent =
+  | TextMessageEvent
+  | MediaMessageEvent
+  | LocationMessageEvent
+  | CarouselMessageEvent
+  | CustomMessageEvent
+  | ConnectionEvent
--- a/packages/botonic-core/src/models/events/index.ts
+++ b/packages/botonic-core/src/models/events/index.ts
@@ -20,7 +20,7 @@ export enum EventTypes {
 export interface BaseEvent {
   id: string
   userId: string
-  eventType: typeof EventTypes[keyof typeof EventTypes]
+  eventType: EventTypes
   createdAt: string
   modifiedAt?: string
 }
--- a/packages/botonic-core/src/models/events/index.ts
+++ b/packages/botonic-core/src/models/events/index.ts
@@ -2,7 +2,12 @@ import { ConnectionEvent } from './connections/index'
 import { CarouselMessageEvent } from './message/carousel'
 import { CustomMessageEvent } from './message/custom'
 import { LocationMessageEvent } from './message/location'
-import { MediaMessageEvent } from './message/media'
+import {
+  AudioMessageEvent,
+  DocumentMessageEvent,
+  ImageMessageEvent,
+  VideoMessageEvent,
+} from './message/media'
 import { TextMessageEvent } from './message/text'
 
 export enum EventTypes {
@@ -21,7 +26,10 @@ export interface BaseEvent {
 
 export type BotonicEvent =
   | TextMessageEvent
-  | MediaMessageEvent
+  | AudioMessageEvent
+  | DocumentMessageEvent
+  | ImageMessageEvent
+  | VideoMessageEvent
   | LocationMessageEvent
   | CarouselMessageEvent
   | CustomMessageEvent
--- a/packages/botonic-core/src/models/events/message/buttons.ts
+++ b/packages/botonic-core/src/models/events/message/buttons.ts
@@ -1,0 +1,23 @@
+export interface BaseButton {
+  title: string
+}
+
+export interface WebviewButton extends BaseButton {
+  webview: any
+  params?: any
+}
+
+export interface UrlButton extends BaseButton {
+  url: string
+  target?: string
+}
+
+export interface PayloadButton extends BaseButton {
+  payload: string
+}
+
+export type Button = PayloadButton | UrlButton | WebviewButton
+
+export interface WithButtons {
+  buttons?: Button[]
+}
--- a/packages/botonic-core/src/models/events/message/carousel.ts
+++ b/packages/botonic-core/src/models/events/message/carousel.ts
@@ -1,8 +1,7 @@
-import { BotonicMessageEvent, MessageEventType } from '.'
+import { BotonicMessageEvent, MessageEventTypes } from '.'
 import { CarouselElement } from './element'
 
-export interface CarouselMessageEvent
-  extends BotonicMessageEvent<MessageEventType.carousel> {
-  type: MessageEventType.carousel
+export interface CarouselMessageEvent extends BotonicMessageEvent {
+  type: MessageEventTypes.CAROUSEL
   elements: CarouselElement[]
 }
--- a/packages/botonic-core/src/models/events/message/carousel.ts
+++ b/packages/botonic-core/src/models/events/message/carousel.ts
@@ -1,0 +1,8 @@
+import { BotonicMessageEvent, MessageEventType } from '.'
+import { CarouselElement } from './element'
+
+export interface CarouselMessageEvent
+  extends BotonicMessageEvent<MessageEventType.carousel> {
+  type: MessageEventType.carousel
+  elements: CarouselElement[]
+}
--- a/packages/botonic-core/src/models/events/message/element.ts
+++ b/packages/botonic-core/src/models/events/message/element.ts
@@ -1,7 +1,7 @@
 import { WithButtons } from './buttons'
 
 export interface CarouselElement extends WithButtons {
-  src: string
+  pic: string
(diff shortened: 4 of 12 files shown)


packages/botonic-core/src/models/events/message/element.ts
```diff
@@ -0,0 +1,7 @@
+import { WithButtons } from './buttons'
+
+export interface CarouselElement extends WithButtons {
+  src: string
+  title: string
+  subtitle?: string
+}
```
asastre, line 7: Is the image missing here, right?

packages/botonic-core/src/models/events/message/carousel.ts
```diff
@@ -0,0 +1,8 @@
+import { BotonicMessageEvent, MessageEventType } from '.'
+import { CarouselElement } from './element'
+
+export interface CarouselMessageEvent
+  extends BotonicMessageEvent<MessageEventType.carousel> {
+  type: MessageEventType.carousel
```
asastre, line 6: Since `type` is already defined in `BotonicMessageEvent` is not needed to be redefined here, right?
This is happening also with `CustomMessageEvent`.

packages/botonic-core/src/models/events/message/media.ts
```diff
@@ -0,0 +1,32 @@
...
+  extends BotonicMessageEvent<MessageEventType.video>,
+    WithButtons {
+  src: string
+}
+
+export type MediaMessageEvent =
+  | AudioMessageEvent
+  | ImageMessageEvent
+  | DocumentMessageEvent
+  | VideoMessageEvent
```
asastre, line 32: Since all media events have the `src` attribute, would it make sense to define `MediaMessageEvent` as:
```
export interface MediaMessageEvent extends WithButtons {
  src: string
}
```
And then declare the specific media events like this?:
```
export interface VideoMessageEvent extends MediaMessageEvent, BotonicMessageEvent<MessageEventType.video> {}
```

packages/botonic-core/src/models/events/message/buttons.ts
```diff
@@ -0,0 +1,23 @@
+export interface BaseButton {
+  title: string
```
asastre, line 2: In my opinion naming this attribute `title` is a little bit confusing because it makes me wonder if this is just a title (without being displayed in the bot, just as a description) for the button or is the actual text that will be displayed in the button.
The same thing is happening for the `Reply` interface.

If this attribute is referring to the text displayed in the button, I would consider renaming it to `text`, what do you think?

asastre (commented):
Nice approach and segmentation of interfaces!

I've added some comments about the implementation and here I'm adding some more observations.

In the diagram showed in the description, there's a relation between `User` and `BaseEvent`, but I can't see any field in `BaseEvent` related to the `User` (I was expecting to see a userId attribute) and in the screenshot shared from the DynamoDB stored data I can see User fields like `isOnline` or `route` but not the `id` field.
How this relation will be handled?

packages/botonic-core/src/models/events/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  ack = 'ack',
+  track = 'track',
+}
+
+export interface BaseEvent<EventType extends EventTypes> {
+  id: string
+  eventType: EventType
+  createdAt: string
+  modifiedAt?: string
+}
```
ericmarcos, line 20: I'm not sure the `<EventType extends EventTypes>` generic is the right approach here. The attribute `eventType` should be typed as one of the pre-defined strings in the enum, not an enum in itself.

I propose this:
```typescript
export enum EventTypes {
  CONNECTION = 'connection',
  MESSAGE = 'message',
  ACK = 'ack',
  TRACK = 'track',
}

// The following line is equivalent to `type EventType = EventTypes.CONNECTION | EventType.MESSAGE | ... `
//https://stackoverflow.com/questions/55142177/how-to-build-a-type-from-enum-values-in-typescript
export type EventType = Record<EventTypes, string>

export interface BaseEvent {
  id: string
  eventType: EventType
  createdAt: string
  modifiedAt?: string
}
```

Then you would just override the attribute with the right value in subclasses (and also use the same approach for its "enum" attributes):
```typescript
export enum ConnectionEventStatuses {
  CONNECTED = 'connected',
  DISCONNECTED = 'disconnected',
}

export type ConnectionEventStatus = Record<ConnectionEventStatuses, string>

export interface ConnectionEvent extends BaseEvent {
  eventType: EventTypes.CONNECTION
  status: ConnectionEventStatus
}
```

extra: notice how with this approach we can now name the Enum attributes with all caps (which is desirable with constants). We were doing lowercase because a previous approach using `keyof typeof EventTypes` required so if we wanted lowercase values in the database.

packages/botonic-core/src/models/events/message/buttons.ts
ericmarcos: I think `text` is too generic, we want a name that is specific enough to describe "the text that is displayed to the user", for example: `label`, `textLabel`, `display`, `displayText`... However, I still think `title` is good enough. It is used by [Facebook Messenger's API](https://developers.facebook.com/docs/messenger-platform/send-messages/buttons/) (that's where we inherit this name from).

ericmarcos (changes requested):

packages/botonic-core/src/models/events/message/buttons.ts
asastre: You're right, `text` may be too generic, I would go with `label` or `textLabel` then. 
Even if Facebook uses it I think it's not the best naming for this variable, for me `title` represents only the description of the button, not something to be displayed within the Button.

vanbasten17: leaving it as `title` atm, in the following iterations (integration of parser), I would consider if we change the name to `label`, `textLabel` or similar.

packages/botonic-core/src/models/events/index.ts
vanbasten17: by trying so:
```
export enum EventTypes {
  CONNECTION = 'connection',
  MESSAGE = 'message',
  ACK = 'ack',
  TRACK = 'track',
}

export type EventType = Record<EventTypes, string>

export interface BaseEvent {
  id: string
  eventType: EventType
  createdAt: string
  modifiedAt?: string
}

export enum ConnectionEventStatuses {
  CONNECTED = 'connected',
  DISCONNECTED = 'disconnected',
}

export type ConnectionEventStatus = Record<ConnectionEventStatuses, string>

export interface ConnectionEvent extends BaseEvent {
  eventType: EventTypes.CONNECTION
  status: ConnectionEventStatus
}
```
I get:
```
Interface 'ConnectionEvent' incorrectly extends interface 'BaseEvent'.
  Types of property 'eventType' are incompatible.
    Type 'import("/Users/mrabat/Documents/hubtype/experiments/2021-04-12-botonic-full-stack-poc/@botonic1.0/core/models/events/connections/index").EventTypes' is not assignable to type 'EventType'.
```
I don't really get why TS is not capable of understanding this.

vanbasten17: okay, I found this to be of interest: https://stackoverflow.com/a/61375519
when we type with `keyof typeof EnumName`:
**e.g:**
```
enum ConnectionEventStatuses {
  CONNECTED = 'connected',
  DISCONNECTED = 'disconnected',
}

export interface ConnectionEvent extends BaseEvent {
  eventType: EventTypes.CONNECTION
  status: keyof typeof ConnectionEventStatuses
}

let c: ConnectionEvent = {
  status: ConnectionEventStatuses.CONNECTED
}
```
we end up with this complaint:
```
Type 'ConnectionEventStatuses.CONNECTED' is not assignable to type '"CONNECTED" | "DISCONNECTED"'.ts(2322)
index.ts(35, 3): The expected type comes from property 'status' which is declared here on type 'ConnectionEvent'
(property) ConnectionEvent.status: "CONNECTED" | "DISCONNECTED"
```
this is because `ConnectionEventStatuses.CONNECTED` will resolve to the value `connected` of the enum instead of the key, that is what we want.
So typing it like this will do the trick:
```
export interface ConnectionEvent extends BaseEvent {
  eventType: EventTypes.CONNECTION
  status: typeof ConnectionEventStatuses[keyof typeof ConnectionEventStatuses]
}
A little redundant in my opinion, but it won't be declared so much times.
```
Then we can make use of the benefits of using enums. I have read some posts regarding this and most of them suggest that it's better to use union types instead of enums. What I don't like about using union types is that we lose the grouping related values within the same symbol and that the editor autocomplete fills it up with string literals.
The alternative with union types would look like this:
**e.g:**
```
type ConnectionEventStatuses = 'connected' | 'disconnected' 
interface ConnectionEvent extends BaseEvent {
  eventType: EventTypes.CONNECTION
  status: ConnectionEventStatuses
}
let c: ConnectionEvent = {
  status: 'connected' // here we are forced to use the string literal
}
```
To work around this, we could define a type for each string literal (type ConnectedStatus = 'connected', ...), but I think it will be very redundant, so I would go with the proposed solution.

packages/botonic-core/src/models/events/message/element.ts
vanbasten17: well, I did rename it as `src` instead of `pic`, here the confusion maybe, as this is how I named it in the parser. Maybe `pic` is more explicit given the context . Renaming it

packages/botonic-core/src/models/events/message/carousel.ts
vanbasten17: yes, that was right before the last commit. With the new approach we need to explicitly add it in order to let typescript tell between the different "BotonicEvents".

packages/botonic-core/src/models/events/message/media.ts
vanbasten17: applied differently! let me know when you take it a look 👍

asastre (approved):
:rocket:

packages/botonic-core/src/models/events/index.ts
ericmarcos: My bad, I didn't understand well how Enum typing worked. I think the easiest possible solution is similar to the original proposal, but dropping the generic, just:
```ts
enum EventTypes {
  CONNECTION = 'connection',
  MESSAGE = 'message',
  ACK = 'ack',
  TRACK = 'track',
}

interface BaseEvent {
  id: string
  eventType: EventTypes
  createdAt: string
  modifiedAt?: string
}
```

Check [this typescript playground code](https://www.typescriptlang.org/play?#code/KYOwrgtgBAogbqALgFQJ4AdgGcoG8BQUUAwgPIBy5MxyAkhVALxQDkAxgPYgjBuICWXFgBpCUALIwAylICCAcRhNWEbFgCGAc2Aixs4gGllLdWwDWuosgBK+o8xaIATqYuiAvvnwB6bwAtERHQsAC5fLERXDgQnADMAGw4AdwA6TghvAEcwbAEuLG8AVkKARgAWACYSgHZq-2SAWkQOBoAjMH54gBMG9SaMYAbYpw4IBtBIBrh1eJysBv4QfswsNid+dEQfbyhkP2AoWI54xKTFzSh4xYOujmwQRygkjiczKHUcYAAPTD5gLvwiAGsAQIBQwOY1l4Ly6AB54Eg0CthFAIusQJoAHxeXwIsFI4AVJiMZh48ErQHAskEomk0Hk7ApMiUah0BgAHxBiIGWBSkhkCiUnOpPJSdigwvpBN5NjsXkWiGAcVMBwAQh9gGS8GJ+F0QqjnOcxMApQN9SKKUQ1sB1IqurJEPq0UaiBAOF1+LF+P8HQB+J2GjH4Tz4TggCLvELqrCa+nKAhEIi6-UAIhKFQAzCnRImoCbuZhzaaVkyKFQaPRyDnE9bbT7HVAUyng6H8ogoK0oxqtcwE4nk4301nq0R8-izY2wzw+IIQCn3jgLdgR1Ba3aHanmyGw1hjsAUolNAAKdQo1oASiAA).

vanbasten17: changed 👍

ericmarcos (approved):

PR #1651, opened 2021-06-18
feat(core): add optional timeout to getOpenQueues & storeCaseRating

dpinol:
## Description

Added timeout argument to 2 of the API calls to backend. If approach is approved, it should be added to the rest of API calls

## Context

* In case a call to backend hangs for many seconds, timeout ensures that the bot is able to react (eg. providing an error message or retrying) before the AWS lambda timeout aborts the bot invocation.

Added in a context argument (like in contentful plugin) so that more options can be added in the future

TODO

## Testing

The pull request...

- has unit tests
- has integration tests
- doesn't need tests because it's a call to backend

### DIFF
--- a/packages/botonic-core/src/handoff.js
+++ b/packages/botonic-core/src/handoff.js
@@ -2,9 +2,17 @@ import axios from 'axios'
 
 const HUBTYPE_API_URL = 'https://api.hubtype.com'
 
-export async function getOpenQueues(session) {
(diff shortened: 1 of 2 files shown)


asastre (approved):
As a future improvement, would it make sense to decrease the default timeout and add an extra logic to retry the request X times (configurable but at least do 1 retry if it fails) @vanbasten17 @ericmarcos?

PR #1671, opened 2021-06-30
feat(react): add new <Raw> component

ericmarcos:
## Description

A new `<Raw>` component that allows developers to send arbitrary data payloads to messaging APIs. It accepts 2 props:
* `data`: A JSON object containing the data to be sent directly to whatever messaging API.
* `alt`: A human friendly string (alternative text) to be displayed in chat history.

## Context

In order to provide the best user experience, developers need to take full advantage of all the features provided by messaging apps. Often times, these features are very specific of a given channel and are not standarizable (like `<WhatsappTemplate>`), making it difficult for Botonic to implement well. Other times, some features are so new or so niche that they are not supported in Botonic just yet.

This component allows developers to directly send JSON payloads to the connected messaging APIs, thus having access to all features without having to wait for official support.

This is an example of sending a Facebook Messenger [product template](https://developers.facebook.com/docs/messenger-platform/send-messages/template/product):

```jsx
<Raw alt="Alternative text" data={{
    "attachment":{
      "type":"template",
        "payload": {
          "template_type": "product",
          "elements": [
            {
              "id": "123"
            },
            {
              "id": "456"
            }
         ]
      }
    }
}} />
```

This component makes no sense on webchat, so we assume it will only be visible on webchat during development ("botonic serve"). That's why it will be displayed like this:

Screenshot 2021-06-30 at 10 33 42
Screenshot 2021-06-30 at 10 33 56

This is how this component will look like on Facebook Messenger:
Screenshot 2021-06-30 at 10 33 56

Channels that don't support that data payload will return an error and won't display anything to the end user.

## Testing

The pull request...

- has unit tests
- has integration tests


vanbasten17 (approved):
Nice!! 👏

asastre (approved):
LGTM!

PR #1700, opened 2021-07-07
Full stack/feature-branch-botonic1.0

vanbasten17:
**Depends on**: https://github.com/hubtype/botonic/pull/1646
## Description
* Added BotonicOutputParser to append `parsedResponse` (JSON with parsed messages instead of our internal xml, which is pretty unusable) to input object.
* Added tests for this parser.
* Modify project config to allow ts-js code and configure tests.
* Bump & publish 1.0.0-alpha.0 version to start iterating.

## Context

BotonicOutputParser have two main methods:
```
xmlToMessageEvents(html: string): Partial<BotonicEvent>[] // parses bot xml responses into an array of BotonicEvent, that will be commited later to the database depending on the environment (local/aws).
parseFromUserInput(input: any): Partial<BotonicEvent> // parses the user input into a BotonicEvent, with the same aim as the previous method.
```
**Note**: Still some things are expected to change. ATM, we are typing it with Partial because some information cannot be accessed at runtime in this part of code but later, so `userId, eventId, from, ack, ...` will be fulfilled from a component on top of its usage.

## Approach
* We have created a factory that given the resolved MessageType, execute the corresponding parsing strategy. First of all, it parses the common properties for all kind of messages, and then appends to it the specific properties for every message.
## Testing

The pull request...
- has unit tests


ericmarcos (approved):
Niceee! Gogogo!

PR #1701, opened 2021-07-07
Full stack/react experimental components

vanbasten17:
**Depends on**: https://github.com/hubtype/botonic/pull/1700

## Description
* Adding `experimental` folder with necessary changes in botonic components for correct functioning with the new parser in @botonic/core (https://github.com/hubtype/botonic/pull/1700) and avoid breaking current @botonic/react package. Every component used in botonic 1.0 project, will be imported from @botonic/react/src/experimental.

Some things have been fixed or simplified in this process:
* Simpler `msgToBotonic` function (no more weird accesses to data attributes, e.g.: now text of a `Text` component comes under the `text` property, not under text.text or text.data)
* Now `Location` component receives delay and typing props as expected. `lat` and `long` are also recovered from localStorage now (was not working).
* Now `Audio, Image, Document, Video` allow buttons within the same component.
* New `WebchatDev` component with Botonic Playground.
* Markdown now works properly with \ for double lines


ericmarcos (approved):
🚀🚀🚀

PR #1702, opened 2021-07-07
Full stack/chore(api): add api package for botonic1.0

vanbasten17:
## Description
* Added first alpha version of @botonic/api package to handle common wrappers for local/aws infrastructures


vanbasten17:
@ericmarcos , here he will have the same problem as with @botonic/core package, as it depends on webpack variables, atm, pulling code directly from src when using it.

ericmarcos (approved):

PR #1710, opened 2021-07-12
Full stack/api local json db

asastre:
## Description
Use of [node-json-db](https://www.npmjs.com/package/node-json-db) package to store data locally (users, events and connections) when no data provider is set.

Also I've done these improvements:
- Fixed some missing types and functions (`getUsers` and `getEvents)` in `DataProvider` interface and its implementations.
- Refactored `dataProviderFactory` to always return a `DataProvider`.
- Bump version of @botonic/api to `1.0.0-alpha.1`

## Context
We were facing an issue while storing data locally in memory because rest and websocket were storing their data independently and we need to have it together. 

## Testing

### DIFF
--- a/packages/botonic-api/src/data-provider/index.ts
+++ b/packages/botonic-api/src/data-provider/index.ts
@@ -6,6 +6,7 @@ import { LocalDevDataProvider } from './local-data-provider'
 
 export const dataProviderProtocols = {
   DYNAMO_DB: 'dynamodb',
+  FILE: 'file',
 }
 
 export interface DataProvider {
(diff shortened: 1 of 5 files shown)


packages/botonic-api/src/data-provider/index.ts
```diff
@@ -4,24 +4,41 @@ import { User } from '@botonic/core/src/models/user'
...
-    return new DynamoDBDataProvider(url)
+    return localDataProvider ?? new LocalDevDataProvider()
+  }
+
+  const protocol = url.split('://')[0]
+  switch (protocol) {
+    case dataProviderProtocols.DYNAMO_DB:
+      return new DynamoDBDataProvider(url)
+    default:
+      return localDataProvider ?? new LocalDevDataProvider()
```
ericmarcos, line 42: Good refactor! Now that we're at it, maybe it makes sense to add a "local database URL" to let developers customize the default local DB. For instance `file://path/to/db.json`. We can also remove the global variable `const localDataProvider` as we needed it to avoid duplicated local DBs, something that it shouldn't be a problem anymore with data persisted to disk.

ericmarcos (approved):
Nice work! Let's add a task to create unit tests for this.

packages/botonic-api/src/data-provider/index.ts
asastre: Totally  agree. Done!

I'm also reloading the JSON file before every action made to make sure the DB is always up to date.

PR #1721, opened 2021-07-14
feat(cli): add new serve command for Botonic 1.0 under flag --preview

ericmarcos:
## Description

Updated `botonic serve` command for Botonic 1.0. The new functionality can be activated with the flag `--preview` or `-p`:

```
botonic serve --preview
```

It launches 3 servers concurrently:
* Webchat (frontend React app)
* API REST (backend Node/Express app)
* Websocket Server

In the future it will launch another server for the webviews.

One of the main features is the linking of the local dev server (API) directly with messaging apps like Whatsapp or Telegram. In order to do that, we generate a [local tunnel](https://github.com/localtunnel/localtunnel) and send its public URL to the new Botonic Playground API, which manages the direct connection with messaging apps.

## Context

With this new command, developers can now try their local botonic projects on Whatsapp and other platforms without the hassle of creating any accounts. This allows them to learn Botonic much faster and be more productive when building their projects.

`botonic serve --preview`

## Testing

The pull request...

- doesn't need tests because it's still an experimental feature. Still a lot of stuff missing and subject to changes.


packages/botonic-cli/src/botonic-playground-service.ts
```diff
@@ -0,0 +1,98 @@
...
+  get code(): string | undefined {
+    return this.playgroundSession?.code
+  }
+
+  async stop(): Promise<void> {
+    try {
+      this.tunnel && (await this.tunnel.close())
+      this.playgroundSession?.id &&
+        (await axios.delete(
+          `${PLAYGROUND_API_HOST}/playground-sessions/${this.playgroundSession.id}`
```
asastre, line 83: Again, use of `this.playgroundSessionsApi` instead?

packages/botonic-cli/src/botonic-playground-service.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  getTunnelSubdomain(url: string): string | null {
+    try {
+      const r = /https:\/\/(.*)\.loca\.lt/
+      return r.exec(url)![1]
+    } catch (e) {
+      return null
+    }
+  }
+}
```
asastre, line 98: What's the purpose of this method?

packages/botonic-cli/src/commands/serve.ts
```diff
@@ -11,20 +16,120 @@ export default class Run extends Command {
...
+      } catch (e) {
+        console.log("Couldn't start the Playground service: ", e)
+      }
+
+      if (apps.includes('websocket')) {
+        try {
+          await kill(API_WS_PORT, 'tcp')
+        } catch (e) {
+          console.error(
+            `Error whilst shutting down "api websocket" port: ${colors.red(
```
asastre, line 56: I think it would be nice to add the port number in the error message.

packages/botonic-cli/src/commands/serve.ts
```diff
@@ -11,20 +16,120 @@ export default class Run extends Command {
...
+            .filter(job => job && job.runWhen()),
+          {
+            //prefix: '{name} |',
+            timestampFormat: 'HH:mm:ss',
+          }
+        )
+      } catch (e) {
+        console.log(e)
+      }
+      await pgs.stop()
```
asastre, line 121: What if `pgs.start()` call failed? Would this call fail and Botonic wouldn't be served?
If this is true, maybe this `pgs.stop()` call should be done only if `pgs.start()` was successful or within a `try/catch`.

asastre (changes requested):
Nice to have this preview of Botonic 1.0!! :partying_face: 
Some things to be reviewed/changed though :wink:

packages/botonic-cli/src/botonic-playground-service.ts
ericmarcos: Ah, true, I did a big refactor and forgot to remove this function, thanks!

packages/botonic-cli/src/commands/serve.ts
ericmarcos: If it failed then this call won't do anything

packages/botonic-cli/src/commands/serve.ts
```diff
@@ -11,20 +16,122 @@ export default class Run extends Command {
...
+      const API_WS_PORT = 9091
+      const WEBCHAT_DIR_SRC = path.resolve(BASE_PATH, './webchat')
+      const WEBCHAT_PORT = 9000
+
+      const pgs = new PlaygroundService(API_REST_PORT)
+      try {
+        await pgs.start()
+      } catch (e) {
+        console.log("Couldn't start the Playground service: ", e)
+      }
```
asastre, line 49: There's no need to add here the `try/catch` since the `start` method has already one inside and does not rethrow the exception if it fails.

asastre (approved):
Just one last comment, the rest LGTM! :rocket:

vanbasten17 (approved):

PR #1722, opened 2021-07-15
feat(api): Implementation of events endpoints

asastre:
## Description
- Implementation of events routes for CRUD operations. No authentication is done so far.
- Added validations to the parameters received in the requests using `express-validator`.
- Added some missing methods in the `DataProvider` interface and its implementations (`deleteEvent` and `updateEvent`).
- Refactor of `LocalDevDataProvider` removing all `try/catch` to avoid failing silently.

## Context
The api endpoints of events and users were not implemented yet.

All the data validation has been done based on the events type declarations.
For the `patch` request, in order to take advantage of all the validantions done, a method has been implemented to transform all fields to optional fields but still keep all the other validations done.

## Testing


ericmarcos (changes requested):

packages/botonic-api/src/rest/routes/events.ts
```diff
@@ -1,38 +1,178 @@
...
+
+      const dp = dataProviderFactory(process.env.DATA_PROVIDER_URL)
+      const event = await dp.getEvent(params.eventId)
+      if (!event) {
+        res
+          .status(404)
+          .send({ error: `Event with ID '${params.eventId}' not found` })
+        return
+      }
+      const updatedEvent = _.merge(event, newEventData)
```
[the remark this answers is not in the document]
asastre: You're right, I think there's no need to do a deep merge here, using `Object.assign` instead and removing `lodash` dependency.

packages/botonic-api/src/rest/routes/validation/events.ts
```diff
@@ -0,0 +1,261 @@
...
+}
+
+export const botonicMessageEventSchema: Schema = {
+  ...baseEventSchema,
+  ack: { ...inBody, ...isRequired, ...isIn(Object.values(MessageEventAck)) },
+  from: { ...inBody, ...isRequired, ...isIn(Object.values(MessageEventFrom)) },
+  type: { ...inBody, ...isRequired, ...isIn(Object.values(MessageEventTypes)) },
+  typing: { ...inBody, ...isRequired, ...isNaturalNumber, ...toInt },
+  delay: { ...inBody, ...isRequired, ...isNaturalNumber, ...toInt },
+}
```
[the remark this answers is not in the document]
asastre: Yes, I thought the same, we'll be having duplicated code by having all the type definitions and this validations, but I haven't seen any easy/automatic way of taking advantage of the type definitions to generate the validations.

I've tried (not much) this approach: [https://github.com/ForbesLindesay/typescript-json-validator](https://github.com/ForbesLindesay/typescript-json-validator) but it seems it's not actively maintained and I couldn't generate the file with the validation, it crashed.

I've seen this other approach: [https://github.com/kimamula/ts-transformer-keys](https://github.com/kimamula/ts-transformer-keys), but I think it's too complicated and I don't think it will solve our problem by having only the keys of an interface...

Another approach would be to have all events/users as classes and add a validation method to it and execute this method when creating/updating one, but I think this way will change too much things right now.

packages/botonic-api/src/rest/routes/events.ts
```diff
@@ -1,38 +1,177 @@
...
+
+      const dp = dataProviderFactory(process.env.DATA_PROVIDER_URL)
+      const event = await dp.getEvent(params.eventId)
+      if (!event) {
+        res
+          .status(404)
+          .send({ error: `Event with ID '${params.eventId}' not found` })
+        return
+      }
+      const updatedEvent = Object.assign(event, newEventData)
```
ericmarcos, line 147: Extra: `const updatedEvent = {...event, ...newEventData}` (I think it's more readable this way and Object.assign modifies the first param which can cause undesired side effects)

packages/botonic-api/src/rest/routes/validation/events.ts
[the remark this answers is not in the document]
ericmarcos: ok, let's leave it like this and we'll research alternatives in the future

packages/botonic-api/src/rest/routes/events.ts
asastre: Yes, I did it this way in the first place but I got a TS error that made me add an `as BotonicEvent` to make sure `updateEvent` has the right type and since `event` can be modified (without making any side effects) I decided to use `Object.assign`.

But I agree with you that it's more readable with deconstruction, changed!

packages/botonic-api/src/data-provider/local-data-provider.ts
```diff
@@ -22,143 +22,117 @@ export class LocalDevDataProvider implements DataProvider {
...
-    } catch (e) {
-      console.error(
-        `Error saving event with user ID '${event.userId}' to local DB`,
-        e
-      )
+    this.db.reload()
+    if (!this.db.exists(path)) {
+      this.db.push(path, event)
+    }
+    return event
```
marcpdw, line 82: Probably not for this PR, but not sure whether we could have commong `update`/`delete`/ ... methods that follow this common pattern. This way, you only need to pass the path and the event to a common method to do the rest.

packages/botonic-api/src/rest/routes/events.ts
```diff
@@ -1,38 +1,177 @@
...
+      try {
+        const errors = validationResult(req)
+        if (!errors.isEmpty()) {
+          res.status(400).send({ errors: errors.array() })
+          return
+        }
+
+        const query = matchedData(req, { locations: ['query'] })
+        const limit = query.limit ?? undefined
+        const offset = query.offset ?? undefined
```
marcpdw, line 32: Question here because I'm not sure: isn't this redundant? When `query.limit` is not defined, it will return `undefined`, right?

packages/botonic-api/src/rest/routes/validation/events.ts
```diff
@@ -0,0 +1,263 @@
...
+export const eventIdParamSchema: ParamSchema = { ...inParams, ...isRequired }
+
+export const buttonsSchema: Schema = {
+  buttons: {
+    ...inBody,
+    ...isOptional,
+    isArray: true,
+    errorMessage: 'buttons have to be an array',
+  },
+  'buttons.*.title': { ...inBody, ...isRequired },
```
marcpdw, line 63: Question, does `*` have a special effect here? Or you just want the key to have that value? It looks like this is the value to me.

marcpdw (commented):
I have a few questions before finishing the review

packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
```diff
@@ -92,6 +92,12 @@ export class DynamoDBDataProvider implements DataProvider {
     return event
   }
 
+  // @ts-ignore
+  async updateEvent(event: BotonicEvent): Promise<BotonicEvent> {} // TODO: Implement
+
+  // @ts-ignore
+  async deleteEvent(id: string): Promise<BotonicEvent | undefined> {} // TODO: Implement
```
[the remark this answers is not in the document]
asastre: Yes, we will implement these empty methods in another PR

packages/botonic-api/src/data-provider/local-data-provider.ts
```diff
@@ -22,143 +22,117 @@ export class LocalDevDataProvider implements DataProvider {
...
-      const userList = this.db.getObject<Record<string, User>>(path)
-      const users = Object.values(userList)
-      const from = offset * limit
-      const to = limit !== 0 ? from + limit : undefined
-      return users.slice(from, to)
-    } catch (e) {
-      console.error('Error fetching users from local DB', e)
-    }
-    return []
+    this.db.reload()
```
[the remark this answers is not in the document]
asastre: The local DB used is a JSON file, and since it can be updated from differents sides (websockets, api) and we have different instances for each entry point, if we don't reload on every call we will use the data loaded in memory instead of using the data from the JSON file (which could have been changed from other side) and we could end up missing some entries.

packages/botonic-api/src/data-provider/local-data-provider.ts
asastre: I think it won't be needed in this situation since all methods differ a little bit from each other and the common method would be too complicated adding specific behaviour for each event.

packages/botonic-api/src/rest/routes/events.ts
asastre: Yes, you're right, the `??` operator here is redundant. Removed!

packages/botonic-api/src/data-provider/local-data-provider.ts
```diff
@@ -22,143 +22,117 @@ export class LocalDevDataProvider implements DataProvider {
...
+    const userList = this.db.getObject<Record<string, User>>(path)
+    const users = Object.values(userList)
+    const from = offset
```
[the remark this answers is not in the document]
asastre: You're right! Before I was doing an operation with `offset` and assigning the result to `from`, but at the end I'm using `offset` directly, so there's no need to rename it to `from`. Thanks!

packages/botonic-api/src/rest/routes/validation/events.ts
asastre: Yes, it's how express validator tackles the validation of arrays -> https://express-validator.github.io/docs/wildcards.html

vanbasten17 (approved):

PR #1730, opened 2021-07-22
Full stack: Persist user/events data in bot-input endpoint

asastre:
## Description
Persist user/events data in `bot-input` endpoint using the `DataProvider`.

Also, I've changed the `bot` argument `rest/routes` is receivinig to be directly the `NodeApp` (see [related PR](https://github.com/hubtype/botonic/pull/1731)) instead of an object containing other unused attributes.

## Context
User and events data was not stored when received from the `bot-input` endpoint.

Instead of adding a `getUserByProviderId` method I preferred to add a generic method (`getUserByField`) to search users by any field.
@ericmarcos @vanbasten17 @marcpdw if you're ok with this implementation maybe it will be nice to deprecate the `getUserByWebsocketId` method and use `getUserByField` instead.

Also, to make the code cleaner, I've add all `bot-input` endpoint's code into a sepparate file, the same way we already have with the other endpoints

## Testing

### DIFF
--- a/packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
+++ b/packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
@@ -68,6 +68,16 @@ export class DynamoDBDataProvider implements DataProvider {
     return result.Item
   }
 
+  async getUserByWebsocketId(websocketId: string): Promise<User | undefined> {
+    const result = await this.userEventsTable.query(websocketId, {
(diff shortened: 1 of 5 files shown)


vanbasten17 (approved):
LGTM! Having `getUserByField` sounds okay to me, but take into account that for supporting the rest of fields in dynamodb,  modifying the table definitions will be needed (adding secondary indexes), so it will be not as usable as for local queries. We should discuss if we need to extend this functionality in dynamodb

PR #1740, opened 2021-07-28
feat(contentful): add paging in searchByKeywords

asastre:
## Description

- Adding `paging` parameter in `searchByKeywords` the same way it's done with `topContents`.
- Fixed an error with the `perf_hooks` dependency that was making the bot's webpack to fail when compiling.

## Context
Contentful requests have a limit of 100 contents and some times there's a need to obtain more than 100 contents when searching by keywords.

If no paging object is passed to `searchByKeywords` the default one (with a limit of 1000) wil be set.

## Testing

### DIFF
--- a/packages/botonic-plugin-contentful/tests/search/search-by-keywords.test.ts
+++ b/packages/botonic-plugin-contentful/tests/search/search-by-keywords.test.ts
@@ -1,6 +1,6 @@
 import 'jest-extended'
 
-import { deepEqual, instance, mock, when } from 'ts-mockito'
+import { anything, deepEqual, instance, mock, when } from 'ts-mockito'
 
 import {
   CommonFields,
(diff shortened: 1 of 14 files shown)


vanbasten17 (approved):

packages/botonic-plugin-contentful/tests/search/search-by-keywords.test.ts
```diff
@@ -134,7 +134,7 @@ export function keywordsWithMockCms(
   stemBlackList: { [locale: string]: StemmingBlackList[] } = {}
 ): SearchByKeywords {
   const mockCms = mock(DummyCMS)
-  when(mockCms.contentsWithKeywords(deepEqual(context))).thenResolve(
+  when(mockCms.contentsWithKeywords(deepEqual(context), undefined)).thenResolve(
```
marcpdw, line 137: Since it's optional, you don't need to pass `undefined`, you can ignore the passing

marcpdw (approved):
Looks good, I just left a comment to consider

packages/botonic-plugin-contentful/tests/search/search-by-keywords.test.ts
asastre: I added the `undefined` because the test was failing.
I've changed it to `anything()` to keep it clearer, but the thing is that mockito doesn't accept the `when` configuration to be set without an optional parameter and then calling the mocked function with this optional parameter setted (and vice versa) because it won't return the desired value, it will return `null`.
As you can see, there's an [open issue](https://github.com/NagRock/ts-mockito/issues/196) regarding this situation, it seems that the expected behaviour when configuring the `when` is not 100% clear.

marcpdw: Ah right, go for it then!

PR #1742, opened 2021-07-29
feat(api): implement /users endpoint

asastre:
## Description
- Implementation of users routes for CRUD operations. No authentication is done so far.
- Added validations to the parameters received in the requests using `express-validator`.
- Added missing method `deleteUser` in the `DataProvider` interface and its implementations.
- Add validation in patch and put routes to ensure the ID received in the param is the same as the one received in the body.
- Refactor validations to unify common validations for events and users endpoints.

## Context
The api endpoint of users were not implemented yet.

The same way it has been done with events endpoint, all the data validation has been done based on the user type declaration.
For the `patch` request, in order to take advantage of all the validantions done, a method has been implemented to transform all fields to optional fields but still keep all the other validations.

## Testing

### DIFF
--- a/packages/botonic-api/src/rest/routes/events.ts
+++ b/packages/botonic-api/src/rest/routes/events.ts
@@ -36,7 +36,7 @@ router
     }
   )
   .post(async (req, res) => {
-    const errors = await validateBotonicEventData(req)
+    const errors = await validateBotonicEventData({ request: req })
     if (!errors.isEmpty()) {
       res.status(400).send({ errors: errors.array({ onlyFirstError: true }) })
       return
@@ -82,7 +82,11 @@ router
     }
   })
   .put(async (req, res) => {
-    const errors = await validateBotonicEventData(req, false, true)
+    const errors = await validateBotonicEventData({
+      request: req,
+      allFieldsOptional: false,
+      withParamEventId: true,
+    })
     if (!errors.isEmpty()) {
       res.status(400).send({ errors: errors.array({ onlyFirstError: true }) })
       return
@@ -102,12 +106,8 @@ router
           .send({ error: `Event with ID '${params.eventId}' not found` })
         return
       }
-      if (event.eventId !== updatedEvent.eventId) {
-        res
-          .status(400)
-          .send({ error: `Both event ID (params and body) must match` })
-        return
-      }
+
+      updatedEvent.eventId = event.eventId
       await dp.updateEvent(updatedEvent)
       res.status(200).send(updatedEvent)
     } catch (e) {
@@ -115,7 +115,11 @@ router
     }
   })
   .patch(async (req, res) => {
-    const errors = await validateBotonicEventData(req, true, true)
+    const errors = await validateBotonicEventData({
+      request: req,
+      allFieldsOptional: true,
+      withParamEventId: true,
+    })
     if (!errors.isEmpty()) {
       res.status(400).send({ errors: errors.array({ onlyFirstError: true }) })
       return
@@ -135,12 +139,8 @@ router
           .send({ error: `Event with ID '${params.eventId}' not found` })
         return
       }
-      if (newEventData.eventId && event.eventId !== newEventData.eventId) {
-        res
-          .status(400)
-          .send({ error: `Both user ID (params and body) must match` })
-        return
-      }
+
+      newEventData.eventId = event.eventId
       const updatedEvent = { ...event, ...newEventData } as BotonicEvent
       await dp.updateEvent(updatedEvent)
       res.status(200).send(updatedEvent)
--- a/packages/botonic-api/src/rest/routes/events.ts
+++ b/packages/botonic-api/src/rest/routes/events.ts
@@ -102,6 +102,12 @@ router
           .send({ error: `Event with ID '${params.eventId}' not found` })
         return
       }
+      if (event.eventId !== updatedEvent.eventId) {
+        res
+          .status(400)
+          .send({ error: `Both event ID (params and body) must match` })
+        return
+      }
       await dp.updateEvent(updatedEvent)
       res.status(200).send(updatedEvent)
     } catch (e) {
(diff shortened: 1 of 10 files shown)


vanbasten17 (approved):

packages/botonic-api/src/rest/routes/events.ts
```diff
@@ -109,26 +102,26 @@ router
           .send({ error: `Event with ID '${params.eventId}' not found` })
         return
       }
+      if (event.eventId !== updatedEvent.eventId) {
+        res
+          .status(400)
+          .send({ error: `Both event ID (params and body) must match` })
+        return
+      }
```
ericmarcos, line 110: This shouldn't be needed. The ID is already encoded in the URL, there should be no ID in the body (or if there is, just ignore it and always use the URL one)

packages/botonic-api/src/rest/routes/events.ts
```diff
@@ -82,20 +81,14 @@ router
       res.status(500).send({ error: e.message })
     }
   })
-  .put(checkSchema({ eventId: eventIdParamSchema }), async (req, res) => {
-    const errors = await validateBotonicEventData(req)
+  .put(async (req, res) => {
+    const errors = await validateBotonicEventData(req, false, true)
```
ericmarcos, line 85: named parameters better? it's hard to understand what this does without looking at the function signature.

packages/botonic-api/src/rest/routes/events.ts
```diff
@@ -142,6 +135,12 @@ router
           .send({ error: `Event with ID '${params.eventId}' not found` })
         return
       }
+      if (newEventData.eventId && event.eventId !== newEventData.eventId) {
+        res
+          .status(400)
+          .send({ error: `Both user ID (params and body) must match` })
+        return
+      }
```
ericmarcos, line 143: Same as with User

packages/botonic-api/src/rest/routes/users.ts
```diff
@@ -1,55 +1,182 @@
...
   .route('/')
-  .get((_, res) => {
-    res.send('GET HTTP method on user resource')
-  })
-  .post((_, res) => {
-    res.send('POST HTTP method on user resource')
+  .get(
+    checkSchema({
+      limit: limitParamSchema,
+      offset: offsetParamSchema,
```
ericmarcos, line 24: The most common way to add pagination to an API is to define a param "page" (by default 1) and an optional "pageSize" (default typically 10 or 20), with pageSize having a maximum allowed (for instance 100).
Is there any particular reason you used "limit" and "offset"?

packages/botonic-api/src/rest/routes/users.ts
```diff
@@ -1,55 +1,182 @@
...
+          .status(404)
+          .send({ error: `User with ID '${params.userId}' not found` })
+        return
+      }
+      if (user.id !== updatedUser.id) {
+        res
+          .status(400)
+          .send({ error: `Both user ID (params and body) must match` })
+        return
+      }
```
ericmarcos, line 115: not needed

packages/botonic-api/src/rest/routes/users.ts
```diff
@@ -1,55 +1,182 @@
...
+            .status(404)
+            .send({ error: `User with ID '${params.userId}' not found` })
+          return
+        }
+        if (newUserData.id && user.id !== newUserData.id) {
+          res
+            .status(400)
+            .send({ error: `Both user ID (params and body) must match` })
+          return
+        }
```
ericmarcos, line 150: not needed

ericmarcos (changes requested):

packages/botonic-api/src/rest/routes/users.ts
```diff
@@ -1,55 +1,182 @@
...
+        const errors = validationResult(req)
+        if (!errors.isEmpty()) {
+          res.status(400).send({ errors: errors.array() })
+          return
+        }
+
+        const query = matchedData(req, { locations: ['query'] })
+        const dp = dataProviderFactory(process.env.DATA_PROVIDER_URL)
+        const users = await dp.getUsers(query.limit, query.offset)
+        res.status(200).send(users)
```
ericmarcos, line 37: Usually when returning results in a paginated API you'd expect an object like this:

```
{
  "prev": "<link to prev page>",
  "next": "<link to next page>",
  "count": <total number of records>,
  "results": [{...user1}, {...user2}, ...]
}
```

packages/botonic-api/src/rest/routes/events.ts
asastre: Make sense, fixed!

packages/botonic-api/src/rest/routes/events.ts
asastre: Agree, changed!

packages/botonic-api/src/rest/routes/events.ts
asastre: Removed!

packages/botonic-api/src/rest/routes/users.ts
asastre: Removed!

packages/botonic-api/src/rest/routes/users.ts
asastre: Removed!

packages/botonic-api/src/rest/routes/users.ts
asastre: I've seen it done both ways and decided to use the limit/offset one. 
For the developer maybe it's easier to use the page/pageSize...I'll change it to page/pageSize in another PR and I'll add also the `prev`, `next`, `count`, `results` in the api response as you said in another comment.

packages/botonic-api/src/rest/routes/users.ts
asastre: :+1: I'll add this into another PR with the `page`/`pageSize` refactor.

ericmarcos (approved):

PR #1746, opened 2021-08-02
feat(cli/pulumi): add deploy command to aws/hubtype to Botonic CLI

vanbasten17:
## Description
* Added extra arguments to deploy command to determine to which cloud provider the bot is deployed.
* Added @botonic/pulumi as CLI's dependency
* Added deploy/destroy commands in PulumiRunner class for simplicity and move build commands to pulumi package.
* Upgraded pulumi/pulumi and pulumi/aws dependencies to use latest versions.

## Context
We needed a way to choose where to deploy our bot now that we have the option of deploying it to AWS in Botonic 1.0.
Calling `botonic deploy` or `botonic deploy hubtype` will fallback to current functionality.

In new Botonic 1.0 project the following file should be added in the root directory:
**aws.config.js**
```
exports.default = {
  region: 'eu-west-1',
  profile: 'my-profile-name',
  projectName: 'myProject',
  tags: {
    Scope: 'Botonic-1.0',
    Maintainer: 'team.botonic',
  },
  customDomain: 'my-custom-domain.com',
  stackName: 'dev',
  tableName: 'user_events',
}
```
also the following script can be added in the project's package.json:
```
"destroy": "yarn bt destroy",
"destroy:aws": "yarn bt destroy aws" // as a shortcut
```

Just run `yarn deploy aws`, `yarn deploy` or `yarn deploy hubtype` in a Botonic 1.0 project configured as mentioned.

### DIFF
--- a/packages/botonic-cli/src/commands/deploy.ts
+++ b/packages/botonic-cli/src/commands/deploy.ts
@@ -26,6 +26,12 @@ let npmCommand: string | undefined
 const BOTONIC_BUNDLE_FILE = 'botonic_bundle.zip'
 const BOTONIC_TEMP_DIRNAME = 'tmp'
 
+interface DeployHubtypeFlags {
+  command?: string
+  botName?: string
+  email?: string
+  password?: string
+}
 export default class Run extends Command {
   static description = 'Deploy Botonic project to cloud provider'
 
@@ -68,37 +74,38 @@ Deploying to AWS...
   /* istanbul ignore next */
   async run(): Promise<void> {
     const { flags, args } = this.parse(Run)
-    const provider = args.provider
-    if (provider && provider !== CLOUD_PROVIDERS.HUBTYPE) {
-      this.deployToProvider(provider)
-    } else {
-      this.telemetry.trackDeploy()
-      npmCommand = flags.command
-      this.botName = flags.botName
-      const email = flags.email
-      const password = flags.password
-      if (email && password) await this.login(email, password)
-      else if (!this.botonicApiService.oauth) await this.signupFlow()
-      else if (this.botName) {
-        await this.deployBotFromFlag(this.botName)
-      } else await this.deployBotFlow()
+    const provider = args.provider || CLOUD_PROVIDERS.HUBTYPE
+    // TODO: -> In a next iteration it would be cool to get a prompt asking which provider we prefer (if it's the 1st deploy) or getting the provider from `.botonic.json` (after 1st deploy)
+    this.telemetry.trackDeploy1_0({ provider })
+    console.log(`Deploying to ${provider}...`)
+    console.log('This can take a while, do not cancel this process.')
+    if (provider === CLOUD_PROVIDERS.AWS) this.deployAWS()
+    else if (provider === CLOUD_PROVIDERS.HUBTYPE) this.deployHubtype(flags)
+  }
+
+  async deployAWS(): Promise<void> {
+    const pulumiRunner = new PulumiRunner(PATH_TO_AWS_CONFIG)
+    try {
+      await pulumiRunner.deploy()
+    } catch (e) {
+      const error = `Deploy Botonic 1.0 ${CLOUD_PROVIDERS.AWS} Error: ${String(
+        e
+      )}`
+      this.telemetry.trackError(error)
+      throw new Error(e)
     }
   }
 
-  async deployToProvider(provider: string): Promise<void> {
-    this.telemetry.trackDeploy1_0({ provider })
-    if (provider === CLOUD_PROVIDERS.AWS) {
-      console.log(`Deploying to ${CLOUD_PROVIDERS.AWS}...`)
-      console.log('This can take a while, do not cancel this process.')
-      const pulumiRunner = new PulumiRunner(PATH_TO_AWS_CONFIG)
-      try {
-        await pulumiRunner.deploy()
-      } catch (e) {
-        const error = `Deploy Botonic 1.0 ${provider} Error: ${String(e)}`
-        this.telemetry.trackError(error)
-        throw new Error(e)
-      }
-    }
+  async deployHubtype(flags: DeployHubtypeFlags): Promise<void> {
+    npmCommand = flags.command
+    this.botName = flags.botName
+    const email = flags.email
+    const password = flags.password
+    if (email && password) await this.login(email, password)
+    else if (!this.botonicApiService.oauth) await this.signupFlow()
+    else if (this.botName) {
+      await this.deployBotFromFlag(this.botName)
+    } else await this.deployBotFlow()
   }
 
   async deployBotFromFlag(botName: string): Promise<void> {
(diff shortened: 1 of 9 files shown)


marcpdw (approved):
LGTM!

packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -54,28 +60,49 @@ Uploading...
...
+      else if (!this.botonicApiService.oauth) await this.signupFlow()
+      else if (this.botName) {
+        await this.deployBotFromFlag(this.botName)
+      } else await this.deployBotFlow()
+    }
+  }
+
+  async deployToProvider(provider: string): Promise<void> {
+    this.telemetry.trackDeploy1_0({ provider })
+    if (provider === CLOUD_PROVIDERS.AWS) {
```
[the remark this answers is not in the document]
vanbasten17: yes, that's the idea! From the start, we'll give support for AWS, but in the future we could consider alternatives like Google Cloud or Microsoft Azure.

packages/botonic-cli/src/commands/destroy.ts
```diff
@@ -0,0 +1,46 @@
+import { PulumiRunner } from '@botonic/pulumi/lib/pulumi-runner'
+import { Command } from '@oclif/command'
+
+import { Telemetry } from '../analytics/telemetry'
+import { CLOUD_PROVIDERS, PATH_TO_AWS_CONFIG } from '../constants'
+
+export default class Run extends Command {
+  static description = 'Deploy Botonic project to hubtype.com'
```
asastre, line 8: Wrong, description, right?

packages/botonic-cli/src/commands/deploy.ts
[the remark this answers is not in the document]
asastre: What about showing a message like `'Provider not available to deploy to'` if a different provider than AWS is being set to deploy to?

packages/botonic-cli/src/commands/destroy.ts
```diff
@@ -0,0 +1,46 @@
...
+    const { args } = this.parse(Run)
+    const provider = args.provider
+    if (provider) {
+      this.destroyProvider(provider)
+    }
+  }
+
+  async destroyProvider(provider: string): Promise<void> {
+    this.telemetry.trackDeploy({ provider })
+    if (provider === CLOUD_PROVIDERS.AWS) {
```
asastre, line 32: Same here, what happens if the developer sets a wrong provider or a provider that is not supported right now? Should we show a message?

asastre (commented):
Looks good overall! Just some small observations.

packages/botonic-cli/src/commands/destroy.ts
vanbasten17: good point! in fact in the `args` member of `Run` class (from occlif) you can set options, which is set with `Object.values(CLOUD_PROVIDERS)`. If the deploy command is called with an argument not in options, the following message will appear:
```
yarn run v1.22.10
$ yarn bt deploy foo
$ /Users/mrabat/Documents/hubtype/botonic/packages/botonic-cli/bin/run deploy foo
 ›   Error: Expected foo to be one of: aws, hubtype
```
as this is resolved as soon as the command is executed, I found it not necessary to add more info in latter steps of the flow.

packages/botonic-cli/src/commands/destroy.ts
vanbasten17: sure, fixed!

packages/botonic-cli/src/commands/destroy.ts
asastre: Thanks for the explanation! Totally agree, no need to add more messages.

asastre (approved):
:rocket:

packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -54,28 +59,49 @@ Uploading...
...
-    if (email && password) await this.login(email, password)
-    else if (!this.botonicApiService.oauth) await this.signupFlow()
-    else if (this.botName) {
-      await this.deployBotFromFlag(this.botName)
-    } else await this.deployBotFlow()
+    const { flags, args } = this.parse(Run)
+    const provider = args.provider
+    if (provider && provider !== CLOUD_PROVIDERS.HUBTYPE) {
+      this.deployToProvider(provider)
+    } else {
```
ericmarcos, line 74: This logic is divided inconsistently between `run()` and `deployToProvider()` functions. Better do something like:

```
async run(): Promise<void> {
  const { flags, args } = this.parse(Run)
  const provider = args.provider || CLOUD_PROVIDERS.HUBTYPE
  // TODO -> In a next iteration it would be cool to get a prompt asking which provider we prefer (if it's the 1st deploy) or getting the provider from `.botonic.json` (after 1st deploy)
  this.telemetry.trackDeploy1_0({ provider })
  console.log(`Deploying to ${provider}...`)
  console.log('This can take a while, do not cancel this process.')
  if (provider === CLOUD_PROVIDERS.AWS) {
    this.deployAWS(...)
  } else if (provider === CLOUD_PROVIDERS.HUBTYPE) {
    this.deployHubtype(...)
  }
}

async deployAWS(): Promise<void> { }
async deployHubtype(): Promise<void> { /* Current code */ }

```

packages/botonic-cli/src/commands/destroy.ts
```diff
@@ -0,0 +1,46 @@
...
+  /* istanbul ignore next */
+  async run(): Promise<void> {
+    const { args } = this.parse(Run)
+    const provider = args.provider
+    if (provider) {
+      this.destroyProvider(provider)
+    }
+  }
+
+  async destroyProvider(provider: string): Promise<void> {
```
ericmarcos, line 30: Follow same function structure as deploy.

packages/botonic-cli/src/commands/destroy.ts
```diff
@@ -0,0 +1,46 @@
...
+      this.telemetry.trackDestroy1_0({ provider })
+      try {
+        const pulumiRunner = new PulumiRunner(PATH_TO_AWS_CONFIG)
+        await pulumiRunner.destroy()
+      } catch (e) {
+        const error = `Destroy Botonic 1.0 ${provider} Error: ${String(e)}`
+        this.telemetry.trackError(error)
+        throw new Error(e)
+      }
+    }
```
ericmarcos, line 44: As a TODO, it shouldn't be too hard to destroy a Hubtype Bot (just a call to `DELETE /bot/<id>` (out of scope of this PR)

ericmarcos (changes requested):

packages/botonic-cli/src/commands/deploy.ts
vanbasten17: I agree, separated the deploy paths consistently 👍

packages/botonic-cli/src/commands/destroy.ts
vanbasten17: Noted as a TODO 👍

PR #1747, opened 2021-08-02
feat(api): add proper pagination to endpoints

asastre:
## Description
Add proper pagination to GET method in `/users` and `/events` endpoints:
- Refactor pagination params: use `page/pageSize` instead of `limit/offset`.
- Return a paginated response using an object like this one:
```
{
  "previous": "<link to previous page>",
  "next": "<link to next page>",
  "count": <total number of records>,
  "results": [{...record1}, {...record2}, ...]
}
```

## Context
`/users` and `/events` endpoints had pagination but were not returning a proper paginated response.

Create a `Paginator` class that handles all the pagination logic.

## Testing

### DIFF
--- a/packages/botonic-api/src/rest/utils/paginator.ts
+++ b/packages/botonic-api/src/rest/utils/paginator.ts
@@ -50,6 +50,7 @@ export class Paginator {
     return {
       previous: this.page > FIRST_PAGE ? this.previousUrl : undefined,
       next: items.length === this.pageSize ? this.nextUrl : undefined,
+      //TODO: count should be the total amount of items, not only the number of items of the actual page
       count: items.length,
       results: items,
     }
--- a/packages/botonic-api/src/rest/utils/paginator.ts
+++ b/packages/botonic-api/src/rest/utils/paginator.ts
@@ -1,0 +1,59 @@
+import { Request } from 'express'
+
+export const FIRST_PAGE = 1
+export const DEFAULT_PAGE_SIZE = 10
+export const MAX_PAGE_SIZE = 100
+
+type JsonResponse<T> = {
+  previous?: string
+  next?: string
+  count: number
+  results: T[]
+}
+
+export class Paginator {
+  private readonly baseUrl: string
+  private readonly page: number
+  private readonly pageSize: number
+
+  constructor(
+    request: Request,
+    page = FIRST_PAGE,
+    pageSize = DEFAULT_PAGE_SIZE
+  ) {
+    this.baseUrl = this.getBaseUrl(request)
+    this.page = page < FIRST_PAGE ? FIRST_PAGE : page
+    this.pageSize = pageSize > MAX_PAGE_SIZE ? MAX_PAGE_SIZE : pageSize
+  }
+
+  get limit(): number {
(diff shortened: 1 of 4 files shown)


packages/botonic-api/src/rest/utils/paginator.ts
```diff
@@ -0,0 +1,59 @@
...
+  private readonly pageSize: number
+
+  constructor(
+    request: Request,
+    page = FIRST_PAGE,
+    pageSize = DEFAULT_PAGE_SIZE
+  ) {
+    this.baseUrl = this.getBaseUrl(request)
+    this.page = page < FIRST_PAGE ? FIRST_PAGE : page
+    this.pageSize = pageSize > MAX_PAGE_SIZE ? MAX_PAGE_SIZE : pageSize
```
vanbasten17, line 26: are you considering negative values here?

vanbasten17 (approved):
Looking good!

packages/botonic-api/src/rest/utils/paginator.ts
asastre: No, I'm forcing always the `page` to be at least 1 (`FIRST_PAGE`) and the `pageSize` to be as maximum as `MAX_PAGE_SIZE`.
Is this what you're asking?

marcpdw: Good observation here

marcpdw (approved):

packages/botonic-api/src/rest/utils/paginator.ts
vanbasten17: yes, I meant for negative `pageSize` values.

asastre: Yes, good point! 
I've added a minimum value of 1 for `pageSize`.

packages/botonic-api/src/rest/utils/paginator.ts
```diff
@@ -0,0 +1,61 @@
...
+  get nextUrl(): string {
+    const nextPage = this.page + 1
+    return `${this.baseUrl}/?page=${nextPage}&pageSize=${this.pageSize}`
+  }
+
+  generateResponse<T>(items: T[]): JsonResponse<T> {
+    return {
+      previous: this.page > FIRST_PAGE ? this.previousUrl : undefined,
+      next: items.length === this.pageSize ? this.nextUrl : undefined,
+      count: items.length,
```
ericmarcos, line 53: This should be the size of the whole collection (if we didn't apply pagination). The way it's implemented now it will return the lenght of the results array (typically this is not passed in the response because it's trivial to do `data.results.lenght` in the client). In any case, I'm not sure it's easy to get the total count of a query from Dynamodb, so let's leave it like this until we find out.

ericmarcos (approved):

packages/botonic-api/src/rest/utils/paginator.ts
asastre: You're right, this should be the total amount of items. 
Added a TODO to change it in the next iteration.

PR #1758, opened 2021-08-03
feat(pulumi): cache invalidations for updated static assets

vanbasten17:
## Description
* Do invalidate cache for updated S3 objects after deploying Botonic 1.0 full-stack.

## Context
As a cache system, cloudfront caches the requested origin objects for some time. This had as a consequence that after modifying static assets (`index.html, webchat.botonic.js, ...`) the changes were not instantly applied so several time had to pass to see the effects ([outdated content](https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-serving-outdated-content-s3/)).
In order to tell cloudfront to request again these files when the web content is loaded, we have to do what is called [cache invalidation](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html).
There are some different ways to solve this problem:
* Creating dynamic providers with Pulumi.
* Create an isolated lambda to detect S3 bucket changes and trigger the invalidations.
* Detecting in some other way what have been the changes applied and run the invalidation with them.

The first two solutions seemed to me to be more complex that what we need, so I came up with:
1. Running the preview command within the automation api (`previewStack`) before deploying (with `updateStack`). This gives us an output with a log with the changed files.
2. Parsing the output by filtering the changes we are interested (filtering lines by `~` which denotes and update and `aws:s3:bucketObject`).
3. Running invalidations after the full-stack has been deployed.

### DIFF
--- a/packages/botonic-pulumi/package.json
+++ b/packages/botonic-pulumi/package.json
@@ -4,11 +4,11 @@
   "scripts": {
     "build": "rm -rf lib && ../../node_modules/.bin/tsc",
     "build:watch": "rm -rf lib && ../../node_modules/.bin/tsc -w",
-    "lint": "npm run lint_core -- --fix",
-    "lint_ci": "npm run lint_core -- -c ../.eslintrc_slow.js",
-    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.ts*'",
+    "lint": "npm run lint-core -- --fix",
+    "lint-ci": "npm run lint-core -- -c ../.eslintrc_slow.js",
+    "lint-core": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.ts*'",
     "test": "../../node_modules/.bin/jest --coverage",
-    "test_ci": "../../node_modules/.bin/jest --coverage --ci --reporters=default --reporters=jest-junit"
+    "test-ci": "../../node_modules/.bin/jest --coverage --ci --reporters=default --reporters=jest-junit"
   },
   "devDependencies": {
     "@types/concurrently": "^6.2.0",
--- a/packages/botonic-pulumi/package.json
+++ b/packages/botonic-pulumi/package.json
@@ -3,7 +3,12 @@
   "version": "1.0.0-alpha.2",
   "scripts": {
     "build": "rm -rf lib && ../../node_modules/.bin/tsc",
-    "build:watch": "rm -rf lib && ../../node_modules/.bin/tsc -w"
+    "build:watch": "rm -rf lib && ../../node_modules/.bin/tsc -w",
+    "lint": "npm run lint_core -- --fix",
+    "lint_ci": "npm run lint_core -- -c ../.eslintrc_slow.js",
+    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.ts*'",
+    "test": "../../node_modules/.bin/jest --coverage",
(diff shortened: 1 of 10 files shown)


ericmarcos (approved):
Let's move on with this, but I want to note that relying on the output log of some 3rd party command is not the best approach. That output can easily change, or there can be instances that we didn't consider that can break our parser.
Programs that rely on parsing logs = fragility
Programs that rely on well defined APIs = robustness

packages/botonic-pulumi/package.json
```diff
@@ -3,7 +3,12 @@
   "version": "1.0.0-alpha.2",
   "scripts": {
     "build": "rm -rf lib && ../../node_modules/.bin/tsc",
-    "build:watch": "rm -rf lib && ../../node_modules/.bin/tsc -w"
+    "build:watch": "rm -rf lib && ../../node_modules/.bin/tsc -w",
+    "lint": "npm run lint_core -- --fix",
+    "lint_ci": "npm run lint_core -- -c ../.eslintrc_slow.js",
+    "lint_core": "../../node_modules/.bin/eslint_d --cache --quiet 'src/**/*.ts*'",
+    "test": "../../node_modules/.bin/jest --coverage",
+    "test_ci": "../../node_modules/.bin/jest --coverage --ci --reporters=default --reporters=jest-junit"
```
asastre, line 11: In `@botonic/dx` we're using hyphen instead of underscore for npm script names after agreeing it internally. I think we should try to unify criteria and use hyphens here too.

packages/botonic-pulumi/src/pulumi-runner.ts
```diff
@@ -222,9 +235,38 @@ export class PulumiRunner {
...
+    updatedBucketObjects: string[]
+  ): Promise<void> {
+    try {
+      const cacheInvalidator = new CacheInvalidator({
+        region: this.projectConfig.region,
+        profile: this.projectConfig.profile,
+        accessKey: this.projectConfig.accessKey,
+        secretKey: this.projectConfig.secretKey,
+        token: this.projectConfig.token,
+      })
```
asastre, line 258: Could `this.projectConfig` be passed directly? All the attributes of the passed object have the same names.

packages/botonic-pulumi/src/pulumi-runner.ts
```diff
@@ -222,9 +235,38 @@ export class PulumiRunner {
...
+      const cloudfrontId = updateResults.outputs['cloudfrontId'].value
+      await cacheInvalidator.invalidateBucketObjects(
+        cloudfrontId,
+        '/',
```
asastre, line 262: Put it in a constant?

asastre (approved):
LGTM! Agree with @ericmarcos's comment, but for as he say, for now let's move on with this!
Added some observations :wink:

packages/botonic-pulumi/package.json
vanbasten17: modified 👍

packages/botonic-pulumi/src/pulumi-runner.ts
vanbasten17: yes, in fact I wanted to be explicit because projectConfig comes with other attributes which doesn't match with the `AWSCredentials` type that CacheInvalidator is expecting. But as it is a subset of the config, I have passed it like `this.projectConfig as AWSCredentials`.

PR #1762, opened 2021-08-04
fix(pulumi): throw error when exceeding name chars threshold

vanbasten17:
## Description
* Limited the amount of valid characters for combinations of `projectName` + `stackName`.

## Context
Some of AWS components (buckets, lambdas, ...) has a limitation in their names length of ~64 characters. As we append a generated prefix (combination of project name + stack name) to name the components and make them easy to identify, sometimes this maximum was hit.

Define a reasonable total maximum string length of 30 (30 + 1 from separator) and throw asap to not continue propagating this error when deploying in Botonic 1.0.

### DIFF
--- a/packages/botonic-pulumi/src/index.ts
+++ b/packages/botonic-pulumi/src/index.ts
@@ -2,10 +2,27 @@ import { Config } from '@pulumi/pulumi'
 import { join } from 'path'
 import { cwd } from 'process'
 
-export const getNamePrefix = (): string => {
+export function getNamePrefix(): string {
   const config = new Config()
-  return `botonic-${config.get('projectName')}-${config.get('stackName')}`
+  const prefix = generatePrefix(
+    config.get('projectName') as string,
+    config.get('stackName') as string
+  )
+  return prefix
 }
+
+export function generatePrefix(projectName: string, stackName: string): string {
+  const SEPARATOR = '-'
+  const prefix = `${projectName}${SEPARATOR}${stackName}`
+  const MAX_LENGTH = 30
+  if (prefix.length > MAX_LENGTH + 1) {
+    throw new Error(
+      `The combination of 'projectName' and 'stackName' names can not exceed ${MAX_LENGTH} chars.`
+    )
+  }
+  return prefix
+}
+
 export const NLP_MODELS_PATH = join(cwd(), 'bot', 'src', 'nlp', 'tasks')
 export const WEBSOCKET_SERVER_PATH = join(cwd(), 'api', 'dist', 'websocket')
 export const REST_SERVER_PATH = join(cwd(), 'api', 'dist', 'rest')
--- a/packages/botonic-pulumi/src/index.ts
+++ b/packages/botonic-pulumi/src/index.ts
@@ -2,22 +2,22 @@ import { Config } from '@pulumi/pulumi'
 import { join } from 'path'
 import { cwd } from 'process'
 
+export const PROJECT_NAME_SEPARATOR = '-'
+export const MAX_PROJECT_NAME_LENGTH = 30
+
 export function getNamePrefix(): string {
   const config = new Config()
-  const prefix = generatePrefix(
+  return generatePrefix(
     config.get('projectName') as string,
     config.get('stackName') as string
   )
-  return prefix
 }
 
 export function generatePrefix(projectName: string, stackName: string): string {
-  const SEPARATOR = '-'
-  const prefix = `${projectName}${SEPARATOR}${stackName}`
-  const MAX_LENGTH = 30
-  if (prefix.length > MAX_LENGTH + 1) {
+  const prefix = `${projectName}${PROJECT_NAME_SEPARATOR}${stackName}`
+  if (prefix.length > MAX_PROJECT_NAME_LENGTH + PROJECT_NAME_SEPARATOR.length) {
     throw new Error(
-      `The combination of 'projectName' and 'stackName' names can not exceed ${MAX_LENGTH} chars.`
+      `The combination of 'projectName' and 'stackName' names can not exceed ${MAX_PROJECT_NAME_LENGTH} chars.`
     )
   }
   return prefix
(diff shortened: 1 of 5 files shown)


packages/botonic-pulumi/src/index.ts
```diff
@@ -2,10 +2,27 @@ import { Config } from '@pulumi/pulumi'
...
+    config.get('stackName') as string
+  )
+  return prefix
 }
+
+export function generatePrefix(projectName: string, stackName: string): string {
+  const SEPARATOR = '-'
+  const prefix = `${projectName}${SEPARATOR}${stackName}`
+  const MAX_LENGTH = 30
+  if (prefix.length > MAX_LENGTH + 1) {
```
asastre, line 18: Use `SEPARATOR.length` instead of `1`?

asastre (approved):

packages/botonic-pulumi/src/index.ts
```diff
@@ -2,10 +2,27 @@ import { Config } from '@pulumi/pulumi'
...
 
-export const getNamePrefix = (): string => {
+export function getNamePrefix(): string {
   const config = new Config()
-  return `botonic-${config.get('projectName')}-${config.get('stackName')}`
+  const prefix = generatePrefix(
+    config.get('projectName') as string,
+    config.get('stackName') as string
+  )
+  return prefix
```
marcpdw, line 11: Unless you're leaving it like this for debugging reasons, you can return the call directly, no need of the variable.

packages/botonic-pulumi/src/index.ts
```diff
@@ -2,10 +2,27 @@ import { Config } from '@pulumi/pulumi'
...
 }
+
+export function generatePrefix(projectName: string, stackName: string): string {
+  const SEPARATOR = '-'
+  const prefix = `${projectName}${SEPARATOR}${stackName}`
+  const MAX_LENGTH = 30
```
marcpdw, line 17: You can consider to move `SEPARATOR` & `MAX_LENGTH` out of the method. They are proper constants, no need to be instantiated on every call.

packages/botonic-pulumi/src/pulumi-runner.ts
```diff
@@ -92,9 +96,10 @@ export class PulumiRunner {
   ): Promise<Stack> {
     const projectName = this.projectConfig?.projectName || 'botonic'
     const stackName = this.projectConfig?.stackName || 'full-stack'
+    const prefix = generatePrefix(projectName, stackName)
     const args: InlineProgramArgs = {
       projectName,
-      stackName: `${stackToDeploy}-${stackName}`,
+      stackName: `${prefix}-${stackToDeploy}`,
```
marcpdw, line 102: Has this `-` anything to do with the `SEPARATOR`? If so, you may want to consider its declaration and export to be reused across files.

marcpdw (commented):
On the right track, take a look at the comments.

vanbasten17:
thanks for the feedback @asastre, @marcpdw , applied your changes, they sound reasonable to me 💯

packages/botonic-pulumi/src/index.ts
```diff
@@ -2,10 +2,27 @@ import { Config } from '@pulumi/pulumi'
...
+    config.get('stackName') as string
+  )
+}
+
+export function generatePrefix(projectName: string, stackName: string): string {
+  const prefix = `${projectName}${PROJECT_NAME_SEPARATOR}${stackName}`
+  if (prefix.length > MAX_PROJECT_NAME_LENGTH + PROJECT_NAME_SEPARATOR.length) {
+    throw new Error(
+      `The combination of 'projectName' and 'stackName' names can not exceed ${MAX_PROJECT_NAME_LENGTH} chars.`
+    )
```
ericmarcos, line 21: This is the error that will be shown to the developer right? Better be more explicit with something like `` `Provided projectName "${projectName}" and stackName "${stackName}" that combined exceed the max allowed length: ${prefix.length - PROJECT_NAME_SEPARATOR.length} / ${MAX_PROJECT_NAME_LENGTH - PROJECT_NAME_SEPARATOR.length}` ``

packages/botonic-pulumi/src/index.ts
```diff
@@ -2,10 +2,27 @@ import { Config } from '@pulumi/pulumi'
...
+export function getNamePrefix(): string {
   const config = new Config()
-  return `botonic-${config.get('projectName')}-${config.get('stackName')}`
+  return generatePrefix(
+    config.get('projectName') as string,
+    config.get('stackName') as string
+  )
+}
+
+export function generatePrefix(projectName: string, stackName: string): string {
```
ericmarcos, line 16: Better more explicit function name: `generateStackNamePrefix()`

packages/botonic-pulumi/src/index.ts
```diff
@@ -2,10 +2,27 @@ import { Config } from '@pulumi/pulumi'
 import { join } from 'path'
 import { cwd } from 'process'
 
-export const getNamePrefix = (): string => {
+export const PROJECT_NAME_SEPARATOR = '-'
+export const MAX_PROJECT_NAME_LENGTH = 30
+
+export function getNamePrefix(): string {
```
ericmarcos, line 8: Can we refactor this name to be more explicit too? Like `getProviderNamePrefix`

ericmarcos (changes requested):

packages/botonic-pulumi/src/index.ts
vanbasten17: sounds good, changed 👍

packages/botonic-pulumi/src/index.ts
vanbasten17: I have changed it to `generateProjectStackNamePrefix`, which is the most explicit one.

packages/botonic-pulumi/src/index.ts
vanbasten17: not only used for naming the provider but for naming also the different pulumi components that are deployed. I've left it as `getProjectStackNamePrefix`

ericmarcos (approved):

PR #1763, opened 2021-08-05
Fix/dx botonic app config

vanbasten17:
## Description
* Change api-ws config to use the same working config as api-rest config.
* Bump and publish 1.0.0-alpha.0 version.

## Context
After including typescript code of outputParser in `@botonic/core`, the botonic 1.0 project started failing with current configs imported from dx.

### DIFF
--- a/packages/botonic-dx/botonic-app-config/webpack/api-ws.js
+++ b/packages/botonic-dx/botonic-app-config/webpack/api-ws.js
@@ -21,10 +21,10 @@ module.exports = ({
   output: {
     path: path.resolve(projectPath, 'dist/websocket'),
     filename: 'server.js',
-    libraryTarget: 'commonjs',
+    libraryTarget: 'umd',
   },
   module: {
-    rules: [transpilingLoaderConfig('babel'), nullLoaderConfig],
+    rules: [transpilingLoaderConfig('esbuild'), nullLoaderConfig],
   },
   resolve: {
     extensions: ['*', '.js', '.jsx', '.ts', '.tsx'],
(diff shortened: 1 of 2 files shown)


packages/botonic-dx/botonic-app-config/webpack/api-ws.js
```diff
@@ -21,10 +21,10 @@ module.exports = ({
   output: {
     path: path.resolve(projectPath, 'dist/websocket'),
     filename: 'server.js',
-    libraryTarget: 'commonjs',
+    libraryTarget: 'umd',
```
marcpdw, line 24: Out of curiosity, what's the benefit or motivation of this change?

marcpdw (approved):

packages/botonic-dx/botonic-app-config/webpack/api-ws.js
vanbasten17: The first fact is that we had an inconsistency between `api-ws.js` and `api-rest.js`. `api-rest.js`'s libraryTarget's was specifying the `umd` version which is the one that makes the project to be run as expected. Both api-ws and api-rest should be transpiled almost in the same way. Curiously, `commonjs` target, which is the one that targets NodeJS syntax, seems to be the logic one, but in pratice it is not working. `umd` ensures that the bundle works both in server-side and in client-side, so we are leaving as is in this iteration. We should do further investigation when optimizing bundle sizes for every artifact generated for Botonic 1.0.

asastre (approved):

ericmarcos (approved):

PR #1764, opened 2021-08-05
feat(core/api): allow accessing dataProvider within Actions and plugins

vanbasten17:
## Description
* Passing dataProvider to NodeApp in order to allow accessing DB information within Actions and plugins.
* Added implementation of `getUsers` for dynamo data provider.
* Generated missing type declarations in @botonic/core (now that we are using the compiled lib folder of core, we can migrate the package to TS and solve the types approach)
## Context
We need a way to access database from Actions or Plugins if we want to do extra operations when bot gets executed.

Passing already instantiated dataProviders in `websocket` and `rest` apis through `input` function.
Then they can be passed through the `request` object and plugins.
```
const dp = ...
const output = await app.input({
    input: message,
    session: JSON.parse(user.session),
    lastRoutePath: user.route,
    dataProvider: dp,
  })
```
```
async input({ input, session, lastRoutePath, dataProvider }) {
...
// also the same for post mode
await runPlugins(
        this.plugins,
        'pre', 
        input,
        session,
        lastRoutePath,
        undefined,
        undefined,
        dataProvider
      )
```
and attach it to the `request`:
```
const request = {
      getString: stringId => this.getString(stringId, session),
      setLocale: locale => this.setLocale(locale, session),
      session: session || {},
      params: output.params || {},
      input: input,
     ...,
      dataProvider,
    }
```
so it will be accessible from actions' requests and plugins' requests.

**Using dataProvider within a class:**
```
import React from 'react'
import { Text } from '@botonic/react/src/experimental'

  static async botonicInit({ dataProvider }) {
    const users = await dataProvider.getUsers()
    return { users: users }
  }
  render() {
    return (
      <>
        <Text>Welcome to Botonic!</Text>
        <Text>{JSON.stringify(this.props.users)}</Text>
      </>
    )
  }
}
```

**Using dataProvider within a plugin:**
```
async pre(request: PluginPreRequest): Promise<void> {
    const users = await request.dataProvider.getUsers()
    try {
    } catch (e) {}
  }
```
## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
+++ b/packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
@@ -10,6 +10,7 @@ import {
   getUserEventsTable,
   GLOBAL_SECONDARY_INDEX_NAME,
   SORT_KEY_NAME,
+  USER_PREFIX,
 } from './dynamodb-utils'
 
 export class DynamoDBDataProvider implements DataProvider {
@@ -56,7 +57,15 @@ export class DynamoDBDataProvider implements DataProvider {
   }
 
   // @ts-ignore
-  async getUsers(limit = 10, offset = 0): Promise<User[]> {} // TODO: Implement
+  async getUsers(limit = 10, offset = 0): Promise<User[]> {
+    // TODO: offset really needed here?
+    const result = await this.userEventsTable.scan({
+      filters: { attr: SORT_KEY_NAME, beginsWith: USER_PREFIX },
+      limit,
+    })
+    if (result.Count === 0) return []
+    return result.Items
+  }
 
   async getUser(id: string): Promise<User | undefined> {
     const userById = {
--- a/packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
+++ b/packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
@@ -63,7 +63,6 @@ export class DynamoDBDataProvider implements DataProvider {
       filters: { attr: SORT_KEY_NAME, beginsWith: USER_PREFIX },
       limit,
     })
-    if (result.Count === 0) return []
     return result.Items
   }
 
--- a/packages/botonic-core/src/core-bot.js
+++ b/packages/botonic-core/src/core-bot.js
@@ -17,10 +17,8 @@ export class CoreBot {
     defaultDelay,
     defaultRoutes,
     inspector,
-    dataProvider,
   }) {
     this.renderer = renderer
-    this.dataProvider = dataProvider
     this.plugins = loadPlugins(plugins)
     this.theme = theme || {}
     this.defaultTyping =
@@ -51,7 +49,7 @@ export class CoreBot {
     session.__locale = locale
   }
 
-  async input({ input, session, lastRoutePath }) {
+  async input({ input, session, lastRoutePath, dataProvider }) {
     session = session || {}
     if (!session.__locale) session.__locale = 'en'
 
@@ -64,7 +62,7 @@ export class CoreBot {
         lastRoutePath,
         undefined,
         undefined,
-        this.dataProvider
+        dataProvider
       )
     }
 
@@ -90,7 +88,7 @@ export class CoreBot {
       defaultTyping: this.defaultTyping,
       defaultDelay: this.defaultDelay,
       lastRoutePath,
-      dataProvider: this.dataProvider,
+      dataProvider,
     }
 
     const actions = [output.action, output.retryAction, output.defaultAction]
@@ -111,7 +109,7 @@ export class CoreBot {
         lastRoutePath,
         response,
         parsedResponse,
-        this.dataProvider
+        dataProvider
       )
     }
 
--- a/packages/botonic-core/src/core-bot.js
+++ b/packages/botonic-core/src/core-bot.js
@@ -17,8 +17,10 @@ export class CoreBot {
     defaultDelay,
     defaultRoutes,
     inspector,
+    dataProvider,
   }) {
     this.renderer = renderer
+    this.dataProvider = dataProvider
     this.plugins = loadPlugins(plugins)
     this.theme = theme || {}
     this.defaultTyping =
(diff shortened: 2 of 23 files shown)


packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
```diff
@@ -56,7 +57,15 @@ export class DynamoDBDataProvider implements DataProvider {
...
   // @ts-ignore
-  async getUsers(limit = 10, offset = 0): Promise<User[]> {} // TODO: Implement
+  async getUsers(limit = 10, offset = 0): Promise<User[]> {
+    // TODO: offset really needed here?
+    const result = await this.userEventsTable.scan({
+      filters: { attr: SORT_KEY_NAME, beginsWith: USER_PREFIX },
+      limit,
+    })
+    if (result.Count === 0) return []
+    return result.Items
```
marcpdw, line 67: If `result.Items` doesn't return a `[]` when it's an empty result, can't you go for a...
```
return result?.Items || []
```

packages/botonic-core/src/plugins.js
```diff
@@ -23,19 +23,22 @@ export async function runPlugins(
...
   response = null,
-  parsedResponse = null
+  parsedResponse = null,
+  dataProvider = null
 ) {
   for (const key in plugins) {
     const p = await plugins[key]
     try {
-      if (mode == 'pre') await p.pre({ input, session, lastRoutePath })
+      if (mode == 'pre')
```
marcpdw, line 32: Preference of `===`. If `botonic-dx` doesn't have it, we should consider to add it.

marcpdw (commented):
Here, tell me whether botonic-dx needs to be reviewed, please

packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
vanbasten17: you were right, it just returns an empty list, returning it directly.

packages/botonic-core/src/plugins.js
vanbasten17: fixed!

vanbasten17: added the rule in `botonic-eslint-config` package: https://eslint.org/docs/rules/eqeqeq
https://github.com/hubtype/botonic/pull/1764/commits/22697215a62c626661bb50fa6b7e37eb9cfcae4d

ericmarcos (changes requested):
I think it would better to pass the DataProvider instance as a parameter to the `input()` function, so we can pass the same instance we're using in the rest API or Websocket. We don't need the DataProvider out of `input()` and in general it's better to limit the responsibility of creating new instances to as few modules as possible.

packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
```diff
@@ -56,7 +57,14 @@ export class DynamoDBDataProvider implements DataProvider {
...
   // @ts-ignore
-  async getUsers(limit = 10, offset = 0): Promise<User[]> {} // TODO: Implement
+  async getUsers(limit = 10, offset = 0): Promise<User[]> {
+    // TODO: offset really needed here?
+    const result = await this.userEventsTable.scan({
+      filters: { attr: SORT_KEY_NAME, beginsWith: USER_PREFIX },
+      limit,
+    })
+    return result.Items
+  }
```
asastre, line 67: If `getUsers` has been called from an api request with pagination, `offset` is needed in order to return only the specific users from the desired page.

With this implementation you're only returning always the first users fixed by the `limit` value.
We should allow pagination here too using the `offset` argument.

packages/botonic-core/src/core-bot.js
```diff
@@ -54,7 +56,16 @@ export class CoreBot {
...
+      await runPlugins(
+        this.plugins,
+        'pre',
+        input,
+        session,
+        lastRoutePath,
+        undefined,
+        undefined,
+        this.dataProvider
+      )
```
asastre, line 68: Maybe it would be better to use named parameters using an object in order to avoid initializing these 2 `undefined` and to know what each argument refers to.

packages/botonic-eslint-config/index.js
```diff
@@ -77,6 +77,7 @@ module.exports = {
     'import/default': 'warn', // syntax "export = xxxx" is not supported
     'node/no-extraneous-import': 'warn', // otherwise it does not find ts-mockito if only defined in parent project
     'array-callback-return': 'error',
+    eqeqeq: 'error',
```
asastre, line 80: What is this!? :astonished:

asastre (changes requested):
Some things to consider/change before approving. 

Weren't all these typings already done (`AudioMessageEvent`, `ImageMessageEvent`...)?

asastre:
> I think it would better to pass the DataProvider instance as a parameter to the `input()` function, so we can pass the same instance we're using in the rest API or Websocket. We don't need the DataProvider out of `input()` and in general it's better to limit the responsibility of creating new instances to as few modules as possible.

I agree with @ericmarcos. This way there will be no need to add the dataProvider initialization in `my-botonic-app/bot/src/index.js`.

packages/botonic-api/src/data-provider/dynamodb-data-provider.ts
vanbasten17: we should check how this can be done with dynamoDB, leaving the TODO here as this should be reviewed but out of the scope of this PR.

packages/botonic-core/src/core-bot.js
vanbasten17: added a task for this, we have to be sure we don't break current functionality: https://linear.app/hubtype/issue/BOT-541/simplify-calls-to-loadrun-plugins-in-botoniccore

vanbasten17:
@ericmarcos , @asastre , now dataProvider is called through the `input` functions of the bot.

packages/botonic-core/src/types/models/events/message/media.d.ts
```diff
@@ -0,0 +1,18 @@
+import { BotonicMessageEvent, MessageEventTypes } from '.'
+import { WithButtons } from './buttons'
+export declare const MEDIA_TYPES: MessageEventTypes[]
+export interface MediaMessageEvent extends BotonicMessageEvent, WithButtons {
+  src: string
+}
+export interface AudioMessageEvent extends MediaMessageEvent {
```
asastre, line 7: Weren't all these typings already done (`AudioMessageEvent`, `ImageMessageEvent`, etc...)?

asastre (approved):
:rocket:

vanbasten17:
Closing this PR functionality it's already included here: https://github.com/hubtype/botonic/pull/1867, after app architecture redesign.

PR #1776, opened 2021-08-11
Full stack/fix prod acks and custom messages

vanbasten17:
## Description
* Fixed ACKs which were shown always with transparency when deployed to AWS.
* Fixed custom messages not capturing all possible props.

1. Fixed ACKs in prod:
Make use of 'draft', 'sent' and 'received' acks to handle UX of not yet sent messages.
For user messages: FRONT --> BACK
* When sending the enduser input, we set the ack as `draft`.
* At the very beginning of the `onMessage` logic, we send back to the webchat this event asap:
```
await send({
      action: 'update_message_info',
      id: frontendMsgId,
      ack: MessageEventAck.SENT,
})
```
* Then, after the event is commited to the db in `onMessage` logic, we execute the following logic to update the UX again webchat:
```
await send({
      action: 'update_message_info',
      id: frontendMsgId,
      ack: MessageEventAck.RECEIVED,
})
```
For bot messages: BACK --> FRONT
The only possibility if the bot has not crashed is to send back the message to the webchat with the `received` ack.

2. Fixed custom messages not capturing all possible props
Right now we were only parsing the `customTypeName` in `botonicOutputParser`, so to capture all the possible props that can be passed through a custom message it's better to encapsulate them into a single object `json`, (name can be discussed) which we can certainly now that will exist when storing it to DynamoDB.


asastre (approved):
LGTM! 
Although I would prefer to rename the `json` variable to a more explicit name like `customFields` or `customProps`. What do you think @vanbasten17 @ericmarcos? Maybe you have a better name for the variable than the ones I suggest :wink:

vanbasten17:
Closing this PR: changes included in https://github.com/hubtype/botonic/pull/1867.

PR #1808, opened 2021-08-25
refactor: Migrate @botonic/core to TS pt.2

vanbasten17:
## Description
* Moved all possible types declared previously manually in `types` directory within current files. Left some of them typed with any and some parts with `// @ts-ignore`, we can improve all of them eventually.
* Removed `types` folder as this was used in order to have `js` files as well as their types. Now it is not needed as the declarations are within the code and they are compiled to js along with declarations.
* Removed `types` attribute from package.json, it is resolved automatically if the "main" is pointing to a folder with declarations.

**NOTE:** Next step would be to add and improve router tests and try to split it. With this migration, it will be easier to centralize all necessary types in core for botonic 1.0.


packages/botonic-core/src/utils.ts
```diff
@@ -67,11 +73,12 @@ export const join = (...segments) => {
...
+  // @ts-ignore
   if (parts[0] === '') newParts.unshift('')
   // Turn back into a single string path.
   return newParts.join('/') || (newParts.length ? '/' : '.')
 }
 
 // A simple function to get the dirname of a path
 // Trailing slashes are ignored. Leading slash is preserved.
-export const dirname = path => join(path, '..')
+export const dirname = (path): string => join(path, '..')
```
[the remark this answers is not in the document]
vanbasten17: types added! thanks @asastre !

packages/botonic-core/src/core-bot.ts
```diff
@@ -37,20 +63,26 @@ export class CoreBot {
...
       ? null
       : new Router(
+          // @ts-ignore
           [...this.routes, ...this.defaultRoutes],
           this.inspector.routeInspector
         )
   }
 
-  getString(stringID, session) {
+  getString(stringID: string, session: Session): string {
```
marcpdw, line 72: When it is typed, do we need to call it `stringId`? I'd say it can be just `id: string`

packages/botonic-core/src/debug/inspector.ts
```diff
@@ -53,30 +74,29 @@ export class LogRouteInspector extends FocusRouteInspector {
...
     let name = obj.path || '<<no path defined>>'
     if (obj.action) {
       name = `'${name}' (to action '${obj.action.name}')`
     }
     if (obj.redirect) {
       name = `'${name}' (with redirect to '${obj.redirect}')`
     }
-
-    return name
+    return name as string
```
marcpdw, line 85: do you need to do this due to `obj.path`? All the rest are strings, so what about doing so ASAP, `obj.path as string`?

packages/botonic-core/src/utils.ts
```diff
@@ -36,26 +42,26 @@ export const isMobile = (mobileBreakpoint = 460) => {
   }
   return false
 }
-export function isFunction(o) {
+export function isFunction(o: any): boolean {
```
marcpdw, line 45: Is this `any` or `string`? If you are restricting it to `string`, the compiler will force you to invoke it only when using `typeof`, but maybe it's not what you want here 🙏

packages/botonic-core/src/router.ts
```diff
@@ -264,7 +306,7 @@ export class Router {
       if (value === undefined) {
         return false
       }
-      return matcher.exec(value)
+      return Boolean(matcher.exec(value))
```
marcpdw, line 309: To me, this is clearer than `!!`. High five here!

marcpdw (approved):
Huge progress made here, well done! I've added a few observations, so have a look before merging please!

Also, a couple of things to mention:
- ⭐    Good practise sorting constants and interface properties
- ⚠️    I've seen `==` in some places instead of `===`. Do you want to consider adding `eqeqeq` on `botonic-dx`? At least to give you a warning?

packages/botonic-core/src/debug/inspector.ts
vanbasten17: I see.. this comes from the argument typed as Route, which has typed path as a stringMatcher, that can be a regex:
```
export interface Route {
  ...
  path?: StringMatcher --> `RegExp | string | ((data: string) => boolean)`
```
hence the "trick". The better solution here pass by strictly reviewing all `@botonic/core` and also try to split complex functions like `processInput` into smaller pieces and with unit tests. I would leave it as is atm and reviewing eventually when working with this package.

packages/botonic-core/src/utils.ts
vanbasten17: it should be any, but now that I have reviewed it, this function is very generic for what it's doing (write it down for next iterations). As here we are only checking if the `routes` defined by the developer are written as an array of routes or as a function returning this routes, maybe we could call it `isFunctionalRouter` or similar and passing the entire `routes` as argument.

vanbasten17:
I think that we already added the `eqeqeq` rule in @botonic/dx, but @botonic/core still doesn't have `@botonic/dx` as a dev dependency (out of scope atm). I have noted it in order to do it for next iterations of the package. 👀

PR #1823, opened 2021-08-31
Contentful: Allow document as a follow-up

guinii:
- Added Document as an option of the follow-up entries of Text.
- Add converter for document
Support for botonic Document (PDF) components in Contentful plugin as follow-up

### DIFF
--- a/packages/botonic-plugin-contentful/src/cms/transform/message-content-filters.ts
+++ b/packages/botonic-plugin-contentful/src/cms/transform/message-content-filters.ts
@@ -4,12 +4,12 @@
 import {
   Button,
   Carousel,
+  Document,
   Element,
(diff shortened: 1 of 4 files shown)


vanbasten17 (approved):
LGTM! Check the import sorting warning (quick actions of VS Code) and go for it

asastre (approved):
As @vanbasten17 said, take a look at the imports that need to be sorted, the rest seems good to me too!
It would be nice to have a test in `tests/contentful/contents/text.test.ts` (with the content created in the Contentful QA account) to test that a text can contain a Document as a follow-up.

asastre:
Should this PR point to `master_lts` branch instead of `master`?

PR #1826, opened 2021-09-01
chore(core): cjs and esm distribution of core, enable @botonic/core tree-shaking

vanbasten17:
## Description

* Distribute `@botonic/core` package as  `/lib/cjs` and `lib/esm`. (ref: https://blog.logrocket.com/publishing-node-modules-typescript-es-modules/)
* Added `sideEffects` flag to false to facilitate webpack to tree-shake the library.

Distributing it with CJS will allow us to use it in environments like Node.js, whereas esm will allow us to use it in browser and applications using webpack.

Despite having the following two lines declared @botonic/core:
```
"main": "./lib/cjs/index.js",
"module": "./lib/esm/index.js",
```
when using webpack, the resolved library will depend on how we import the code.
**e.g.:**

```
import { isFunction } from '@botonic/core' // --> will load automatically /lib/esm version of the code.
console.log(isFunction)
```
by compiling this with webpack in production mode: `npx webpack` we end up with a minified file of **`4.0KB`**.
**reason:** ESM packages are tree-shakeable and will only bundle the `isFunction` method.

```
const { isFunction } = require('@botonic/core') // --> will load automatically /lib/cjs version of the code.
console.log(isFunction)
```
whereas by compiling this with webpack in production mode: `npx webpack` we end up with a minified file of **`216.0KB`**.
**reason:** CJS packages are not tree-shakeable and will bundle the `isFunction` along with the rest of the code exported in core's index.

### DIFF
--- a/packages/botonic-core/package.json
+++ b/packages/botonic-core/package.json
@@ -3,13 +3,14 @@
   "version": "0.19.0",
   "license": "MIT",
   "description": "Build Chatbots using React",
-  "main": "lib/index.js",
+  "main": "./lib/cjs/index.js",
+  "module": "./lib/esm/index.js",
   "scripts": {
     "test": "../../node_modules/.bin/jest --coverage",
     "prepare": "node ../../preinstall.js",
     "lint": "npm run lint_ci -- --fix",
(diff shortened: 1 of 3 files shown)


asastre (approved):
Very nice! :clap: :clap:

MarcosCA (approved):

PR #1845, opened 2021-09-08
Full stack/bump necessary packages to 1.0.0-dev version

vanbasten17:
**NOTE**: Review only first 4 commits. Last commit contains a new folder with development project structure.
## Description
* Bumping necessary packages for running development template to 1.0.0-dev version
* Fix minor issues (a way to abort localtunnel promise as they have an issue that makes them unacessible: https://github.com/localtunnel/localtunnel/issues/434) and safely accessing process.env (which is compiled in mode development because tree-shaking optimization is not done)
* Create a new dev-template to be pulled by `create botonic-app@dev` with all the current setup
* add `prepublishOnly` scripts to be run in order to not forget to build packages before they are published.
* Remove `name` in aws.Queue Pulumi component to avoid having collisions with other queue names within the same organization.


packages/botonic-cli/src/botonic-playground-service.ts
```diff
@@ -6,7 +6,19 @@ import { PlaygroundSessionInfo } from './interfaces'
...
 
 const PLAYGROUND_API_HOST = 'https://api.playground.botonic.io'
-
+/**
+ * Temporary solution to allow aborting promise after a certain timeout.
+ */
+function withTimeout<T>(
+  promise: Promise<T>,
+  errorMsg: string,
+  timeout = 2 * 1000
```
asastre, line 15: Specify the units in the variable's name -> `timeoutMs`?

packages/botonic-cli/src/botonic-playground-service.ts
```diff
@@ -24,7 +36,10 @@ export class PlaygroundService {
 
   async start() {
     try {
-      this.tunnel = await localtunnel({ port: this.port })
+      this.tunnel = await withTimeout<localtunnel.Tunnel>(
+        localtunnel({ port: this.port }),
+        'Tunneling service not reachable.'
+      )
```
asastre, line 42: Maybe 2 seconds are not enough for `localtunnel` to be executed? How about increasing a little bit the timeout?

packages/botonic-core/package.json
```diff
@@ -1,13 +1,14 @@
 {
   "name": "@botonic/core",
-  "version": "0.19.0",
+  "version": "1.0.0-dev.0",
```
asastre, line 3: Should all packages be in version `1.0.0-dev.1`?

packages/botonic-cli/src/botonic-playground-service.ts
vanbasten17: changed! 👍

packages/botonic-cli/src/botonic-playground-service.ts
vanbasten17: I've tested it with 1 second (now localtunnel works again) works pretty good. Leaving it to 2 seconds, it should be more than enough.

packages/botonic-core/package.json
vanbasten17: Atm I and as we are going to release more often dev releases, I prefer to leave it like this, because in this way packages that are dependant on this one don't need to have anything changed as they are already pointing to 'dev' version.

asastre: Seems good then!

asastre (approved):

PR #1867, opened 2021-09-10
Full stack/refactor data persistence pt2

vanbasten17:
## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-react/src/experimental/webchat-app.jsx
```diff
@@ -141,7 +141,11 @@ export class WebchatApp {
   onServiceEvent(event) {
     if (event.action === 'connectionChange')
       this.webchatRef.current.setOnline(event.online)
-    else if (event.action === 'update_message_info')
+    // TODO: Temporary solution, decide how we will send these events in next iterations
+    else if (event.message.action === 'update_message_info') {
+      const { message } = event.message
+      this.updateMessageInfo(message.id, message)
+    } else if (event.action === 'update_message_info')
```
asastre, line 148: Should this be if be removed? Can `'update_message_info'` come from 2 different sides (`event.action` and `event.message.action`)?

Why not using `event.message.type` instead? I see that we're using multiple attributes to define the type/action and I think it's a little bit confusing, maybe would be a good idea to set them just in one attribute.

asastre (commented):
Looks good! Just a few comments before approving.

packages/botonic-core/package.json
```diff
@@ -37,7 +37,8 @@
     "fast-xml-parser": "^3.19.0",
     "html-entities": "^2.3.2",
     "node-json-db": "^1.3.0",
-    "pusher-js": "^5.1.1"
+    "pusher-js": "^5.1.1",
```
ericmarcos, line 40: we can remove Pusher dependency

packages/botonic-core/src/core-bot.ts
```diff
@@ -89,15 +93,34 @@ export class CoreBot {
...
-      // pass dataProvider
-      await runPlugins(this.plugins, 'pre', input, session, lastRoutePath)
+      await runPlugins(
+        this.plugins,
+        'pre',
+        input,
+        session,
+        lastRoutePath,
+        undefined,
+        undefined,
```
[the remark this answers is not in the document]
ericmarcos: maybe we can take the chance to refactor this and pass an object args

ericmarcos (approved):

PR #1881, opened 2021-09-15
Directus: initial structure

AlbertGom:
## Description
Changes done:

- Added the initial folder structure of the Directus plugin.
- Added the cms models definition.

## Context
New botonic plugin that will integrate the [Directus](https://directus.io/) cms.
For now text, button and image will be the models that will be used in the plugin.

In plugin.ts file:
There is the plugin constructor. For now the DirectusOptions are the credentials used to authenticate through the directus sdk. The token is a static token defined in the frontend.

In cms/contents.ts file:

- Created an abstract class called Content that stores the common information shared by the three subclasses.(Text,Button,Image)

- Added the three subclasses that will store specific information of each model.

In cms/cms.ts file:

- Added the supported content types and locales.
- Created an interface called CMS that will be used to access the contents through an id and a context (locale).

In directus/contents folder:
There will be the code to convert the received data from contentful to a cms content model.

In directus/delivery folder:
There will be the code that will acces to the data through directus sdk.

In render folder:
There will be the class that will be used to convert the cms content model to content that botonic/react will recognize and will be able to render.

## Testing

The pull request...

- doesn't need tests because is the initial structure of the plugin.


packages/botonic-plugin-directus/src/cms/contents.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export class Text extends Content {
+  readonly text: string
+  readonly buttons?: Button[]
+  constructor(opt: { common: CommonFields; text: string; buttons?: Button[] }) {
+    super(opt.common)
+    this.text = opt.text
+    this.buttons = opt.buttons
+  }
+}
```
AinaVendrell, line 32: I think `buttonsStyle` is missing

packages/botonic-plugin-directus/src/cms/cms.ts
```diff
@@ -0,0 +1,21 @@
+import { Button, Text, Image } from './contents'
+
+export enum ContentType {
+  TEXT = 'text',
+  BUTTON = 'button',
+  IMAGE = 'image',
+}
```
AinaVendrell, line 7: Now there are not that many contents, but in the future it may be useful to split them in `SubContentType` and `TopContentType` as we do in Contentful
```
export enum SubContentType {
  BUTTON = 'button',
  ELEMENT = 'element',
}

export type ContentType = TopContentType | SubContentType
export const ContentType = { ...TopContentType, ...SubContentType }
```

packages/botonic-plugin-directus/src/cms/contents.ts
```diff
@@ -0,0 +1,50 @@
+export interface CommonFields {
+  id: string
+  name: string
+  shortText?: string
+  keywords?: string[]
+  followup?: Text | Image
```
AinaVendrell, line 6: Related to the previous comment, here you could do `followup?: TopContentType`

packages/botonic-plugin-directus/src/cms/contents.ts
```diff
@@ -0,0 +1,50 @@
...
+  constructor(opt: { common: CommonFields; text: string; buttons?: Button[] }) {
+    super(opt.common)
+    this.text = opt.text
+    this.buttons = opt.buttons
+  }
+}
+
+export class Button extends Content {
+  readonly text: string
+  readonly target: string
```
asastre, line 36: `target` should be fixed to a string or should be a union of types like `Text` or `Image` or `string`? 
I'm asking this because I don't know what value will be placed here if the button is targeting a `Text` or an `Image`.

packages/botonic-plugin-directus/src/cms/contents.ts
```diff
@@ -0,0 +1,50 @@
...
+  constructor(opt: { common: CommonFields; text: string; target: string }) {
+    super(opt.common)
+    this.text = opt.text
+    this.target = opt.target
+  }
+}
+
+export class Image extends Content {
+  readonly image: string
```
asastre, line 45: Should `src` or `url` be a better name for this attribute than `image`? This attribute is representing the URL/path of the `Image`, right?

packages/botonic-plugin-directus/src/cms/contents.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  get id(): string {
+    return this.common.id
+  }
+}
+
+export class Text extends Content {
+  readonly text: string
+  readonly buttons?: Button[]
+  constructor(opt: { common: CommonFields; text: string; buttons?: Button[] }) {
```
asastre, line 27: I would define specific interfaces for the fields of every content, for example:
```
export interface TextFields {
    common: CommonFields
    text: string
    buttons?: Button[]
}
```
And then the constructor would be like this:
```
constructor(opt: TextFields) {
```

packages/botonic-plugin-directus/src/plugin.ts
```diff
@@ -0,0 +1,23 @@
...
+export default class BotonicPluginDirectus {
+  readonly cms: cms.CMS
+
+  constructor(opt: DirectusOptions) {
+    this.cms = new Directus(opt)
+  }
+
+  pre(r: { input: any; session: any; lastRoutePath: any }) {}
+
+  post(r: { input: any; session: any; lastRoutePath: any; response: any }) {}
```
asastre, line 22: Fill the types with the correct ones, try to not use `any` as a type.

packages/botonic-plugin-directus/src/plugin.ts
```diff
@@ -0,0 +1,23 @@
+import * as cms from './cms'
+import { Directus } from './directus/directus'
```
asastre, line 2: Why `src/directus/directus` is not in this PR?

asastre (commented):
Some comments/doubts

PR #1900, opened 2021-09-21
Contentful: custom content model

AlbertGom:
## Description
Create `custom` content

## Context
There was no custom content model where you can add custom fields without having to be defined in the plugin.

 Created a new component that has id,name and fields. Fields field is a record that has the custom fields defined in contentful.

## Testing

The pull request...

- has unit tests


Iru89 (approved):

asastre (approved):

asastre:
Although I have approved the PR @AlbertGom take a look at the tests, they're failing!

asastre:
The changes done before the last commit allowed to fetch custom content from Contentful only if the content model was named 'custom'.
With the changes done in the [last commit ](https://github.com/hubtype/botonic/pull/1900/commits/3b9d07d61c3a69a844eca54cb7203b6da6fbb0ff) allow fetching all type of content models different than the known ones (text, image, button, etc.).
@AlbertGom can you review this last commit?

AinaVendrell (approved):
🚀

packages/botonic-plugin-contentful/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "scripts": {
     "build": "rm -rf lib && ../../node_modules/.bin/tsc",
+    "build:watch": "../../node_modules/.bin/tsc --watch",
```
vanbasten17, line 4: in order to not duplicate the same command for this, you can do:
`"build:watch": "npm run build -- --watch"` to pass variables through a npm script

vanbasten17 (approved):

packages/botonic-plugin-contentful/package.json
asastre: I thought about it too, but in this case, there's no need to remove the `lib` folder when executing the build with the `--watch` flag, that's why I was "duplicating" the command.

vanbasten17: as the watch process will be running in long term, I think that lib won't be removed every time (I haven't tested it, but I think it behaves like this). Not an important thing. Up to you!

PR #2030, opened 2021-11-11
feat(react): add imagePreviewer property in theme and allow rendering custom components in webchat

vanbasten17:
## Description
Added the following behaviors in webchat: 
1. If  a React component is defined under `theme.imagePreviewer`, it will be rendered when an image (`Image` component) within webchat is clicked is clicked. Otherwise, it will behave as before. **It is not provided by default by the framework in order to not increase the bundle sizes of all bots.**
2. Allow rendering custom components that will be injected in webchat.

## Context
We didn't have a way to add a previewer for images.

Created an external package to be used within this functionality. Example providen below

Screenshot 2021-11-12 at 10 20 58

Desktop:

Screenshot 2021-12-21 at 17 27 14

Mobile:
Screenshot 2021-12-21 at 17 28 02

Rendering a custom component with Botonic API:

`webchat/index.js`
```
export const webchat = {
  onInit: app => app.open(),
  onOpen: app => {
    app.renderCustomComponent(
      <ImagePreviewer
        isPreviewerOpened={true}
        src={'https://www.scalestatistics.com/uploads/3/0/4/1/30413390/7622723_orig.jpg'}
        closePreviewer={() => Botonic.unmountCustomComponent()} />
    )
  },
}
```

Defining your custom ImagePreviewer:
```
import { ImagePreviewer } from '@hubtype/image-previewer'

export const webchat = {
  theme: {
    imagePreviewer: ImagePreviewer
  }
}
```

## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-react/src/webchat/components/portal-modal/content.jsx
```diff
@@ -0,0 +1,14 @@
+import styled from 'styled-components'
+
+export const ModalContent = styled.div`
```
guinii, line 3: Add `display: flex`?

guinii (changes requested):

Iru89 (approved):

PR #2127, opened 2021-12-29
Feat/add environment manager #FLO-107

AinaVendrell:
## Description
- Add `ManageEnvironment` class

## Context
- In order to publish a flow with the flow builder, we need to have access to the available environments and apply changes such as:
   - Create or delete environments 
   - Change the environment alias target

## Testing

The pull request...

- has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/contentful/manage/manage-environment.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/manage/manage-environment.ts
@@ -1,0 +1,84 @@
+import { createClient } from 'contentful-management'
+import { ClientAPI } from 'contentful-management/dist/typings/create-contentful-api'
+import {
+  Collection,
+  Environment,
+  EnvironmentAlias,
+  EnvironmentProps,
+  Space,
+} from 'contentful-management/dist/typings/export-types'
+
+import { ContentfulOptions } from '../../plugin'
+
+export interface ManageCmsEnvironment {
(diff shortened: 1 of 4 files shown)


AinaVendrell:
I've just seen that some tests are failing (in the master branch too) because: 
```
Loading content failed on content 'text' with id '0.1256731741416981'. The resource could not be found.
```
But I've just checked Contentful and the content does exist. Any idea on why is the error happening?

packages/botonic-plugin-contentful/src/contentful/manage/manage-environment.ts
```diff
@@ -0,0 +1,72 @@
...
+  public async deleteEnvironment(environmentId: string): Promise<void> {
+    const environment = await (await this.space).getEnvironment(environmentId)
+    return environment.delete()
+  }
+
+  public async createEnvironmentWithId(
+    environmentId: string
+  ): Promise<Environment> {
+    return (await this.space).createEnvironmentWithId(environmentId, {
+      name: environmentId,
```
vanbasten17, line 53: should the name be the same as the `environmentId` or it could be another name?

vanbasten17 (approved):
Looking good!!

packages/botonic-plugin-contentful/src/contentful/manage/manage-environment.ts
AinaVendrell: usually it will be the same

PR #2171, opened 2022-01-17
Test/create manage contentful environment #BOT-644

AinaVendrell:
## Description
- Move the manage contentful test to a new contentful environment
- Skip failing Profiler test

## Context
- Some tests were failing due to new contents created in the manage contentful test. These tests create and delete entries every time they are executed, but sometimes the new entries are not deleted and they can affect the other tests.

- Create a new environment called `manage-contentful` that will be used in the ManageContentful tests. It should only contain one asset. 

## Testing

The pull request...

- has unit tests
- has integration tests


vanbasten17 (approved):
Nice work here!! Thanks @AinaVendrell ! 💯

PR #2231, opened 2022-04-28
feat(core): Handoff improvements (autoAssignOnWaiting and forceAssignIfNotAvailable)

asastre:
## Description
Added 2 new functions in `HandoffBuilder`:
- `withForceAssignIfNotAvailable`: `True` by default. Allow to not assign the case (and set it in waiting status) to the selected agent if he is not available.
- `withAutoAssignOnWaiting`: `False` by default. Allow to assign the case automatically to an agent when the case is in the waiting queue.

## Context
We want to be able to:
- Not assign cases to away/busy agents (`forceAssignIfNotAvailable`)
- Allow the agents to not assign the cases in waiting queue manually and let the system assign them automatically.

- `forceAssignIfNotAvailable`: `new HandoffBuilder(session).withForceAssignIfNotAvailable(false)`
- `autoAssignOnWaiting`: `new HandoffBuilder(session).withAutoAssignOnWaiting(true)`

## Testing

The pull request has unit tests.

### DIFF
--- a/packages/botonic-core/src/handoff.ts
+++ b/packages/botonic-core/src/handoff.ts
@@ -183,7 +183,7 @@ async function _humanHandOff(
   onFinish: string,
   agentEmail = '',
   agentId = '',
-  forceAssignIfNotAvailable: boolean,
+  forceAssignIfNotAvailable = true,
   autoAssignOnWaiting = false,
   caseInfo = '',
   note = '',
@@ -191,6 +191,9 @@ async function _humanHandOff(
   shadowing = false
 ) {
   const params: HubtypeHandoffParams = {}
+
+  params.force_assign_if_not_available = forceAssignIfNotAvailable
+
   if (queueNameOrId) {
     params.queue = queueNameOrId
   }
@@ -199,9 +202,6 @@ async function _humanHandOff(
   }
   if (agentId) {
     params.agent_id = agentId
-  }
-  if (forceAssignIfNotAvailable !== undefined) {
-    params.force_assign_if_not_available = forceAssignIfNotAvailable
   }
   if (autoAssignOnWaiting) {
     params.auto_assign_on_waiting = autoAssignOnWaiting
--- a/packages/botonic-core/src/handoff.ts
+++ b/packages/botonic-core/src/handoff.ts
@@ -63,6 +63,7 @@ export class HandOffBuilder {
   _email: string
   _agentId: string
   _forceAssignIfNotAvailable: boolean
+  _autoAssignOnWaiting: boolean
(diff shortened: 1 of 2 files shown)


packages/botonic-core/src/handoff.ts
```diff
@@ -167,6 +183,8 @@ async function _humanHandOff(
   onFinish: string,
   agentEmail = '',
   agentId = '',
+  forceAssignIfNotAvailable: boolean,
```
MarcosCA, line 186: should be true by default? This is how it works now no?

packages/botonic-core/src/handoff.ts
```diff
@@ -182,6 +200,12 @@ async function _humanHandOff(
   if (agentId) {
     params.agent_id = agentId
   }
+  if (forceAssignIfNotAvailable !== undefined) {
```
MarcosCA, line 203: and then this could be if(forceAssignIfNotAvailable)

packages/botonic-core/src/handoff.ts
JianChen-IT: yes

asastre: I've not set any default value here because if it's set to `true`, then the `force_assign_if_not_available` parameter will be sent always in all handoffs and since the default value in the backend is also `true`, I preffer not to send the parameter if it's not defined by the developer.

@MarcosCA that's why I added the condition of `if (forceAssignIfNotAvailable !== undefined)` (regarding your next comment) in order to add the parameter only if the developer has set a value to it.

Also, sending always this new parameter  in all handoff actions (even if it's not set by the developer) it may break tests done in some bots.

packages/botonic-core/src/handoff.ts
asastre: @MarcosCA I've answered this comment in the previous one.

packages/botonic-core/src/handoff.ts
ericmarcos: Even if the default is true in the backend, I'd rather set the default here explicitly as well.
@asastre I don't know what tests this might break, but we can't make design choices based on some tests we have somewhere. We need to make the right choice looking forward and fix any tests if needed.

asastre: @ericmarcos @MarcosCA @JianChen-IT As you've suggested, I've set `true` as the default value for `forceAssignIfNotAvailable` attribute.

ericmarcos (approved):

AinaVendrell (approved):

PR #2241, opened 2022-06-02
Feat/add destroy method on webchat app

AinaVendrell:
## Description
- Add a `destory` method on the webchat app that will destory the pusher of the hubtype service, unmount the webchat component and remove the storage.
- Define the default storage values in the `webchat-app` file so the methods in have access to the final values.
- Fix some types.

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-react/src/webchat-app.jsx
+++ b/packages/botonic-react/src/webchat-app.jsx
@@ -312,8 +312,8 @@ export class WebchatApp {
         enableAttachments={enableAttachments}
         enableUserInput={enableUserInput}
         enableAnimations={enableAnimations}
-        storage={storage}
-        storageKey={storageKey}
+        storage={this.storage}
+        storageKey={this.storageKey}
         defaultDelay={defaultDelay}
         defaultTyping={defaultTyping}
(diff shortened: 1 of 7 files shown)


packages/botonic-react/src/webchat-app.jsx
```diff
@@ -286,9 +286,9 @@ export class WebchatApp {
     enableAnimations = enableAnimations || this.enableAnimations
     defaultDelay = defaultDelay || this.defaultDelay
     defaultTyping = defaultTyping || this.defaultTyping
-    storage = storage || this.storage
-    storageKey = storageKey || this.storageKey
     server = server || this.server
+    this.storage = storage || this.storage
+    this.storageKey = storageKey || this.storageKey
```
asastre, line 291: The props of `storage` and `storageKey` sent to the `Webchat` component have to be changed to `this.storage` and `this.storageKey`, otherwise no storage will be created unless stated at runtime.

asastre (changes requested):
Just one change have to be made, the rest LGTM.

asastre (approved):
:rocket:

PR #2249, opened 2022-06-29
feat(plugin-contentful): add document as a followup

AlbertGom:
## Description
Allow document content model to be used as a followup.

## Context
Document content model was not allowed as a followup.

## Testing

The pull request...

- doesn't need tests


AinaVendrell (changes requested):
Check lint errors

AinaVendrell (approved):

PR #2258, opened 2022-07-12
refactor(react): Allow storage other than localStorage and sessionStorage

asastre:
## Description
Refactor the `useStorageState` function to be more flexible and allow other types of storage than `localStorage` and `sessionStorage`.

## Context
There was no possibility to create a custom in-memory storage for Botonic.

Instead of using the `useLocalStorage` and `useSessionStorage` hooks from [react-use-storage](https://github.com/leny/react-use-storage) package we've copied the `useStorage` function of this same package to be more flexible and allow other type of storages.

## Testing

The pull request has unit tests.

### DIFF
--- a/packages/botonic-react/tests/webchat/storage.test.jsx
+++ b/packages/botonic-react/tests/webchat/storage.test.jsx
@@ -3,6 +3,7 @@ import TestRenderer, { act } from 'react-test-renderer'
 
 import { WEBCHAT } from '../../src/constants'
 import { Webchat } from '../../src/webchat/webchat'
+import { InMemoryStorage } from '../helpers/in-memory-storage'
 
 describe('TEST: storage', () => {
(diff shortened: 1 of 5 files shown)


ericmarcos (approved):
Nice!!

packages/botonic-react/tests/webchat/storage.test.jsx
```diff
@@ -56,6 +57,23 @@ describe('TEST: storage', () => {
     expect(sessionStorage.getItem('botonicState')).not.toBeNull()
   })
 
+  it('Stores botonicState in the memory storage', async () => {
```
manuelfidalgo, line 60: How about renaming the test to something like "botonicState can be stored in the in-memory storage and retrieved afterwards?

manuelfidalgo (approved):

PR #2260, opened 2022-07-25
Added one Useful Link

bikashdaga:
Have Added Scaler Topics JavaScript Page under JavaScript / TypeScript Styleguide. I think this will help all the learners with great sources for learning JavaScript basics. Kindly Please check the suggested changes.

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,7 +57,7 @@ For the sake of clarity and organization, we follow [this guideline](https://hac
 
 ### JavaScript / TypeScript Styleguide
 
-All JavaScript must adhere to [JavaScript Standard Style](https://standardjs.com/).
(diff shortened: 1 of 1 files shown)


manuelalferez:
Hi @bikashdaga, first of all thanks for the contribution. After reading your suggestion, I think this PR moves away from how documentation documents should be:
- They should have the minimal of references 
- Those resources cannot be a debt for the future. In other words, the link and the content to which the link redirects must not change

This is why most of the links redirect to an "unique source of truth"; the official documentation of the tool / programming language / etc, to which the documentation refer to.

In addition, documentation of a project explains the project itself. It should not have the jurisdiction to teach about the languages that the project is using.

ericmarcos:
Spam

PR #2265, opened 2022-09-20
Feature: Add NonSense prediction to hubtype babel plugin#DAT-332

elozano98:
## Description

The inference should return also if the input text has sense or not.

## Context

We need a solution for detecting when the input text is totally random (`asdfasdf`).

Add random text detection during inference in backend.


elozano98:
TODO: `botonic-core` and `botonic-plugin-hubtype-babel` release.

castledom04 (approved):

JianChen-IT (approved):

PR #2273, opened 2022-10-13
Botonic Plugin Contentful: handoff messages as text references

AlbertGom:
## Description
Modifed handoff fields: message and failMessage to be able to have them as Text references and not only strings.

## Context
This change will simplify the bots since we won't need to do another call to have these Text messages, they will now be in the handoff content.

## Testing

The pull request...

- has unit tests


packages/botonic-plugin-contentful/src/cms/contents.ts
```diff
@@ -488,8 +488,8 @@ export class Handoff extends TopContent {
   constructor(
     readonly common: CommonFields,
     readonly onFinish: OnFinish,
-    readonly message?: string,
-    readonly failMessage?: string,
+    readonly message?: Text | string,
+    readonly failMessage?: Text | string,
```
Iru89, line 492: In this line you say that it can be a string but I don't see in the rest of the code that a string is allowed for the message and failMessage fields. 
In the HandoffBuilder these fields should be allowed to be Text or string, right?

AlbertGom: These messages come from the handoff object. In the handoff delivery class when the handoff messages are received I check if they are a string, otherwise I put them as references.

Iru89 (approved):

AinaVendrell (approved):

PR #2281, opened 2022-10-26
Fixing Image previewer

manuelalferez:
## Description

In order to be able to display in full size an image shared within the webchat an image previewer was implemented in Botonic in order to display a modal in the middle of the screen with the image once it was clicked.

The problem is that when we tried the full solution we were getting an error of "Invalid hook call" (https://reactjs.org/warnings/invalid-hook-call-warning.html)

What is happening because we are calling the image previewer as a function instead of as a component.

Clicking on an image in the Bot opens a pop up with the image where we can zoom in and out, click outside the image to close the pop up or through the top bar with menus.

### DIFF
--- a/packages/botonic-react/src/components/image.jsx
+++ b/packages/botonic-react/src/components/image.jsx
@@ -31,7 +31,6 @@ export const Image = props => {
     null
   )
   if (isBrowser()) {
-    console.log(ImagePreviewer)
     content = (
       <>
         <StyledImage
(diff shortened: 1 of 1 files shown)


packages/botonic-react/src/components/image.jsx
```diff
@@ -26,25 +26,27 @@ export const Image = props => {
   const closePreviewer = () => setIsPreviewerOpened(false)
 
   const { getThemeProperty } = useContext(WebchatContext)
-  const imagePreviewer = getThemeProperty(
+  const ImagePreviewer = getThemeProperty(
     WEBCHAT.CUSTOM_PROPERTIES.imagePreviewer,
     null
   )
   if (isBrowser()) {
+    console.log(ImagePreviewer)
```
asastre, line 34: Remove `console.log`

asastre (approved):
Remove the console.log and you're good to go!

AinaVendrell (approved):

PR #2285, opened 2022-11-22
feat(react): Wrap all message buttons into a div

asastre:
## Description
Wrap all `<button>` elements of a message within a `<div>` element.

## Context
All message buttons are rendered as `<button>` elements alongside the `<div>` element of the message content, and to be able to apply styles to all buttons as a block (not only to the button element) a wrapper element is needed.

Add the class `botonic-buttons-container` to this new wrapper `div` to be able to change its style easily.

## Testing

### DIFF
--- a/packages/botonic-react/src/components/message.jsx
+++ b/packages/botonic-react/src/components/message.jsx
@@ -313,7 +313,9 @@ export const Message = props => {
               ) : (
                 <BlobText blob={blob}>{textChildren}</BlobText>
               )}
-              {buttons}
+              {!!buttons.length && (
+                <div className='message-buttons-container'>{buttons}</div>
+              )}
               {Boolean(blob) && hasBlobTick() && getBlobTick(6)}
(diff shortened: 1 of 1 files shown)


packages/botonic-react/src/components/message.jsx
```diff
@@ -313,7 +313,9 @@ export const Message = props => {
               ) : (
                 <BlobText blob={blob}>{textChildren}</BlobText>
               )}
-              {buttons}
+              {!!buttons.length && (
```
Iru89, line 316: buttons will always be an array, I think it is not necessary to negate twice (!!)

Iru89 (approved):

castledom04 (approved):

packages/botonic-react/src/components/message.jsx
asastre: Yes, buttons will always be an array but we need to convert the length value to boolean because otherwise the length value will be displayed in the component

PR #2288, opened 2022-12-02
Feature: Flag to enable has_sense detection on plugin-hubtype-babel

elozano98:
## Description

Added a flag that indicates if `has_sense` prediction should be applied.

The has_sense prediction is not enabled by default. If needed, it should be enabled by using the new flag. @asastre

### DIFF
--- a/packages/botonic-plugin-hubtype-babel/src/hubtype-babel-api-service.ts
+++ b/packages/botonic-plugin-hubtype-babel/src/hubtype-babel-api-service.ts
@@ -1,4 +1,4 @@
-import axios from 'axios'
+import axios, { AxiosResponse } from 'axios'
 
 export class HubtypeBabelApiService {
   constructor(
@@ -6,7 +6,11 @@ export class HubtypeBabelApiService {
     readonly host: string = 'https://api.hubtype.com'
   ) {}
 
-  async inference(text: string, accessToken: string, hasSenseEnabled: boolean) {
+  async inference(
+    text: string,
+    accessToken: string,
+    includeHasSense: boolean
+  ): Promise<AxiosResponse> {
     return await axios({
       method: 'POST',
       url: `${this.host}/v1/babel/projects/${this.projectId}/inference/`,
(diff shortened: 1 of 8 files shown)


packages/botonic-plugin-hubtype-babel/src/index.ts
```diff
@@ -45,15 +47,22 @@ export default class BotonicPluginHubtypeBabel implements Plugin {
...
         request.input.intent = response.data['intents'][0]['label']
         request.input.confidence = response.data['intents'][0]['confidence']
         request.input.intents = response.data['intents'].map(x => ({
           intent: x['label'],
           confidence: x['confidence'],
         }))
-        request.input.hasSense = response.data['has_sense']
+
+        if (this.hasSenseEnabled) {
+          request.input.hasSense = response.data['has_sense']
```
gbarba, line 64: so now the `input.hasSense` might be there or not, maybe this is undesirable and we prefer to put a `true` if no has_sense is provided by endpoint? or type properly this `input` to indicate this field might not exists

packages/botonic-plugin-hubtype-babel/src/hubtype-babel-api-service.ts
```diff
@@ -6,15 +6,15 @@ export class HubtypeBabelApiService {
     readonly host: string = 'https://api.hubtype.com'
   ) {}
 
-  async inference(text: string, accessToken: string) {
+  async inference(text: string, accessToken: string, hasSenseEnabled: boolean) {
```
gbarba, line 9: I don't think the `Enabled` is a good naming but better `includeHasSense` (see https://github.com/metis-ai/hubtype-backend/pull/1270/files#r1038188933).
also add a return type whith is an object with `intents` and, **optionally** `has_sense` boolean

gbarba (changes requested):

gbarba (approved):

elozano98:
Need to release new `botonic-core` and `botonic-pluguin-hubtype-babel`.

PR #2298, opened 2023-02-10
Contentful - Plugin: include references to content method

AlbertGom:
## Description
Include number of references to include in the content method.

## Context
When using the content method to get a handoff content, it was failling because the schedule was not being sent.

Include the number of references to include. Set to 6 as we don't have any content with more than 4 references.

## Testing

The pull request...

- has unit tests

### DIFF
--- a/packages/botonic-plugin-contentful/src/contentful/cms-contentful.ts
+++ b/packages/botonic-plugin-contentful/src/contentful/cms-contentful.ts
@@ -9,6 +9,7 @@ import {
   Context,
   DateRangeContent,
   DEFAULT_CONTEXT,
+  DEFAULT_REFERENCES_TO_INCLUDE,
   PagingOptions,
   ScheduleContent,
   TopContent,
@@ -238,7 +239,7 @@ export class Contentful implements cms.CMS {
   async content(
     id: string,
     context = DEFAULT_CONTEXT,
-    referencesToInclude = 6
+    referencesToInclude = DEFAULT_REFERENCES_TO_INCLUDE
   ): Promise<Content> {
     const entry = await this._delivery.getEntry(id, context, {
       include: referencesToInclude,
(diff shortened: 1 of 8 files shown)


Iru89 (approved):

packages/botonic-plugin-contentful/src/contentful/cms-contentful.ts
```diff
@@ -236,7 +236,10 @@ export class Contentful implements cms.CMS {
   }
 
   async content(id: string, context = DEFAULT_CONTEXT): Promise<Content> {
-    const entry = await this._delivery.getEntry(id, context)
+    const REFERENCES_TO_INCLUDE = 6
+    const entry = await this._delivery.getEntry(id, context, {
+      include: REFERENCES_TO_INCLUDE,
+    })
```
asastre, line 242: I think this parameter should be configurable so maybe the best option would be to pass this parameter as an input parameter for the `content` function making `6` as a default value.

AlbertGom: Yes, better this way, changed!

packages/botonic-plugin-contentful/src/contentful/cms-contentful.ts
```diff
@@ -235,10 +235,13 @@ export class Contentful implements cms.CMS {
     )
   }
 
-  async content(id: string, context = DEFAULT_CONTEXT): Promise<Content> {
-    const REFERENCES_TO_INCLUDE = 6
+  async content(
+    id: string,
+    context = DEFAULT_CONTEXT,
+    referencesToInclude = 6
```
asastre, line 241: Sorry to be so picky, but I think that it's better to define a `DEFAULT_REFERENCES` constant and set that constant to `6` instead of setting the default value to `6` in all methods.

asastre (changes requested):

packages/botonic-plugin-contentful/src/contentful/cms-contentful.ts
AlbertGom: applied!

asastre (approved):

PR #2308, opened 2023-03-06
fix: Webviews not redirecting correctly on WhatsApp Cloud #PLA-1598

Khaeshah:
## Description
Webviews were not redirecting correctly when sent through WhatsApp Cloud API. With this change, the issue is fixed. Linked to [this PR](https://github.com/metis-ai/hubtype-backend/pull/1373).
 

## Context
Use the unformatted phone number instead of the imp_id as WhatsApp Cloud API does not follow the same rules as ordinary WhatsApp

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-react/src/webview.jsx
+++ b/packages/botonic-react/src/webview.jsx
@@ -43,7 +43,8 @@ class App extends React.Component {
     const provider = this.state.session.user.provider
     const impId = this.state.session.user.imp_id
     if (provider === PROVIDER.WHATSAPP) {
-      location.href = 'https://wa.me/' + impId
+      const phone_number = this.state.session.user.unformatted_phone_number
+      location.href = 'https://wa.me/' + phone_number
     }
     if (provider === PROVIDER.TELEGRAM) {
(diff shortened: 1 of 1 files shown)


packages/botonic-react/src/webview.jsx
```diff
@@ -43,7 +43,8 @@ class App extends React.Component {
     const provider = this.state.session.user.provider
     const impId = this.state.session.user.imp_id
     if (provider === PROVIDER.WHATSAPP) {
-      location.href = 'https://wa.me/' + impId
+      const phone_number = this.state.session.user.unformatted_phone_number
```
Iru89, line 46: Will we always have this field to read the bot number or only when it is a WhatsApp cloud?

Iru89 (approved):

packages/botonic-react/src/webview.jsx
Khaeshah: this is a particular case where it will be populated only on whatsapp, other providers will get this as null

PR #2314, opened 2023-03-15
Feat: custom user input buttons

Iru89:
## Description
This PR adds the feature to define in the bot the icons to send an emoji and attach a file as you can do with the send text button by declaring a custom component

## Context
I didn't want to modify more things but I think the css should also be refactored to reduce the number of nested divs with fixed paddings to separate the elements. As you can see in the image the custom buttons are not well aligned.

Before change icons
Captura de Pantalla 2023-03-15 a las 12 10 45
After change icons
Captura de Pantalla 2023-03-15 a las 12 11 40

You can define custom components for each button like this
Captura de Pantalla 2023-03-15 a las 12 32 19

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-react/src/webchat/components/attachment.jsx
+++ b/packages/botonic-react/src/webchat/components/attachment.jsx
@@ -1,21 +1,34 @@
-import React from 'react'
+import React, { useContext } from 'react'
 
 import AttachmentIcon from '../../assets/attachment-icon.svg'
-import { ROLES } from '../../constants'
+import { ROLES, WEBCHAT } from '../../constants'
+import { WebchatContext } from '../../contexts'
 import { Icon, IconContainer } from './common'
 
-export const Attachment = ({ onChange, accept }) => (
-  <IconContainer role={ROLES.ATTACHMENT_ICON}>
-    <label htmlFor='attachment' style={{ marginTop: 4 }}>
-      <Icon src={AttachmentIcon} />
-    </label>
-    <input
-      type='file'
-      name='file'
-      id='attachment'
-      style={{ display: 'none' }}
-      onChange={onChange}
-      accept={accept}
-    ></input>
-  </IconContainer>
-)
+export const Attachment = ({ onChange, accept, customAttacments }) => {
+  const { getThemeProperty } = useContext(WebchatContext)
+
+  const CustomAttachments = getThemeProperty(
+    WEBCHAT.CUSTOM_PROPERTIES.customAttachments,
+    undefined
+  )
+  return (
+    <IconContainer role={ROLES.ATTACHMENT_ICON}>
+      <label htmlFor='attachment' style={{ marginTop: 4 }}>
+        {CustomAttachments ? (
+          <CustomAttachments />
+        ) : (
+          <Icon src={AttachmentIcon} />
+        )}
+      </label>
+      <input
+        type='file'
+        name='file'
+        id='attachment'
+        style={{ display: 'none' }}
+        onChange={onChange}
+        accept={accept}
+      />
+    </IconContainer>
+  )
+}
--- a/packages/botonic-react/src/webchat/components/attachment.jsx
+++ b/packages/botonic-react/src/webchat/components/attachment.jsx
@@ -3,32 +3,51 @@ import React, { useContext } from 'react'
 import AttachmentIcon from '../../assets/attachment-icon.svg'
 import { ROLES, WEBCHAT } from '../../constants'
 import { WebchatContext } from '../../contexts'
-import { Icon, IconContainer } from './common'
+import { ConditionalAnimation } from '../components/conditional-animation'
+import { Icon } from './common'
 
-export const Attachment = ({ onChange, accept, customAttacments }) => {
+export const Attachment = ({ onChange, accept, enableAttachments }) => {
   const { getThemeProperty } = useContext(WebchatContext)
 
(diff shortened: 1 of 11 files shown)


packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -774,18 +774,40 @@ export const Webchat = forwardRef((props, ref) => {
...
+    return (
+      getThemeProperty(
+        WEBCHAT.CUSTOM_PROPERTIES.enableEmojiPicker,
+        props.enableEmojiPicker
+      ) ?? hasCustomEmojiPicker
+    )
+  }
+  const emojiPickerEnabled = isEmojiPickerEnabled()
+
+  const isAttacmentsEnabled = () => {
```
vanbasten17, line 792: Typo: Attacments -> Attachments. There are some of them.

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -774,18 +774,40 @@ export const Webchat = forwardRef((props, ref) => {
...
-    WEBCHAT.CUSTOM_PROPERTIES.enableEmojiPicker,
-    props.enableEmojiPicker
-  )
-  const attachmentsEnabled = getThemeProperty(
-    WEBCHAT.CUSTOM_PROPERTIES.enableAttachments,
-    props.enableAttachments
-  )
+
+  const isEmojiPickerEnabled = () => {
+    const hasCustomEmojiPicker = !!getThemeProperty(
```
vanbasten17, line 779: Is !! needed? Negating twice will leave the same value as if we don't have it.

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -774,18 +774,40 @@ export const Webchat = forwardRef((props, ref) => {
...
+      getThemeProperty(
+        WEBCHAT.CUSTOM_PROPERTIES.enableEmojiPicker,
+        props.enableEmojiPicker
+      ) ?? hasCustomEmojiPicker
+    )
+  }
+  const emojiPickerEnabled = isEmojiPickerEnabled()
+
+  const isAttacmentsEnabled = () => {
+    const hasCustomAttacments = !!getThemeProperty(
```
vanbasten17, line 793: Same here, also correct the typo

vanbasten17 (approved):
LGTM! 🤟

packages/botonic-react/src/webchat/components/attachment.jsx
```diff
@@ -1,21 +1,34 @@
...
-    ></input>
-  </IconContainer>
-)
+export const Attachment = ({ onChange, accept, customAttacments }) => {
+  const { getThemeProperty } = useContext(WebchatContext)
+
+  const CustomAttachments = getThemeProperty(
+    WEBCHAT.CUSTOM_PROPERTIES.customAttachments,
+    undefined
+  )
```
asastre, line 14: Since we're getting this theme property in the `webchat.jsx` file and we're already checking there if a custom attachment has been declared, why not leave the `Attachment` component as it is now and add the logic of deciding which one to render in the `webchat.jsx` file (as it is right now with the `SendButton` component)?

This way we keep the components much simpler. What do you think @Iru89 @vanbasten17?

asastre (commented):
Agree with @vanbasten17 comments and also I suggest to move the custom rendering logic to the `webchat.jsx` file (the same way it's done with the `SendButton` component), let me know if it make sense to you.

packages/botonic-react/src/webchat/webchat.jsx
Iru89: it would not be necessary but I do it because the function getThemeProperty does not return a boolean but I use the result as a boolean

packages/botonic-react/src/webchat/webchat.jsx
Iru89: it would not be necessary but I do it because the function getThemeProperty does not return a boolean but I use the result as a boolean.

packages/botonic-react/src/webchat/components/attachment.jsx
Iru89: I am of the opposite opinion. I would create files for each component and manage the logic in that file. This way when viewing the webchat file I would only see where the EmojiPicker, Attachments or Sendbutton is. When you go into the file for each component you would see if there is a custom and what that component does.

packages/botonic-react/src/webchat/webchat.jsx
Iru89: done

packages/botonic-react/src/webchat/components/attachment.jsx
Iru89: I have made a commit with WIP so that you can understand what I explained in the previous comment.

asastre: I like this approach also :+1:! The most important thing is to have consistency by applying the same logic in all 3 components.

Iru89:
in the [commit](https://github.com/hubtype/botonic/pull/2314/commits/74ffb1da0ce83ff68ccfcf65d79ebffe4fac38c0) I have removed the IconContainer that is only used in this button what really changes is the padding but I think it still looks good when tweaking the styled component UserInputContainer.

vanbasten17 (approved):
Go for it 🚀

PR #2330, opened 2023-04-13
Botonic-dx: additional settings for bots libraries and testing #SER-1765

alexgimenez02:
## Description
This code adds more paths to the `testPathIgnorePatterns` and to transformIgnorePatterns, this is added to avoid some errors when testing using Jest library.

Checking which files produced an extended coverage that slowed down the execution of the unit tests, we added those file paths (or file folders) using regular expressions to then, avoid the coverage to run those lines of code and speed up the testing process both locally and in CI.


vanbasten17 (approved):
Cool! What was the issue with `axios` exactly?

alexgimenez02:
> Cool! What was the issue with `axios` exactly?

When executing tests that contain some call to axios, the tests would fail, but adding `|axios|` inside transformIgnorePatterns seems to solve that error. 
The error was: `Cannot use import statement outside a module`

manuelfidalgo (approved):
Awesome work!

PR #2353, opened 2023-04-24
fix(react): lat, long to be numbers instead of strings

vanbasten17:
## Description
* Fixed `lat`, `long` properties of `Location` to be numbers.

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/docs/docs/components/components-location.md
+++ b/docs/docs/components/components-location.md
@@ -9,30 +9,24 @@ The `Location` component is used to display precise locations, for example if a 
 
 ![](https://botonic-doc-static.netlify.com/images/doc_location.png)
 
-
 ## Properties
 
 | Property | Type   | Description                     | Required | Default value |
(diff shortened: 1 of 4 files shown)


manuelfidalgo:
Thanks @vanbasten17 for taking care of this!
Should we also remove quotes from [example in doc](https://github.com/hubtype/botonic/blob/master-lts/docs/docs/components/components-location.md)?

vanbasten17:
> Thanks @vanbasten17 for taking care of this!
> Should we also remove quotes from [example in doc](https://github.com/hubtype/botonic/blob/master-lts/docs/docs/components/components-location.md)?

Yes @manuelfidalgo , it should be `lat={number} long={number}`, I'll add it later 👌

MarcosCA (approved):

manuelfidalgo (approved):
thanks!

PR #2364, opened 2023-04-26
fix(core): fix plugin types, allow to have plugins without pre/post, added new ResolvedPlugins type

vanbasten17:
## Description
* Fixed types for `Plugin` interface. Now `pre` and `post` are optional and they can return a promise or not.
* Execute methods in `runPlugins` only if they are defined.
* Added new type `ResolvedPlugins` to distinguish between the ones that are loaded in memory and the ones read from plugins file.

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-core/src/models/legacy-types.ts
+++ b/packages/botonic-core/src/models/legacy-types.ts
@@ -76,6 +76,13 @@ export interface PluginConfig<T> {
   resolve: { default: PluginConstructor<T> }
 }
 
+export interface ResolvedPlugin extends Plugin {
+  id: string
+  name: string
+  config: any
+}
+export type ResolvedPlugins = Record<string, ResolvedPlugin>
+
 export type InputType =
   | typeof INPUT.AUDIO
   | typeof INPUT.BUTTON_MESSAGE
@@ -225,10 +232,10 @@ export interface BotResponse extends BotRequest {
 }
 
 export interface PluginPreRequest extends BotRequest {
-  plugins: Plugin[]
+  plugins: ResolvedPlugins
 }
 export interface PluginPostRequest extends BotResponse {
-  plugins: Plugin[]
+  plugins: ResolvedPlugins
 }
 
 export interface Plugin {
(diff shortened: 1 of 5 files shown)


alexgimenez02 (approved):

packages/botonic-core/src/models/legacy-types.ts
```diff
@@ -232,8 +232,8 @@ export interface PluginPostRequest extends BotResponse {
 }
 
 export interface Plugin {
-  post(request: PluginPostRequest): void
-  pre(request: PluginPreRequest): void
+  post?(request: PluginPostRequest): void | Promise<void>
+  pre?(request: PluginPreRequest): void | Promise<void>
```
asastre, line 236: The way it's done right now, the plugins require to have both `pre` and `post` methods because they're always called within the `runPlugins` method, so the `?` should not be added here unless the behaviour is changed in the `runPlugins` method to call the `pre` and `post` methods only if they exist.

asastre (changes requested):

packages/botonic-core/src/models/legacy-types.ts
vanbasten17: fixed @asastre 👌

packages/botonic-core/src/plugins.ts
```diff
@@ -32,11 +39,12 @@ export async function runPlugins(
   dataProvider?: DataProvider
 ): Promise<void> {
   for (const key in plugins) {
-    const p = await plugins[key]
+    const p: Plugin = await plugins[key]
```
asastre, line 42: It would be nice to solve the `//TODO` we have in the line 32 of this same file and create a proper type for the `plugins` argument to avoid adding this `Plugin` type here.
Not to do it now, but in a near future :wink:

asastre (approved):

Iru89 (approved):

packages/botonic-core/src/plugins.ts
vanbasten17: added 👍

PR #2386, opened 2023-04-28
Botonic doc: Updating clients that use it

manuelalferez:
## Description

The customer list was out of date. 

I have updated the list with an image to make it more visual and cleaner.


manuelalferez:
@vanbasten17 if you want me to add something else to this PR to make it not so simple?

vanbasten17:
Nice @manuelalferez !! 👌✨

PR #2417, opened 2023-05-09
fix: when the same attachment is uploaded twice, the second one will not trigger the onChange event

TDzouchen:
## Description
When the same attachment is uploaded twice, the second one will not trigger the `onChange` event

### Before
https://user-images.githubusercontent.com/86048542/236981295-c82db95e-7c65-49ad-9dae-03997ee107a4.mov

### After
https://user-images.githubusercontent.com/86048542/236982343-a2a43d1d-b56a-43e9-82f9-8bb3f03dd0f8.mov

Solved it by emptying the input.value value

## Testing

The pull request...

- has unit tests
- has integration tests
- doesn't need tests because... **[very very minor changes]**

### DIFF
--- a/packages/botonic-react/src/webchat/components/attachment.jsx
+++ b/packages/botonic-react/src/webchat/components/attachment.jsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react'
+import React, { useContext, useRef } from 'react'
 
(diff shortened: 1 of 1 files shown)


vanbasten17:
Nice catch @TDzouchen ! Thanks for your contribution! 🤗

PR #2463, opened 2023-05-14
Chore/improve tree shaking and update to React 17

vanbasten17:
## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-react/src/webchat/components/emoji-picker.jsx
```diff
@@ -64,7 +64,13 @@ export const OpenedEmojiPicker = props => {
     <div ref={ref}>
       {isComponentVisible && (
         <Container role={ROLES.EMOJI_PICKER}>
-          <Picker onEmojiClick={props.onEmojiClick} disableAutoFocus={true} />
+          <Picker
+            width='100%'
+            height='19rem'
```
Iru89, line 69: if we use rem and the user has a very large font size in the default browser, this will be larger than expected.
I think it is better to fix this with px

Iru89 (approved):

packages/botonic-react/src/webchat/use-storage-state-hook.js
```diff
@@ -1,9 +1,20 @@
 import { useCallback, useEffect, useState } from 'react'
 
-//Code taken from https://github.com/leny/react-use-storage
-const evtTarget = new EventTarget()
+// Code taken from https://github.com/leny/react-use-storage/blob/master/src/index.js
+const IS_BROWSER =
```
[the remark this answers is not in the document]
vanbasten17: let's keep this atm as is how they used it in their library

packages/botonic-react/src/webchat/components/emoji-picker.jsx
vanbasten17: let's change it in the next PR 👍

PR #2468, opened 2023-05-16
Plugin flow builder: payloads

Iru89:
## Description
This PR adds the logic to detect that a button is of type payload read the payload content and put the text defined as payload in the payload that the bot receives. 
This way for example we can make that after a Handoff a payload rating is sent and this executes an action of the bot that depends on the channel it will send a custom component (webchat) or if it is receiving the rating it will save it. It will also be used to open webviews (in whatsapp we will use tamplates as it is necessary to create a button with the webview attribute)

## Context
When you create a button in flow builder that has a payload, this is a content with a payload object and what was received in the bot was the id of this content

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/index.ts
+++ b/packages/botonic-plugin-flow-builder/src/index.ts
@@ -26,9 +26,9 @@ import {
   StartNode,
 } from './flow-builder-models'
 import { DEFAULT_FUNCTIONS } from './functions'
+import { updateButtonUrls } from './helpers'
 import { BotonicPluginFlowBuilderOptions } from './types'
 import { resolveGetAccessToken } from './utils'
-import { updateButtonUrls } from './helpers'
 
 export default class BotonicPluginFlowBuilder implements Plugin {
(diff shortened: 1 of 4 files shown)


packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -230,12 +242,14 @@ export default class BotonicPluginFlowBuilder implements Plugin {
...
-      ?.values.map(value => ({ [value.name]: value.value }))
-    if (!nameValues) {
-      throw new Error(`No arguments found for node with id ${functionNodeId}`)
-    }
+    const nameValues =
+      functionNode.content.arguments
+        .find(arg => arg.locale === locale)
+        ?.values.map(value => ({ [value.name]: value.value })) || []
+
+    // if (!nameValues) {
```
vanbasten17, line 250: please remove comment

vanbasten17 (approved):
LGTM! 👌

PR #2476, opened 2023-05-17
New Textarea component

Iru89:
## Description
- keeps the autosize function up to 4 lines
- emoji picker works
- all tests are in green

## Context
When moving to node 16, npm 8 with react 17, we have problems with the react-textarea-autosize library.
The latest version fails the tests because it uses a forwardRef that doesn't work with jest, and the previous versions don't use react 17 so I have problems with peer dependencies.
With this we can reduce the final bundle by using less libraries.

I have created a textarea component with react styled-components and a hook to control the state. This way in the future from webchat we could make a custom component for the textarea and use the same hook to control the state.  
It would be necessary to move all the attributes that are used inside the hook use-text-area to a context, so it would not be necessary to pass anything when calling this hook. I will make a refactor to achieve this.
By extracting the functions into a hook we also reduce the size of the webchat file which is too big.

## Testing

The pull request...

- has unit tests
- has integration tests


vanbasten17:
@Iru89 is this still in draft?

Iru89:
> @Iru89 is this still in draft?

yes

PR #2500, opened 2023-05-22
Workflows: Create a reusable workflow for most flows #SER-2260

alexgimenez02:
## Description
Creating a new reusable workflow to simplify action updates or version updates.

## Context
With @vanbasten17 we noticed that the workflows for each botonic package had a similar structure, and to simplify both individual workflows and actions version updates, we decided to create a reusable workflow

Checking each workflow, we could extract some common steps to use on each workflow. Also added input flags that enable or disable different steps.

[TBW]

## Testing

The pull request has no tests


vanbasten17 (approved):
This is amazing 😍

PR #2501, opened 2023-05-22
Plugin flow builder: multichannel

Iru89:
## Description
- add a function to get a content through the code. It will be needed for campaigns
- adds a new action extending the FlowBuilderAction that uses the multichannel in the render

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/action.tsx
+++ b/packages/botonic-plugin-flow-builder/src/action.tsx
@@ -1,4 +1,4 @@
-import { ActionRequest, RequestContext } from '@botonic/react'
+import { ActionRequest, Multichannel, RequestContext } from '@botonic/react'
 import React from 'react'
 
 import { FlowContent } from './content-fields'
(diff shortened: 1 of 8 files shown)


packages/botonic-plugin-flow-builder/src/action.tsx
```diff
@@ -57,3 +57,16 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
...
   }
 }
+
+export class FlowBuilderMultichannelAction extends FlowBuilderAction {
+  render(): JSX.Element | JSX.Element[] {
+    // @ts-ignore
+    const { contents, handoffNode } = this.props
+    return (
+      //@ts-ignore
+      <Multichannel text={{ buttonsAsText: false }}>
```
vanbasten17, line 67: we should think how we configure the multichannel from the outside in the case of flow-builder-plugin. for now it's okay

packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -68,6 +68,16 @@ export default class BotonicPluginFlowBuilder implements Plugin {
     return content
   }
 
+  async getContentByCode(code: string): Promise<NodeComponent> {
+    const flow = await this.flow
+    const content = flow.nodes.find(node => {
+      if ('code' in node) return node.code === code
+      return undefined
+    })
```
vanbasten17, line 76: early return normally suggest to do this in the other order: https://medium.com/swlh/return-early-pattern-3d18a41bba8
```
if(exception) return 
// happy path
```
```suggestion
const content = flow.nodes.find(node => {
      if (!('code' in node)) return undefined
      return node.code === code
    })
```

vanbasten17 (approved):

packages/botonic-plugin-flow-builder/src/index.ts
Iru89: I have applied the change in the following PR (tracking)

`const content = flow.nodes.find(node =>
      'code' in node ? node.code === code : false
    )`

PR #2505, opened 2023-05-23
Plugin flow builder: tracking

Iru89:
## Description
Add the possibility to pass a function from the bot configuration to do the tracking.
- track the name of each first content, not the "followups"
- track the state of the queue as it passes through a conditional queue. At this point it would be interesting for the queue status component to store the name of the project. 
- track HANDOFF_SUCESSFULL when doing a handoff

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/action.tsx
+++ b/packages/botonic-plugin-flow-builder/src/action.tsx
@@ -46,6 +46,11 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
       locale
     )
 
+    if (flowBuilderPlugin.trackEvent) {
+      // TODO: track all targets nodes?
+      await flowBuilderPlugin.trackEvent(request, contents[0].code)
+    }
+
     if (handoffNode) await doHandoff(request, locale, handoffNode)
 
     return { contents, handoffNode }
--- a/packages/botonic-plugin-flow-builder/src/action.tsx
+++ b/packages/botonic-plugin-flow-builder/src/action.tsx
@@ -6,7 +6,7 @@ import { doHandoff } from './handoff'
 import { getFlowBuilderPlugin } from './helpers'
 
 type FlowBuilderActionProps = {
-  content?: FlowContent[]
+  contents: FlowContent[]
   isHandoff?: boolean
 }
 
@@ -52,9 +52,8 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
   }
 
   render(): JSX.Element | JSX.Element[] {
-    // @ts-ignore
-    const { contents, handoffNode } = this.props
-    return contents!.map(content => content.toBotonic(content.id))
+    const { contents } = this.props
+    return contents.map(content => content.toBotonic(content.id))
   }
 }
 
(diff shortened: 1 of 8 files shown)


packages/botonic-plugin-flow-builder/src/action.tsx
```diff
@@ -46,26 +46,28 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
       locale
     )
 
+    if (flowBuilderPlugin.trackEvent) {
+      // TODO: track all targets nodes?
```
vanbasten17, line 50: @asastre should we track just the "parent" or all the follow-up/target contents?

packages/botonic-plugin-flow-builder/src/functions/conditional-queue-status.ts
```diff
@@ -26,5 +34,18 @@ export async function conditionalQueueStatus({
     }
   )
   const isAvailable = response.data.available
+
+  console.log('conditionalQueueStatus', { data: response.data })
```
vanbasten17, line 38: Remove console log

packages/botonic-plugin-flow-builder/src/functions/conditional-queue-status.ts
```diff
@@ -26,5 +34,18 @@ export async function conditionalQueueStatus({
...
+  const flowBuilderPlugin = getFlowBuilderPlugin(request.plugins)
+  if (flowBuilderPlugin.trackEvent) {
+    const eventName = 'QUEUE_STATUS'
+    const args = {
+      queue_id,
+      queue_name,
+      status: isAvailable ? 'open' : 'closed',
+    }
+    await flowBuilderPlugin.trackEvent(request, eventName, args)
```
vanbasten17, line 47: For next iterations, I would use the decorator pattern to automatically call this, but atm is quite overengineering. Let's if we need to extend this logic in the future

packages/botonic-plugin-flow-builder/src/handoff.ts
```diff
@@ -36,15 +42,19 @@ export async function doHandoff(
   // }
 
   await handOffBuilder.handOff()
+
+  if (flowBuilderPlugin.trackEvent) {
+    await flowBuilderPlugin.trackEvent(request, 'HANDOFF_SUCCESSFULL', {
```
vanbasten17, line 47: What about passing also the list of enums to track event?

vanbasten17 (approved):
@Iru89 you rock!! Thanks @Iru89 🤗

packages/botonic-plugin-flow-builder/src/handoff.ts
```diff
@@ -36,15 +42,19 @@ export async function doHandoff(
...
+      queueName: queueFound?.name,
```
AinaVendrell, line 48: i think this event should be more similar to the `QUEUE_STATUS` event.
One only tracks the `queueName` and the other the `queue_id` and `queue_name`. Also one is in camel case and the other in snake case

packages/botonic-plugin-flow-builder/src/handoff.ts
Iru89: I think it is better to have default events that are always the same. Maybe we could pass a list optionally and if it is not passed we could use the default events.

vanbasten17: cool! 👌

AinaVendrell (approved):

packages/botonic-plugin-flow-builder/src/action.tsx
vanbasten17: cool! let's iterate it for next sprints

PR #2506, opened 2023-05-24
feat(core): Don't execute bot when receiving a 'chatevent' 

asastre:
## Description
Avoid executing the bot when receiving a 'chatevent' because so far it's used only to send typing on/off notifications to the backend (using the bot as a messenger)

## Context
When the user types something in the webchat, a 'chatevent' message is sent to the bot as an input message and this was generating an error when using the bot locally because there is no parser for this type of messages. Also, the bot was being executed and there is no need to be executed

The bot will be executed anyway when a 'chatevent' is received so to avoid executing all the `input` logic, I've added an early return if the input received is of 'chatevent' type.

## Testing

The pull request has no new tests but some old tests have been fixed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,721 +2,214 @@
   "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
-    "@babel/code-frame": {
-      "version": "7.8.3",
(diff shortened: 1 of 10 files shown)


vanbasten17 (approved):
LGTM!! 🚀  We would need to review this pre-commit check which is failing 🤔

PR #2510, opened 2023-05-30
feat(core/react): add WhatsappButtonList component in Botonic

vanbasten17:
## Description
* (WIP) Added support for [WhatsappButtonList](https://developers.facebook.com/docs/whatsapp/guides/interactive-messages/#list-messages)

## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-react/src/components/whatsapp-button-list.tsx
```diff
@@ -0,0 +1,67 @@
...
+        type={INPUT.WHATSAPP_BUTTON_LIST}
+      >
+        {message}
+      </Message>
+    )
+  }
+
+  const renderNode = () => {
+    return (
+      // @ts-ignore
```
AlbertGom, line 54: Why is needed this ts-ignore?

packages/botonic-react/src/components/whatsapp-button-list.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <message
+        {...props}
+        header={props.header}
+        body={props.body}
+        footer={props.footer}
+        action={props.action}
```
AlbertGom, line 60: It is not necessary to pass these props since you are passing the` {...props}` this way, no?

packages/botonic-react/src/components/whatsapp-button-list.tsx
```diff
@@ -0,0 +1,67 @@
...
+const serialize = _whatsappTemplateProps => {
+  // TODO: Implement to have data persistance in localStorage, not needed for this WhatsApp development
+  return {}
+}
+
+export interface WhatsappButtonListProps {
+  header: string
+  body: string
+  footer: string
+  action: Action
```
AlbertGom, line 34: Why is the header and the footer props needed?

packages/botonic-react/src/components/whatsapp-button-list.tsx
```diff
@@ -0,0 +1,67 @@
...
+    )
+  }
+
+  const renderNode = () => {
+    return (
+      // @ts-ignore
+      <message
+        {...props}
+        header={props.header}
+        body={props.body}
```
AlbertGom, line 58: I think the body should be passed this way `body: { text: 'Message' }`, no?

AlbertGom (commented):
I have some questions and doubts! Also, in a future iteration it would be great if the component adds ellipsis to the title, description and to the text of the button that opens the WA list when they exceed the characters limit.

AlbertGom:
Also, botonic core and botonic react tests are faliling, take a look!

packages/botonic-react/src/components/whatsapp-button-list.tsx
Iru89: they should be optional

packages/botonic-react/src/components/whatsapp-button-list.tsx
vanbasten17: i don' know atm, is not top prior as we need to have the feature ASAP, I'll add it to a new task and let's add all the feedback there

vanbasten17 (commented):
WIP

asastre (changes requested):
Small change and some comments/doubts before approving

packages/botonic-react/src/components/whatsapp-button-list.tsx
```diff
@@ -0,0 +1,64 @@
...
+      >
+        {message}
+      </Message>
+    )
+  }
+
+  const renderNode = () => {
+    return (
+      // @ts-ignore Property 'message' does not exist on type 'JSX.IntrinsicElements'.
+      <message {...props} type={INPUT.WHATSAPP_BUTTON_LIST} />
```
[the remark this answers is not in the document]
vanbasten17: As far as I remember, we talked about this. We should rethink the architecture of this in order to keep it as simple as possible.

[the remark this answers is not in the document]
vanbasten17: Btw, I agree with you that `<message>` is not one of the HTML tags

packages/botonic-react/src/components/whatsapp-button-list.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+interface TextMessage {
+  text: string
+}
+
+export interface WhatsappButtonListProps {
+  header?: string
+  body: TextMessage
+  footer?: TextMessage
```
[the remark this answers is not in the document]
vanbasten17: taking a look!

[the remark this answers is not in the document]
AinaVendrell: I change it to string and convert it to the proper type when rendering the <message>

asastre (approved):

PR #2515, opened 2023-05-31
Check failing workflows #SER-2321

alexgimenez02:
## Description
Pump up Pre-commit version and rollbacked to the previous pre-commit script created by Dani 3 years ago

## Context
Detect secrets failed when running the pre-commit workflow

Checking newer versions of detect-secrets to see if it does fails the pre-commit workflow.
Pumped detect-secrets version to v1.4.0.
Changed pre-commit scripts to run old lint script.


vanbasten17 (approved):
Go for it now if it doesn't break! Let's iterate it in next PRs. 💪

PR #2517, opened 2023-06-06
feature: Removed client_secret #PLA-20

Khaeshah:
## Description
We have configured our application client as public. 
We cannot use PKCE as initially intended because we are both the authorization server and the resource server. Since the vulnerability identified by the penetration test lies in the client_secret present in our code, and we are not using it by definition, we have decided to remove it as it has no effect.

## Context
Screenshot from 2023-06-06 11-05-33

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/.github/workflows/botonic-cli-tests.yml
+++ b/.github/workflows/botonic-cli-tests.yml
@@ -14,6 +14,7 @@ jobs:
     with:
       PACKAGE_NAME: Botonic cli tests
       PACKAGE: botonic-cli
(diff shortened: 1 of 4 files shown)


MarcosCA (approved):

vanbasten17:
IMHO, I think we should merge it and then let’s fix the tests in next iterations @Khaeshah

gbarba (approved):

PR #2531, opened 2023-06-15
Flow builder plugin: Fix url buttons + refactor

asastre:
## Description
Multiple things have been done in this PR:
- Add pre-commit to the Flow Builder plugin.
- Fix the URL buttons issue to make them point correctly to the URL instead of to the content ID.
- Create a new `FlowBuilderApi` class to handle all the accesses to the flow instead of doing it in the plugin itself.

## Context
The URL buttons were not correctly generated in Botonic and they were pointing to the content ID instead of pointing to the URL.
Trying to solve this issue I came accross with a limitation of how the contents were being retreived from the flow and how the plugin was generating the final content to be sent to Botonic, that's why I ended up doing a refactor of this part.

The process of generating the flow contents (`FlowButton`, `FlowText`, etc.) was not being done within those clases but also in the `getContents` function of the plugin and this was making the code more complex to mantain and to understand.

Now, each flow class is responsible of fetching correctly all the other contained components (buttons within a text, for example). To be able to do this, we pass the `FlowBuilderApi` instance to any flow class that needs to fetch nodes from the flow.

## Future improvements
- Add cache to the `FlowBuilderApi`'s `init` method to not fetch the whole flow in each bot interaction.
- Try to apply the same logic with the function and follow-up nodes and move all its logic from `getContent` method to independent classes (the same way it's done with `FlowText`, `FlowCarousel`...)
- Add tests!

## Testing

The pull request has no test.

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/action.tsx
+++ b/packages/botonic-plugin-flow-builder/src/action.tsx
@@ -34,11 +34,7 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
         request.input.data,
         locale
       )
-      if (intentNode) {
-        targetNode = intentNode
-      } else if (keywordNode) {
-        targetNode = keywordNode
-      }
+      targetNode = intentNode ?? keywordNode ?? targetNode
     }
 
     if (!targetNode) {
@@ -48,7 +44,7 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
       alternateFallbackMessage = !alternateFallbackMessage
     }
 
-    const { contents, handoffNode } = await flowBuilderPlugin.getContent(
+    const { contents, handoffNode } = await flowBuilderPlugin.getContents(
       targetNode,
       locale
     )
--- a/packages/botonic-plugin-flow-builder/src/action.tsx
+++ b/packages/botonic-plugin-flow-builder/src/action.tsx
@@ -2,6 +2,7 @@ import { ActionRequest, Multichannel, RequestContext } from '@botonic/react'
 import React from 'react'
 
 import { FlowContent } from './content-fields'
+import { HtNodeWithContent } from './content-fields/hubtype-fields'
 import { doHandoff } from './handoff'
 import { getFlowBuilderPlugin } from './helpers'
 
@@ -18,31 +19,37 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
     const flowBuilderPlugin = getFlowBuilderPlugin(request.plugins)
     const locale = flowBuilderPlugin.getLocale(request.session)
     const payload = request.input.payload
-    let targetContentId: string | undefined = payload
+    let targetNode: HtNodeWithContent | string | undefined = payload
+
     if (!payload && request.session.is_first_interaction) {
-      targetContentId = await flowBuilderPlugin.getStartId()
+      targetNode = flowBuilderPlugin.cmsApi.getStartNode()
     }
-    if (!payload) {
-      const intentPayload = await flowBuilderPlugin.getPayloadByIntent(
+
+    if (!payload && request.input.data) {
+      const intentNode = flowBuilderPlugin.cmsApi.getNodeByIntent(
         request.input,
         locale
       )
-      if (intentPayload) targetContentId = intentPayload
-      const keywordPayload = await flowBuilderPlugin.getPayloadByKeyword(
-        request.input,
+      const keywordNode = flowBuilderPlugin.cmsApi.getNodeByKeyword(
+        request.input.data,
         locale
       )
-      if (keywordPayload) targetContentId = keywordPayload
+      if (intentNode) {
+        targetNode = intentNode
+      } else if (keywordNode) {
+        targetNode = keywordNode
+      }
     }
-    if (!targetContentId) {
-      targetContentId = await flowBuilderPlugin.getFallbackId(
+
+    if (!targetNode) {
+      targetNode = flowBuilderPlugin.cmsApi.getFallbackNode(
         alternateFallbackMessage
       )
       alternateFallbackMessage = !alternateFallbackMessage
     }
 
-    const { contents, handoffNode } = await flowBuilderPlugin.getContents(
-      targetContentId,
+    const { contents, handoffNode } = await flowBuilderPlugin.getContent(
+      targetNode,
       locale
     )
 
--- a/packages/botonic-plugin-flow-builder/src/api.ts
+++ b/packages/botonic-plugin-flow-builder/src/api.ts
@@ -18,6 +18,8 @@ export class FlowBuilderApi {
   url: string
   flow: HtFlowBuilderData
 
+  private constructor() {}
+
   static async create(options: FlowBuilderApiOptions): Promise<FlowBuilderApi> {
     const newApi = new FlowBuilderApi()
 
--- a/packages/botonic-plugin-flow-builder/src/api.ts
+++ b/packages/botonic-plugin-flow-builder/src/api.ts
@@ -15,24 +15,26 @@ import {
 import { FlowBuilderApiOptions } from './types'
 
 export class FlowBuilderApi {
-  private readonly url: string
-  public flow: HtFlowBuilderData
+  url: string
+  flow: HtFlowBuilderData
 
-  constructor(options: FlowBuilderApiOptions) {
-    this.url = options.url
-    if (options.flow) {
-      this.flow = options.flow
-    }
+  static async create(options: FlowBuilderApiOptions): Promise<FlowBuilderApi> {
+    const newApi = new FlowBuilderApi()
+
+    newApi.url = options.url
+    newApi.flow = options.flow ?? (await newApi.getFlow(options.accessToken))
+
+    return newApi
   }
 
-  async init(token: string) {
+  private async getFlow(token: string): Promise<HtFlowBuilderData> {
     const { data } = await axios.get(this.url, {
       headers: { Authorization: `Bearer ${token}` },
     })
-    this.flow = data
+    return data as HtFlowBuilderData
   }
 
-  getNode<T extends HtNodeComponent>(id: string): T {
+  getNodeById<T extends HtNodeComponent>(id: string): T {
     const node = this.flow.nodes.find(node => node.id === id)
     if (!node) throw Error(`Node with id: '${id}' not found`)
     return node as T
@@ -51,7 +53,7 @@ export class FlowBuilderApi {
       node => node.type === HtNodeStartType.STARTUP
     ) as HtStartNode | undefined
     if (!startUpNode) throw new Error('Start-up id must be defined')
-    return this.getNode(startUpNode.target.id)
+    return this.getNodeById(startUpNode.target.id)
   }
 
   getFallbackNode(alternate: boolean): HtNodeWithContent {
@@ -67,11 +69,11 @@ export class FlowBuilderApi {
     }
     const fallbackSecondMessage = fallbackNode.content.second_message
     if (!fallbackSecondMessage) {
-      return this.getNode(fallbackFirstMessage.id)
+      return this.getNodeById(fallbackFirstMessage.id)
     }
     return alternate
-      ? this.getNode(fallbackFirstMessage.id)
-      : this.getNode(fallbackSecondMessage.id)
+      ? this.getNodeById(fallbackFirstMessage.id)
+      : this.getNodeById(fallbackSecondMessage.id)
   }
 
   getNodeByIntent(input: Input, locale: string): HtNodeWithContent | undefined {
@@ -91,7 +93,7 @@ export class FlowBuilderApi {
         )
         return (
           matchedIntentNode?.target &&
-          this.getNode<HtNodeWithContent>(matchedIntentNode?.target.id)
+          this.getNodeById<HtNodeWithContent>(matchedIntentNode?.target.id)
         )
       }
     } catch (error) {
@@ -131,7 +133,9 @@ export class FlowBuilderApi {
         this.matchKeywords(node, userInput, locale)
       )
       if (matchedKeywordNodes.length > 0 && matchedKeywordNodes[0].target) {
-        return this.getNode<HtNodeWithContent>(matchedKeywordNodes[0].target.id)
+        return this.getNodeById<HtNodeWithContent>(
+          matchedKeywordNodes[0].target.id
+        )
       }
     } catch (error) {
       console.error(`Error getting node by keyword '${userInput}': `, error)
@@ -152,11 +156,6 @@ export class FlowBuilderApi {
   }
 
   private containsAnyKeywords(input: string, keywords: string[]): boolean {
-    for (let i = 0; i < keywords.length; i++) {
-      if (input.includes(keywords[i])) {
-        return true
-      }
-    }
-    return false
+    return keywords.some(keyword => input.includes(keyword))
   }
 }
--- a/packages/botonic-plugin-flow-builder/src/api.ts
+++ b/packages/botonic-plugin-flow-builder/src/api.ts
@@ -1,0 +1,162 @@
+import { Input } from '@botonic/core'
+import axios from 'axios/index'
+
+import {
+  HtFallbackNode,
+  HtFlowBuilderData,
+  HtIntentNode,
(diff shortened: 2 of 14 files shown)


packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -0,0 +1,162 @@
...
+    return Boolean(result)
+  }
+
+  private containsAnyKeywords(input: string, keywords: string[]): boolean {
+    for (let i = 0; i < keywords.length; i++) {
+      if (input.includes(keywords[i])) {
+        return true
+      }
+    }
+    return false
```
AlbertGom, line 160: This can be simplified to` return !!keywords.find((keyword) => input.includes(keyword))`

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -0,0 +1,162 @@
...
+  public flow: HtFlowBuilderData
+
+  constructor(options: FlowBuilderApiOptions) {
+    this.url = options.url
+    if (options.flow) {
+      this.flow = options.flow
+    }
+  }
+
+  async init(token: string) {
```
AlbertGom, line 28: I think this name can be more descriptive like `initializeFlow()` or something like that?

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -0,0 +1,162 @@
...
+    const { data } = await axios.get(this.url, {
+      headers: { Authorization: `Bearer ${token}` },
+    })
+    this.flow = data
+  }
+
+  getNode<T extends HtNodeComponent>(id: string): T {
```
AlbertGom, line 35: Rename it by `getNodeById()` ?

AlbertGom (approved):
Nice refactor!!! And I agree that the plugin should have tests as soon as possible since the bigger it gets, the more difficult it will be to do tests.

packages/botonic-plugin-flow-builder/src/action.tsx
```diff
@@ -18,31 +19,37 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
...
+      }
     }
-    if (!targetContentId) {
-      targetContentId = await flowBuilderPlugin.getFallbackId(
+
+    if (!targetNode) {
+      targetNode = flowBuilderPlugin.cmsApi.getFallbackNode(
         alternateFallbackMessage
       )
       alternateFallbackMessage = !alternateFallbackMessage
```
AlbertGom, line 48: Remove this since this variable is not used

packages/botonic-plugin-flow-builder/src/action.tsx
```diff
@@ -18,31 +19,37 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
...
+      } else if (keywordNode) {
+        targetNode = keywordNode
+      }
     }
-    if (!targetContentId) {
-      targetContentId = await flowBuilderPlugin.getFallbackId(
+
+    if (!targetNode) {
+      targetNode = flowBuilderPlugin.cmsApi.getFallbackNode(
         alternateFallbackMessage
```
AlbertGom, line 46: This value is always `false`. Is this parameter needed then?

AlbertGom (changes requested):

packages/botonic-plugin-flow-builder/src/action.tsx
asastre: It will alternate between `true` and `false` but we have to fix the way this value is initialized because now it's shared with all users.

packages/botonic-plugin-flow-builder/src/action.tsx
```diff
@@ -18,31 +19,37 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
...
       )
       alternateFallbackMessage = !alternateFallbackMessage
     }
 
-    const { contents, handoffNode } = await flowBuilderPlugin.getContents(
-      targetContentId,
+    const { contents, handoffNode } = await flowBuilderPlugin.getContent(
```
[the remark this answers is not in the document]
asastre: I'll leave it as it was before (`getContents`) for now.

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -0,0 +1,162 @@
...
+    locale: string
+  ): HtNodeWithContent | undefined {
+    try {
+      const keywordNodes = this.flow.nodes.filter(
+        node => node.type == HtNodeWithContentType.KEYWORD
+      ) as HtKeywordNode[]
+      const matchedKeywordNodes = keywordNodes.filter(node =>
+        this.matchKeywords(node, userInput, locale)
+      )
+      if (matchedKeywordNodes.length > 0 && matchedKeywordNodes[0].target) {
```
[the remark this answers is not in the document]
asastre: I prefer to refactor this part in another PR

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -0,0 +1,162 @@
...
+      ) as HtIntentNode[]
+      const inputIntent = input.intent
+      const inputConfidence = input.confidence
+      if (inputIntent) {
+        const matchedIntentNode = intents.find(
+          node =>
+            inputIntent &&
+            this.hasIntent(node, inputIntent, locale) &&
+            inputConfidence &&
+            this.hasMetConfidenceThreshold(node, inputConfidence)
```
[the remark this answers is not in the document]
asastre: I prefer to refactor this part in another PR

packages/botonic-plugin-flow-builder/src/api.ts
asastre: Done!

packages/botonic-plugin-flow-builder/src/api.ts
asastre: I've refactored the way the `FlowBuilderApi` is initialized, now you have to call a static method `create` that will create a new `FlowBuilderApi` with the flow already fetched.

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -15,24 +15,26 @@ import {
...
-  public flow: HtFlowBuilderData
+  url: string
+  flow: HtFlowBuilderData
 
-  constructor(options: FlowBuilderApiOptions) {
-    this.url = options.url
-    if (options.flow) {
-      this.flow = options.flow
-    }
+  static async create(options: FlowBuilderApiOptions): Promise<FlowBuilderApi> {
```
AlbertGom, line 21: Just to know, why this function should be static?

packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -40,10 +43,8 @@ export default class BotonicPluginFlowBuilder implements Plugin {
   ) => Promise<void>
 
   constructor(readonly options: BotonicPluginFlowBuilderOptions) {
-    this.cmsApi = new FlowBuilderApi({
-      url: options.flowUrl,
-      flow: options.flow,
-    })
+    this.flowUrl = options.flowUrl
+    this.flow = options.flow
```
AlbertGom, line 47: Juat to know, in which case will you have the flow before? In fact, this class is used to fetch the flow among other things...

AlbertGom (approved):

packages/botonic-plugin-flow-builder/src/api.ts
asastre: Static functions can be used to create new instances of objects. 

In this case I thought that it made no sense to be able to create a `FlowBuilderApi` instance and be able to call its methods without having the flow fetched. 
That's why I'm using this static method to ensure that every new `FlowBuilderApi` instance will have always the flow already fetched to be used.

I've added the `constructor` as a `private` class, this way we ensure this class won't be never created with the new keyword (meaning with an empty flow).

packages/botonic-plugin-flow-builder/src/index.ts
asastre: Yes, you're right and I had the same doubt...! I find it also weird, but maybe this was something used in the initial versions and now is no longer needed. I'm keeping it just to ensure the logic remain the same as it was before this refactor.
Maybe @manuelalferez, @vanbasten17 or @AinaVendrell know why this was done this way.

packages/botonic-plugin-flow-builder/src/api.ts
asastre: For the reasons explained above.

If this method was not static, in order to call it, a `FlowBuilderApi` instance would be needed and that's what I'm trying to prevent by setting this method as static because otherwise you could be able to execute a `getNodeById` call after creating a new `FlowBuilderApi` instance (with the `new` method) without having the flow fetched.

Having this method as static I'm assuring that every `FlowBuilderApi` instance created (using the `create` method) will have already the flow fetched.

AlbertGom: Okay, I see thx!!

packages/botonic-plugin-flow-builder/src/index.ts
Iru89: Is for the case that you have a local flow without flowbuilder backend only for development.

packages/botonic-plugin-flow-builder/src/action.tsx
```diff
@@ -18,31 +19,33 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
     const flowBuilderPlugin = getFlowBuilderPlugin(request.plugins)
     const locale = flowBuilderPlugin.getLocale(request.session)
     const payload = request.input.payload
-    let targetContentId: string | undefined = payload
+    let targetNode: HtNodeWithContent | string | undefined = payload
```
Iru89, line 22: Better not to store the payload in this variable and initially make it undefined.
`let targetNode: HtNodeWithContent | string | undefined = undefined`

packages/botonic-plugin-flow-builder/src/action.tsx
```diff
@@ -18,31 +19,33 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
...
+
     if (!payload && request.session.is_first_interaction) {
-      targetContentId = await flowBuilderPlugin.getStartId()
+      targetNode = flowBuilderPlugin.cmsApi.getStartNode()
```
Iru89, line 25: better to create a constant with the name startTargetNode
`startTargetNode = flowBuilderPlugin.cmsApi.getStartNode()`

packages/botonic-plugin-flow-builder/src/action.tsx
```diff
@@ -18,31 +19,33 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
...
       )
-      if (intentPayload) targetContentId = intentPayload
-      const keywordPayload = await flowBuilderPlugin.getPayloadByKeyword(
-        request.input,
+      const keywordNode = flowBuilderPlugin.cmsApi.getNodeByKeyword(
+        request.input.data,
         locale
       )
-      if (keywordPayload) targetContentId = keywordPayload
+      targetNode = intentNode ?? keywordNode ?? targetNode
```
Iru89, line 37: in conjunction with the above comments
`targetNode = intentNode ?? keywordNode ?? startTargetNode ?? payload`

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -0,0 +1,163 @@
...
+  HtKeywordNode,
+  HtNodeComponent,
+  HtNodeStartType,
+  HtNodeWithContent,
+  HtNodeWithContentType,
+  HtStartNode,
+} from './content-fields/hubtype-fields'
+import { FlowBuilderApiOptions } from './types'
+
+export class FlowBuilderApi {
```
Iru89, line 17: I like how this looks in a class after this refactor

packages/botonic-plugin-flow-builder/src/content-fields/flow-button.tsx
```diff
@@ -1,24 +1,53 @@
...
+  ): FlowButton {
+    const payloadId = this.getPayloadId(cmsButton, locale)
+    const urlId = this.getUrlId(cmsButton, locale)
+
+    const newButton = new FlowButton(cmsButton.id)
+    newButton.text = this.getTextByLocale(locale, cmsButton.text)
+    if (cmsButton.target) {
+      newButton.payload = cmsButton.target.id
+    }
+    if (cmsButton.payload && payloadId) {
```
Iru89, line 31: I think it is a very good idea to do this logic within the content class instead of with functions that are not grouped in any class

Iru89 (approved):
Very nice refactor I really like the approach with the FlowBuilderApi class and the logic inside the FlowButton FlowCarousel classes ...

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -0,0 +1,163 @@
...
+  url: string
+  flow: HtFlowBuilderData
+
+  private constructor() {}
+
+  static async create(options: FlowBuilderApiOptions): Promise<FlowBuilderApi> {
+    const newApi = new FlowBuilderApi()
+
+    newApi.url = options.url
```
vanbasten17, line 26: `newApi` sounds strange to me, what about renaming it to just api? as it's a local variable no need to be "new", IMHO.

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -0,0 +1,163 @@
...
+      if (matchedKeywordNodes.length > 0 && matchedKeywordNodes[0].target) {
+        return this.getNodeById<HtNodeWithContent>(
+          matchedKeywordNodes[0].target.id
+        )
+      }
+    } catch (error) {
+      console.error(`Error getting node by keyword '${userInput}': `, error)
+    }
+
+    return undefined
```
vanbasten17, line 146: looking good! i'll try to do early-return instead: https://szymonkrajewski.pl/why-should-you-return-early/

packages/botonic-plugin-flow-builder/src/content-fields/flow-button.tsx
vanbasten17: same here with `new` prefix

packages/botonic-plugin-flow-builder/src/index.ts
vanbasten17: this is a WIP project, we have it noted in order to keep improving it, both code and docs (and test, of course).

vanbasten17 (approved):
Kudos for this amazing refactor! 💯

packages/botonic-plugin-flow-builder/src/index.ts
asastre: We can review it in future PRs

packages/botonic-plugin-flow-builder/src/action.tsx
asastre: As we discussed, if we enter to this action through a payload, the payload will be the targetNode's ID, so I would leave it as it is, this way we will have it already set it from the beginning of the action.

packages/botonic-plugin-flow-builder/src/action.tsx
asastre: Before doing any changes here, I think we should agree about the priority between the starting content, the keywords and the intents. 
The way the code was done (and it is also now), we're prioritizing the intents over the keywords over the starting content, is this what we really want? Shouldn't we prioritize the starting content instead?

packages/botonic-plugin-flow-builder/src/action.tsx
asastre: Maybe it's a good idea, yes, but let's first try to agree about [these priorities](https://github.com/hubtype/botonic/pull/2531#discussion_r1234527629).

packages/botonic-plugin-flow-builder/src/api.ts
asastre: This code has just been moved from one file to another, I haven't done any change to it and any change to be applied here I prefer to do it into a separated PR.
I agree with you that this code can be improved, we can do it in the next iteration :wink:!

vanbasten17:
@asastre , PR can be merged, go ahead if you consider it necessary

PR #2537, opened 2023-06-26
Flow builder: tracking with Hubtype analytics events

Iru89:
## Description
Add tracking to each FlowBuilder action option. Events are created which are those defined in hubtype-analytics. Although I have put as a dependency @botonic/plugin-hubtype-analytics I am only using the types to create the events. The call to the trackEvent function will enter as a FlowBuilder dependency.

## Context
The most important thing in this PR is to generate the concrete event at each point. When I tried to do this I saw that the files were getting too big and I applied a refactor to separate everything better.
So I have transformed the action file into a folder and created a file for the way to get the node: by user input, by intent, by keyword and another one with the logic to call the trackEvent function that will not always be defined in the flow builder plugin

Apply a refactor in the action splitting it into several files: 
- one with the logic to do the tracking function,
- one for the intents logic, creating the events and doing the tracking
- one for the keywords logic, creating the events and doing the tracking
It also refactors the doHandoff function, making a request to obtain the availability of the queue and do the corresponding tracking.

## Testing
The flow builder plugin logic is yet to be tested.

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/action/index.tsx
+++ b/packages/botonic-plugin-flow-builder/src/action/index.tsx
@@ -20,7 +20,9 @@ type FlowBuilderActionProps = {
 export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
   static contextType = RequestContext
 
-  static async botonicInit(request: ActionRequest): Promise<any> {
+  static async botonicInit(
+    request: ActionRequest
+  ): Promise<FlowBuilderActionProps> {
     const flowBuilderPlugin = getFlowBuilderPlugin(request.plugins)
     const locale = flowBuilderPlugin.getLocale(request.session)
 
--- a/packages/botonic-plugin-flow-builder/src/action/index.tsx
+++ b/packages/botonic-plugin-flow-builder/src/action/index.tsx
@@ -1,10 +1,17 @@
+import {
+  EventBotFaq,
+  EventFallback,
+  EventName,
+} from '@botonic/plugin-hubtype-analytics/lib/cjs/types'
 import { ActionRequest, Multichannel, RequestContext } from '@botonic/react'
 import React from 'react'
 
-import { FlowBuilderApi } from './api'
-import { FlowContent, FlowHandoff } from './content-fields'
-import { HtNodeWithContent } from './content-fields/hubtype-fields'
-import { getFlowBuilderPlugin } from './helpers'
+import { FlowBuilderApi } from '../api'
+import { FlowContent, FlowHandoff } from '../content-fields'
+import { HtNodeWithContent } from '../content-fields/hubtype-fields'
+import { getFlowBuilderPlugin } from '../helpers'
+import { trackEvent } from './tracking'
+import { getNodeByUserInput } from './user-input'
 
 type FlowBuilderActionProps = {
   contents: FlowContent[]
@@ -17,17 +24,16 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
     const flowBuilderPlugin = getFlowBuilderPlugin(request.plugins)
     const locale = flowBuilderPlugin.getLocale(request.session)
 
-    const targetNode = getTargetNode(flowBuilderPlugin.cmsApi, locale, request)
+    const targetNode = await getTargetNode(
+      flowBuilderPlugin.cmsApi,
+      locale,
+      request
+    )
 
     const contents = await flowBuilderPlugin.getContentsByNode(
       targetNode,
       locale
     )
-
-    if (flowBuilderPlugin.trackEvent) {
-      // TODO: track all targets nodes?
-      await flowBuilderPlugin.trackEvent(request, contents[0].code)
-    }
 
     const renderContents = contents.filter(async content => {
       if (content instanceof FlowHandoff) {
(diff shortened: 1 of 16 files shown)


packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const { contents } = this.props
+    return contents.map(content => content.toBotonic(content.id))
+  }
+}
+
+export class FlowBuilderMultichannelAction extends FlowBuilderAction {
+  render(): JSX.Element | JSX.Element[] {
+    const { contents } = this.props
+    return (
+      <Multichannel text={{ buttonsAsText: false }}>
```
vanbasten17, line 59: if this is a common usage. we can try to reduce it to something like `MultichannelCustomText` that already have these properties defined

packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  if (targetNode) {
+    const event: EventBotFaq = {
+      event_type: EventName.botFaq,
+      event_data: {
+        faq_name: targetNode.code,
+      },
+    }
+    await trackEvent(request, event)
+    return targetNode
+  }
```
vanbasten17, line 87: can this be simplified?

packages/botonic-plugin-flow-builder/src/action/intent.ts
```diff
@@ -0,0 +1,48 @@
...
+      intent: request.input.intent as string,
+      confidence: request.input.confidence as number,
+      confidence_successful: true,
+    },
+  }
+  if (isIntentValid(intentNode, request, cmsApi) && intentNode?.target?.id) {
+    await trackEvent(request, eventBotAiModel)
+    return cmsApi.getNodeById<HtNodeWithContent>(intentNode.target.id)
+  } else {
+    eventBotAiModel.event_data.confidence_successful = false
```
vanbasten17, line 32: beware modifying directly the attribute of the object unless this is the behavior you expect for this object. ow, you should use something to clone the data in order to have a "stable" object.

packages/botonic-plugin-flow-builder/src/action/keyword.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { FlowBuilderApi } from '../api'
+import { trackEvent } from './tracking'
+
+export async function getNodeByKeyword(
+  cmsApi: FlowBuilderApi,
+  locale: string,
+  request: ActionRequest,
+  userInput: string
+) {
```
vanbasten17, line 15: missing return type

packages/botonic-plugin-flow-builder/src/action/keyword.ts
```diff
@@ -0,0 +1,28 @@
...
+    event_type: EventName.botKeywordsModel,
+    event_data: {
+      confidence_successful: true,
+    },
+  }
+  if (keywordNode) {
+    await trackEvent(request, eventBotKeywordModel)
+    return keywordNode
+  }
+  return undefined
```
vanbasten17, line 27: nit: try it with early return?

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -0,0 +1,155 @@
...
+  url: string
+  flow: HtFlowBuilderData
+
+  private constructor() {}
+
+  static async create(options: FlowBuilderApiOptions): Promise<FlowBuilderApi> {
+    const newApi = new FlowBuilderApi()
+
+    newApi.url = options.url
+    newApi.flow = options.flow ?? (await newApi.getFlow(options.accessToken))
```
vanbasten17, line 27: I would create a method that given the object options, gets the flow, or directly (much better) passing both flow and access token

vanbasten17 (approved):
Looking good overall, I would try to make the PR smaller so it's easier to review.

packages/botonic-plugin-flow-builder/src/action/index.tsx
Iru89: Maybe using a ternary for the first if else?

packages/botonic-plugin-flow-builder/src/action/intent.ts
Iru89: I have done so because if the intentNode is valid (confidence higher than expected) I want to track with confidence_successful: true. Otherwise this attribute has to be false. 
This way I don't have to create two identical objects with only one attribute changing.

packages/botonic-plugin-flow-builder/src/api.ts
Iru89: Sorry Marc this is from another PR I didn't put the base branch correctly. Now it is correct. 
options.flow is the case where the flow is in a local json (defined inside the bot). If this is not defined, this is when the call is actually made with the access token.

packages/botonic-plugin-flow-builder/src/action/index.tsx
Iru89: This is only temporary until the whatsapp button list is working properly.

AinaVendrell (approved):

PR #2552, opened 2023-07-04
Trigger Button with notifications

Iru89:
## Description
Be able to show in the trigger button the number of received and unread messages. I consider a message unread when I receive it and the webchat is closed.

## Context
In local is not working well as the message sets the ack to 1 directly.
Display a notification with the number of unread messages. 
CustomTriggerButton
Captura de Pantalla 2023-07-04 a las 13 59 39
DefaultTriggerButton
DefaultTriggerButton

Each time the action of webchatReducer addMessageComponent is called if the webchat is closed and the message ack is 0, I increment the state of unreadMessages by 1. When I reset unreadMessages

The div displaying unread messages has className='trigger-notifications' to be able to adjust the css

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-react/src/webchat/hooks.js
+++ b/packages/botonic-react/src/webchat/hooks.js
@@ -64,6 +64,7 @@ export const webchatInitialState = {
   lastMessageUpdate: undefined,
   currentAttachment: undefined,
   jwt: null,
+  unreadMessages: 0,
 }
 
 export function useWebchat() {
@@ -71,14 +72,10 @@ export function useWebchat() {
     webchatReducer,
     webchatInitialState
   )
-  const [unreadMessages, setUnreadMessages] = useState(0)
-
   const addMessage = message =>
     webchatDispatch({ type: ADD_MESSAGE, payload: message })
-  const addMessageComponent = message => {
-    setUnreadMessages(value => value + 1)
+  const addMessageComponent = message =>
     webchatDispatch({ type: ADD_MESSAGE_COMPONENT, payload: message })
-  }
   const updateMessage = message =>
     webchatDispatch({ type: UPDATE_MESSAGE, payload: message })
   const updateReplies = replies =>
@@ -124,8 +121,6 @@ export function useWebchat() {
     })
 
   const toggleWebchat = toggle => {
-    console.log('toggleWebchat hook function')
-    setUnreadMessages(0)
     webchatDispatch({
       type: TOGGLE_WEBCHAT,
       payload: toggle,
@@ -203,7 +198,6 @@ export function useWebchat() {
     updateHandoff,
     updateTheme,
     updateDevSettings,
-    unreadMessages,
     toggleWebchat,
     toggleEmojiPicker,
     togglePersistentMenu,
--- a/packages/botonic-react/src/webchat/hooks.js
+++ b/packages/botonic-react/src/webchat/hooks.js
@@ -71,11 +71,14 @@ export function useWebchat() {
     webchatReducer,
     webchatInitialState
   )
+  const [unreadMessages, setUnreadMessages] = useState(0)
 
   const addMessage = message =>
     webchatDispatch({ type: ADD_MESSAGE, payload: message })
-  const addMessageComponent = message =>
+  const addMessageComponent = message => {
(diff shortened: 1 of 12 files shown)


packages/botonic-react/src/webchat/trigger-button/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    WEBCHAT.CUSTOM_PROPERTIES.triggerButtonStyle
+  )
+
+  const CustomTriggerButton = getThemeProperty(
+    WEBCHAT.CUSTOM_PROPERTIES.customTrigger,
+    undefined
+  )
+
+  const handleClick = (event: any) => {
+    //@ts-ignore
```
asastre, line 49: Why is this `//@ts-ignore` needed?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -918,14 +863,11 @@ export const Webchat = forwardRef((props, ref) => {
...
-            event.preventDefault()
-          }}
-        >
-          {triggerButton()}
-        </div>
+        <TriggerButton
+          theme={theme}
+          unreadMessages={unreadMessages}
+          webchatState={webchatState}
+        />
```
asastre, line 870: :clap:

packages/botonic-react/src/webchat/hooks.js
```diff
@@ -71,11 +71,14 @@ export function useWebchat() {
     webchatReducer,
     webchatInitialState
   )
+  const [unreadMessages, setUnreadMessages] = useState(0)
 
   const addMessage = message =>
     webchatDispatch({ type: ADD_MESSAGE, payload: message })
-  const addMessageComponent = message =>
+  const addMessageComponent = message => {
+    setUnreadMessages(value => value + 1)
```
asastre, line 79: Just as an observation, this is not 100% correct because not all messages will be unread messages and we're assuming this here.

I know this will work because everytime the webchat is closed this counter is set to 0, so no bubble will be displayed at first with the unread messages when the webchat is closed, but the counter will have a wrong value when the webchat is opened, am I right?

If we can check if the webchat is open/close at this point, maybe it's better to only increase the `unreadMessages` when the webchat is closed, what do you think?

packages/botonic-react/src/webchat/trigger-button/index.tsx
Iru89: the function was not well defined, that's why it was necessary. I have already corrected it.

packages/botonic-react/src/webchat/webchat.jsx
Iru89: With the latest refactor there is no need for a prop 🚀

packages/botonic-react/src/webchat/hooks.js
Iru89: Yes you are right, I have refactored all this logic to have inside the webchatstate the value of unreadMessages and increment it taking into account the ack value.

asastre (approved):
:clap: Very nice and clean refactor!! :rocket:

packages/botonic-react/src/webchat/trigger-button/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const handleClick = (event: any) => {
+    toggleWebchat(true)
+    event.preventDefault()
+  }
+
+  return (
+    <div
+      onClick={event => {
+        handleClick(event)
+      }}
```
AinaVendrell, line 49: extra: onClick={handleClick}

packages/botonic-react/src/webchat/trigger-button/styles.ts
```diff
@@ -0,0 +1,39 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 50%;
+  position: absolute;
+  right: 20px;
+  bottom: 80px;
+  width: 24px;
+  height: 24px;
+  background-color: #ff426f;
```
AinaVendrell, line 31: define the color inside COLORS constant

AinaVendrell (approved):

AinaVendrell:
check the test that are failing

PR #2553, opened 2023-07-04
Feat(flow-builder): allow having multiple buttons pointing to the same target #BLT-163

AinaVendrell:
## Description
- Allow having multiple buttons pointing to the same target on whatsapp

## Context
- Whatsapp buttons and rows on the same message must have unique id

- Add extra info on buttons and rows payload to allow having multiple buttons on the a message pointing the same target

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/content-fields/flow-element.tsx
+++ b/packages/botonic-plugin-flow-builder/src/content-fields/flow-element.tsx
@@ -36,7 +36,7 @@ export class FlowElement extends ContentFieldsBase {
         <Pic src={this.image} />
         <Title>{this.title}</Title>
         <Subtitle>{this.subtitle}</Subtitle>
-        {this.button?.renderButton()}
+        {this.button?.renderButton(0)}
       </Element>
     )
   }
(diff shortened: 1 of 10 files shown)


asastre (approved):

packages/botonic-plugin-flow-builder/src/content-fields/flow-element.tsx
```diff
@@ -36,7 +36,7 @@ export class FlowElement extends ContentFieldsBase {
         <Pic src={this.image} />
         <Title>{this.title}</Title>
         <Subtitle>{this.subtitle}</Subtitle>
-        {this.button?.renderButton()}
+        {this.button?.renderButton(0)}
```
Iru89, line 39: If you have more than one element inside a Carousel and you always pass the number 0 as buttonIndex, if two elements point to the same target they will be the same. Maybe this buttonIndex should come from the element number inside the Carousel. 
I know this can't happen in Whatsapp but the buttonIndex should be different.

Iru89 (approved):

packages/botonic-plugin-flow-builder/src/content-fields/flow-element.tsx
AinaVendrell: by now we only support on button per element in the flow builder, if we add the option of having more button we'll change it

PR #2559, opened 2023-07-10
Flow Builder: Add a variable in a text from Flow Builder

Iru89:
## Description
The objective is to be able to add a variable in a text from Flow Builder. 

## Context
For example if you put {language} in the text this part of the text will be replaced by the current language which will be stored in extra_data.language. If the variable does not exist in the session it will be left {language}.
It also performs the fallback function. If extra_data is not defined, it is set to isFirstFallbackOption = true.

I have made a refactor to have the request in the cmsApi class so it is possible to replace the variables in the FLowText.fromHubtypeCMS function. This way in the action we already have the final text and we don't have to do extra logic in the action.

replaceVariables is a static function of the FlowText class. This function receives the extraData object which is where the variable to be replaced must be defined. This variable can be an empty string. This function returns the replaced text.

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/action/index.tsx
+++ b/packages/botonic-plugin-flow-builder/src/action/index.tsx
@@ -8,7 +8,7 @@ import React from 'react'
 
 import { FlowBuilderApi } from '../api'
 import { SOURCE_INFO_SEPARATOR } from '../constants'
-import { FlowContent, FlowHandoff } from '../content-fields'
+import { FlowContent, FlowHandoff, FlowText } from '../content-fields'
 import { HtNodeWithContent } from '../content-fields/hubtype-fields'
 import { getFlowBuilderPlugin } from '../helpers'
 import { trackEvent } from './tracking'
@@ -43,11 +43,22 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
     ) as FlowHandoff
     if (handoffContent) await handoffContent.doHandoff(request)
 
-    const renderContents = contents.filter(content =>
+    const contentsWithoutHandoff = contents.filter(content =>
       content instanceof FlowHandoff ? false : true
     )
 
-    return { contents: renderContents }
+    const contentsWithVariables = contentsWithoutHandoff.map(content => {
+      if (content instanceof FlowText) {
+        console.log(content.text)
+        content.text = replaceVariables(
+          content.text,
+          request.session.user.extra_data
+        )
+      }
+      return content
+    })
+
+    return { contents: contentsWithVariables }
   }
 
   render(): JSX.Element | JSX.Element[] {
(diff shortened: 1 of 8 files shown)


packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -43,11 +43,22 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
...
 
-    const renderContents = contents.filter(content =>
+    const contentsWithoutHandoff = contents.filter(content =>
       content instanceof FlowHandoff ? false : true
     )
 
-    return { contents: renderContents }
+    const contentsWithVariables = contentsWithoutHandoff.map(content => {
+      if (content instanceof FlowText) {
+        console.log(content.text)
```
AinaVendrell, line 52: remove console log

packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -103,3 +117,17 @@ async function getFallbackNode(cmsApi: FlowBuilderApi, request: ActionRequest) {
...
+  if (matches && extraData) {
+    matches.forEach(match => {
+      const variable = match.slice(1, -1)
+      const value = extraData[variable] ? extraData[variable] : match
+      text = text.replace(match, value)
+    })
+  }
+
+  return text
+}
```
AinaVendrell, line 133: maybe this can be moved inside the `FlowText` class

vanbasten17: > maybe this can be moved inside the `FlowText` class

yup, or create a function utils for different components. If it is only needed for this component, then makes sense to have it within this class out of the scope of the class

packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -43,11 +43,18 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
...
-    const renderContents = contents.filter(content =>
+    const contentsWithoutHandoff = contents.filter(content =>
       content instanceof FlowHandoff ? false : true
     )
 
-    return { contents: renderContents }
+    const contentsWithVariables = contentsWithoutHandoff.map(content => {
+      if (content instanceof FlowText) {
+        content.text = content.replaceVariables(request.session.user.extra_data)
+      }
```
asastre, line 53: Change: I think this should be called inside the `fromHubtypeCMS` method of the `FlowText` class. As @AinaVendrell all these logic should be placed and executed inside the `FlowText` class, the action should not be aware of this logic.

To make it work you will have to pass the request to the `FlowBuilderApi` class, maybe you can add it as an argument of the constructor.

asastre (changes requested):

AinaVendrell (approved):

PR #2588, opened 2023-07-20
Flow builder/remove hubtype analytics dependency

Iru89:
## Description
Remove hubtype analytics dependency from flow builder plugin

## Context
I have copied the same enum with the event names that the hubtype analytics plugin has.
I have also added the export that the plugin builder uses in the index so it is easier to import into the bot when initialising the flow builder plugin.

## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-plugin-flow-builder/src/action/user-input.ts
```diff
@@ -17,10 +13,7 @@ export async function getNodeByUserInput(
 ): Promise<HtNodeWithContent | undefined> {
   if (request.session.is_first_interaction) {
     const startNode = cmsApi.getStartNode()
-    const event: EventBotStart = {
-      event_type: EventName.botStart,
-    }
-    await trackEvent(request, event)
+    await trackEvent(request, EventName.botStart)
```
asastre, line 16: Doubt: Taking a look at this change, where you're removing the event object of being sent because it only contained the same value as the event name itself, I think that maybe it makes no sense to send the event with the actual format from the Flow Builder.

Taking a look at the main event sent in the action (`EventName.botFaq`), does it make sense to send this object with the `event_type` key in it?
```
const event = {
  event_type: EventName.botFaq,
  event_data: {
    faq_name: targetNode.code,
  },
}
```
Wouldn't it be better to send only this?
`const event = { faq_name: targetNode.code }`

The `event_type` and `event_data` keys are part of the Hubtype Analytics plugin, and if we want to detach it from the Flow Builder I think that it make sense to remove all these keys and leave only the needed ones here.

asastre (commented):
All good, but I have a doubt that I would like to clarify before approving this PR

packages/botonic-plugin-flow-builder/src/action/user-input.ts
Iru89: Yes, I think it would be more decoupled if we do this

packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -79,13 +74,10 @@ async function getTargetNode(
     : (cmsApi.getNodeById(contentId) as HtNodeWithContent)
 
   if (targetNode) {
-    const event: EventBotFaq = {
-      event_type: EventName.botFaq,
-      event_data: {
-        faq_name: targetNode.code,
-      },
+    const event = {
```
AinaVendrell, line 77: suggestion: instead of `event` this could be named `eventArgs` or `eventProps`

AinaVendrell (approved):

PR #2602, opened 2023-08-14
botonic-react: render images sent to the bot by user

Iru89:
## Description
If the user sends an image through the bot it is not rendered correctly. This fails because staticAsset only renders images from a public URL. 

## Context
User attaches an image to the webchat
(first commit only works in local) In order to render an image within a message, add the `bolb` format in the isURL function which is used by staticAsset.
(second commit) To make the first render after attaching an image you can use the input.data which contains the base 64 string of the image. If you reload the webchat this string will no longer be used , in the localStorage message the url where the image was uploaded is saved. 

Captura de pantalla 2023-08-14 a las 11 19 12

### DIFF
--- a/packages/botonic-react/src/components/image.jsx
+++ b/packages/botonic-react/src/components/image.jsx
@@ -20,7 +20,9 @@ const serialize = imageProps => {
 }
 
 export const Image = props => {
-  props = { ...props, src: staticAsset(props.src) }
+  props = props.input?.data
+    ? { ...props, src: props.input.data }
+    : { ...props, src: staticAsset(props.src) }
   let content = props.children
 
   const [isPreviewerOpened, setIsPreviewerOpened] = useState(false)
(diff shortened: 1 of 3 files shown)


alexgimenez02 (approved):

vanbasten17 (approved):

packages/botonic-react/src/components/image.jsx
```diff
@@ -20,7 +20,10 @@ const serialize = imageProps => {
 }
 
 export const Image = props => {
-  props = { ...props, src: staticAsset(props.src) }
+  console.log('Image', { props, src: props.src, input: props.input })
```
vanbasten17, line 23: Remove console.log

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -491,13 +491,16 @@ export const Webchat = forwardRef((props, ref) => {
       )
     } else if (isMedia(input)) {
       const temporaryDisplayUrl = URL.createObjectURL(input.data)
+      console.log({ input })
```
vanbasten17, line 494: Beware leaving lots 🙈

vanbasten17 (approved):
Looking good to me!

PR #2606, opened 2023-08-17
Botonic react: multichannel text use whatsapp button list

Iru89:
## Description
Until now, when a text with more than 3 buttons was sent to the multichannel and the bot was executed in whatsapp, several messages were created and each one of them had 3 buttons. From now on we will use a WhatsApp button list.

## Context
If a text arrives with more than 3 whatsapp buttons a whatsapp button list will be created. But in a whatsapp button list you can't use buttons pointing to urls or webviews.
A prop is added to the Multichannel, in the same way that a string could already be passed to split messages ("More Options" by default). Now you can pass a prop to define the text of the WhatsappButtonList button ("Menu" by default).

A whatsapp button list will be created for each group of 4 to 10 buttons that have a payload or a path. 
If there is a group of less than 4 buttons left, a message text will be created with the missing buttons. 
If the initial text had buttons that point to urls an extra message will be added with the urls (they are not buttons but can be clicked to go to the website).
If the initial text had buttons with webviews a message will be added with the urls that lead to the webview. 
Captura de pantalla 2023-08-17 a las 16 57 13

If you pass a text component like this to the multichannel
Captura de pantalla 2023-08-17 a las 16 51 44
the result in whatsapp will look like this
Captura de pantalla 2023-08-17 a las 16 53 04

## Testing
Update the multichannel-text.test.jsx to check that a WhatsappButtonList is used when there are more than 3 buttons.

### DIFF
--- a/packages/botonic-react/src/components/multichannel/multichannel-text.jsx
+++ b/packages/botonic-react/src/components/multichannel/multichannel-text.jsx
@@ -81,7 +81,7 @@ export const MultichannelText = props => {
       const asText =
         type === buttonTypes.POSTBACK ? postbackButtonsAsText : true
       const newline =
-        multichannelContext.messageSeparator === null &&
+        multichannelContext.messageSeparator == null &&
         !newLineFirstButton &&
         i === 0
           ? ''
@@ -113,7 +113,6 @@ export const MultichannelText = props => {
     return messages
   }
 
-  // START WHATSAPP LOGIC
   if (isWhatsapp(requestContext)) {
     const texts = getText(props.children)
     const { postbackButtons, urlButtons, webviewButtons } = getWhatsappButtons()
@@ -130,11 +129,10 @@ export const MultichannelText = props => {
     const buttonsTextSeparator =
       props.buttonsTextSeparator || DEFAULT_WHATSAPP_MAX_BUTTON_SEPARATOR
 
-    //MORE THAN 3 BUTTONS
-    if (
-      !postbackButtonsAsText &&
-      postbackButtons.length > WHATSAPP_MAX_BUTTONS
-    ) {
+    const exceedWhatsAppMaxButtonNumber =
+      !postbackButtonsAsText && postbackButtons.length > WHATSAPP_MAX_BUTTONS
+
+    if (exceedWhatsAppMaxButtonNumber) {
       const menuButtonTextWhatsappList =
         props.menuButtonTextWhatsappList || MENU_BUTTON_WHATSAPP_BUTTON_LIST
 
@@ -220,14 +218,11 @@ export const MultichannelText = props => {
               </Text>
             )
           })}
-          {urlButtonElements.length ? messageWithUrlButtonElements : null}
-          {webviewButtonElements.length
-            ? messageWithWebviewButtonElements
-            : null}
+          {urlButtonElements.length > 0 && messageWithUrlButtonElements}
+          {webviewButtonElements.length > 0 && messageWithWebviewButtonElements}
         </>
       )
     }
-    // END LOGIC WHATSAPP WITH MORE THAN 3 BUTTONS
 
     multichannelContext.currentIndex = getDefaultIndex()
     const postbackButtonElements = postbackButtons.map(
@@ -261,9 +256,7 @@ export const MultichannelText = props => {
 
     return <>{messages}</>
   }
-  // END WHATSAPP LOGIC
 
-  // START FACEBOOK LOGIC
   if (isFacebook(requestContext)) {
     const text = getText(props.children)
     const multichannelFacebook = new MultichannelFacebook()
--- a/packages/botonic-react/src/components/multichannel/multichannel-text.jsx
+++ b/packages/botonic-react/src/components/multichannel/multichannel-text.jsx
@@ -81,7 +81,7 @@ export const MultichannelText = props => {
       const asText =
         type === buttonTypes.POSTBACK ? postbackButtonsAsText : true
       const newline =
-        multichannelContext.messageSeparator == null &&
+        multichannelContext.messageSeparator === null &&
         !newLineFirstButton &&
         i === 0
           ? ''
@@ -161,9 +161,12 @@ export const MultichannelText = props => {
           }
           const rows = postbackButtons.map(postbackButton => {
             const row = {
-              id: postbackButton.props.payload,
+              id: postbackButton.props.path
+                ? `__PATH_PAYLOAD__${postbackButton.props.path}`
+                : postbackButton.props.payload,
               title: postbackButton.props.children,
             }
+            console.log({ row })
             return row
           })
           const whatsbuttonlistProps = {
--- a/packages/botonic-react/src/components/multichannel/multichannel-text.jsx
+++ b/packages/botonic-react/src/components/multichannel/multichannel-text.jsx
@@ -153,7 +153,7 @@ export const MultichannelText = props => {
 
       const messages = messagesPostbackButtonList.map(
         (postbackButtons, index) => {
-          if (postbackButtons.length < 4) {
+          if (postbackButtons.length <= WHATSAPP_MAX_BUTTONS) {
             return {
               type: INPUT.TEXT,
               children: [...buttonsTextSeparator, ...postbackButtons],
--- a/packages/botonic-react/src/components/multichannel/multichannel-text.jsx
+++ b/packages/botonic-react/src/components/multichannel/multichannel-text.jsx
@@ -1,6 +1,8 @@
+import { INPUT } from '@botonic/core/lib/esm/models/legacy-types'
 import React, { useContext } from 'react'
 
 import { RequestContext } from '../../contexts'
+import { WhatsappButtonList } from '..'
 import { Text } from '../text'
 import { MultichannelFacebook } from './facebook/facebook'
 import { MultichannelButton } from './multichannel-button'
@@ -16,7 +18,9 @@ import {
   getMultichannelReplies,
   isFacebook,
   isWhatsapp,
+  MENU_BUTOON_WHATSAPP_BUTTON_LIST,
   MULTICHANNEL_WHATSAPP_PROPS,
+  WHATSAPP_LIST_MAX_BUTTONS,
   WHATSAPP_MAX_BUTTONS,
 } from './multichannel-utils'
 import { whatsappMarkdown } from './whatsapp/markdown'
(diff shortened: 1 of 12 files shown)


vanbasten17 (commented):
Go ahead with WBL! 💯

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -96,14 +103,19 @@ export const MultichannelText = props => {
...
+      let i = 0;
+      i < postbackButtons.length;
+      i += WHATSAPP_LIST_MAX_BUTTONS
+    ) {
+      messages.push(postbackButtons.slice(i, i + WHATSAPP_LIST_MAX_BUTTONS))
     }
     return messages
   }
 
+  // START WHATSAPP LOGIC
```
AlbertGom, line 118: Why is this comment added here?

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -120,45 +132,104 @@ export const MultichannelText = props => {
     const buttonsTextSeparator =
       props.buttonsTextSeparator || DEFAULT_WHATSAPP_MAX_BUTTON_SEPARATOR
 
+    //MORE THAN 3 BUTTONS
```
AlbertGom, line 135: Why is this comment necessary?

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -120,45 +132,104 @@ export const MultichannelText = props => {
...
     if (
       !postbackButtonsAsText &&
       postbackButtons.length > WHATSAPP_MAX_BUTTONS
     ) {
+      const menuButtonTextWhatsappList =
+        props.menuButtonTextWhatsappList || MENU_BUTOON_WHATSAPP_BUTTON_LIST
```
AlbertGom, line 141: typo: `MENU_BUTOON_WHATSAPP_BUTTON_LIST` -> `MENU_BUTTON_WHATSAPP_BUTTON_LIST`

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -120,45 +132,104 @@ export const MultichannelText = props => {
...
-          return [].concat(
-            ...textElements,
-            ...urlButtonElements,
-            ...postbackButtons
-          )
-        } else {
-          return [].concat(buttonsTextSeparator, ...postbackButtons)
+      const messages = messagesPostbackButtonList.map(
+        (postbackButtons, index) => {
+          if (postbackButtons.length < 4) {
```
AlbertGom, line 156: Change this for `postbackButtons.length <= WHATSAPP_MAX_BUTTONS` ?

packages/botonic-react/src/components/multichannel/multichannel-utils.js
```diff
@@ -8,8 +8,10 @@ import { Providers } from '@botonic/core'
 export const MULTICHANNEL_WHATSAPP_PROPS = { markdown: false }
 
 export const WHATSAPP_MAX_BUTTONS = 3
+export const WHATSAPP_LIST_MAX_BUTTONS = 10
 export const WHATSAPP_MAX_BUTTON_CHARS = 20
 export const DEFAULT_WHATSAPP_MAX_BUTTON_SEPARATOR = 'More options:'
+export const MENU_BUTOON_WHATSAPP_BUTTON_LIST = 'Menu'
```
AlbertGom, line 14: Isn't it better to set the default text as `Show options`?

AlbertGom (commented):
Questions/comments:

- The codecov plugin is complaining because some added lines are not covered by tests. I would work more on the tests to cover these new lines.
- Is it possible to add sections?

packages/botonic-react/src/components/multichannel/multichannel-utils.js
Iru89: ok

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
Iru89: It is not necessary but it helps to see where the logic begins. I would like to make a refactor to extract this into a MultichannelWhatsappText component.

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
Iru89: Same as previous comment

Iru89:
> Questions/comments:
> 
> * The codecov plugin is complaining because some added lines are not covered by tests. I would work more on the tests to cover these new lines.
> * Is it possible to add sections?

At the moment it is not possible to create sections. Please note that this is so that a text with more than 3 buttons is rendered correctly in WhatsApp. If you have to use a whtasapp button list in the bot, you can use the Bototnic component WhatsappButtonList where you can define sections and description in each row.

AlbertGom (approved):

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -77,11 +81,14 @@ export const MultichannelText = props => {
       const asText =
         type === buttonTypes.POSTBACK ? postbackButtonsAsText : true
       const newline =
-        multichannelContext.messageSeparator == null &&
+        multichannelContext.messageSeparator === null &&
```
asastre, line 84: Observation/Change: Adding the `===`, if the `messageSeparator` is `undefined` the condition will return `false` and I think this is not the wanted behaviour. 
I would leave it as == or check both values `null` and `undefined`.

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -77,11 +81,14 @@ export const MultichannelText = props => {
...
         !newLineFirstButton &&
         i === 0
           ? ''
           : '\n'
+      if (type !== buttonTypes.POSTBACK && multichannelButton.props.payload) {
+        delete multichannelButton.props.payload
+      }
```
asastre, line 91: Doubt: When will this happen? Under what circumstances can we have a non POSTBACK button with a payload?

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -96,14 +103,19 @@ export const MultichannelText = props => {
...
+      i < postbackButtons.length;
+      i += WHATSAPP_LIST_MAX_BUTTONS
+    ) {
+      messages.push(postbackButtons.slice(i, i + WHATSAPP_LIST_MAX_BUTTONS))
     }
     return messages
   }
 
+  // START WHATSAPP LOGIC
   if (isWhatsapp(requestContext)) {
```
asastre, line 119: Remove comment? I think the condition is clear enoguh to know we will enter the WhatsApp logic.

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -120,45 +132,104 @@ export const MultichannelText = props => {
     const buttonsTextSeparator =
       props.buttonsTextSeparator || DEFAULT_WHATSAPP_MAX_BUTTON_SEPARATOR
 
+    //MORE THAN 3 BUTTONS
     if (
       !postbackButtonsAsText &&
       postbackButtons.length > WHATSAPP_MAX_BUTTONS
     ) {
```
asastre, line 139: Instead of adding a comment, why not creating an `exceedWhatsAppMaxButtonNumber` (or something like this) function that executes this condition?

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -120,45 +132,104 @@ export const MultichannelText = props => {
...
+                {...props}
+              >
+                {message.children}
+              </Text>
+            )
+          })}
+          {urlButtonElements.length ? messageWithUrlButtonElements : null}
+          {webviewButtonElements.length
+            ? messageWithWebviewButtonElements
+            : null}
```
asastre, line 228: Suggestion: Better use `&&`?
```suggestion
          {urlButtonElements.length && messageWithUrlButtonElements}
          {webviewButtonElements.length && messageWithWebviewButtonElements}
```

asastre (changes requested):
Just one small change that I think may affect the actual behaviour, the rest are just comments/doubts/suggestions

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
Iru89: Yes you are right. I'm changing the logic with this triple equal...

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
Iru89: Now with flow builder we are adding in all buttons a part at the end of the payload. To make sure that all payloads of a text or whatsapp button list are different. What happens is that it is always added if the payload is empty it is also added, maybe we should change this logic.

Maybe we should change this logic so that if it is a button with url but without payload is not added this final part of the payload.

```typescript
return (
      <Button
        key={this.id}
        url={this.url}
        payload={this.getButtonPayload(buttonIndex)}
      >
        {this.text}
      </Button>
    )
  }

  private getButtonPayload(buttonIndex: number): string {
    return `${this.payload}${SOURCE_INFO_SEPARATOR}${buttonIndex}`
  }
```

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
Iru89: ok. I create a exceedWhatsAppMaxButtonNumber

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
Iru89: Yes, but I should change the condition to length>0 because if it is an empty array it would render with 0 &&. message.....

Iru89: applied

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
Iru89: I have modified this logic in the flow builder plugin

asastre: Yes, we have to adapt this logic in the Flow Builder's plugin. The code of a Botonic plugin should never affect the code of Botonic, the plugins have to be the ones to adapt to Botonic's code.

packages/botonic-react/src/components/multichannel/multichannel-text.jsx
asastre: You're right!

asastre (approved):

PR #2608, opened 2023-08-21
Botonic react: WebchatState and webchatContext with typescript

Iru89:
## Description
In this PR the files related to the WebchatState with its two recudecrs (webchatReducer and the messageReducer) and the WebchatContext with the hook useWebchat are passed to typescript.
Apply a refactor to create a hooks folder and have each hook in a separate file. All are exported from hooks/inde.ts

Set the typescript version to 5.0.4 if we upgrade to 5.2 we have problems with the lint and it is also necessary to use in tsconfig.json "module": "NodeNext", "moduleResolution": "NodeNext". I will check it in the future.

## Context
The intention is to move all the code from botonic-react to typescript. 

Use the types already defined in the index-types files to write in typescript the above mentioned files


vanbasten17 (approved):
Looking good overall!! Have you tested it at least manually? We should ensure nothing is broken

Iru89:
> Looking good overall!! Have you tested it at least manually? We should ensure nothing is broken

Yes, I have tested it locally and deployed it in a bot.

PR #2610, opened 2023-09-05
Knowledge base plugin and inject to flow builder plugin 

Iru89:
## Description
**@botonic/plugin-knowledge-base**
This new plugin allows you to connect to a previously uploaded dashboard knowledge base and then get answers about this document.

**@botonic/plugin-flow-builder**
You can pass the function to get the response from the knowledge base through the constructor. If in the fallback node you have activated the knowledge base, and it gives a correct response. The flow builder plugin will create a node to respond with the information obtained by the knowledge base plus the associated followup.

## Context
**@botonic/plugin-knowledge-base**
This plugin has the logic to make the call to our backend api using the authToken to get the response from a knowledge base. In the constructor it is necessary to pass the id of the knowledge base and the host where the call is made https:api.hubtype.com/ for production, you can also pass the authToken if you are developing locally and a timeout (usually 10 seconds).   
Although it is a plugin it doesn't do anything in the pre or post function, it only has a getIaResponse function that allows you to get the response from the knowledge base. This response is an object of the following type  
```typescript
interface KnowledgebaseResponse {
    ai: string
    hasKnowledge: boolean
    sources: string[]
}
```
The parameter ai will have the answer and the hasKnowledge indicates if this answer is related to the documents uploaded to the dashboard, sources are the names of the documents used to get the answer.

**@botonic/plugin-flow-builder**
In the flow builder plugin there is no direct dependency on the knowledge base plugin. You pass the function through the builder and plug everything back in there. You add the logic to use the knowledge base just before responding with the fallback.

First of all you need to add these two plugins independently in the bot where you want to use them.

config.ts
```typescript
const knowledgeBase: PluginKnowledgeBaseOptions = {
  knowledgeBaseId: '50976b51-111-22ec-9bef-8888d25f2c5e',
  host: 'https://api.hubtype.com/',
  authToken: '51cf4b3b8316a02123456df789ef01',  // Used locally,
}
const flowBuilder:  BotonicPluginFlowBuilderOptions = {
    flowUrl: isProduction(env)
      ? 'https://api.ent0.flowbuilder.prod.hubtype.com/flow/latest'
      : 'https://api.dev.flowbuilder.dev.hubtype.com/flow/draft',
    getLocale: (session: Session) => context(session as BotSession).locale,
    getAccessToken: () => 'f3f123ab12345cde6789f03628d7fe', // Used locally,
    getKnowledgeResponse: async (request: BotRequest) => {
      const knowladgeBasePlugin = request.plugins
        .BotonicPluginKnowledgeBases as BotonicPluginKnowledgeBase
      return await knowladgeBasePlugin.getIaResponse(request.session)
    },
  }

export const config = {
    knowledgeBase,
    flowBuilder
}
```

plugins.ts
```typescript
export const plugins: core.PluginConfig<any>[] = [
  {
    id: 'BotonicPluginFlowBuilder',
    resolve: require('@botonic/plugin-flow-builder'),
    options: config.flowBuilder,
  },
  {
    id: 'BotonicPluginKnowledgeBases',
    resolve: require('@botonic/plugin-knowledge-bases'),
    options: config.knowledgeBase,
  }
]
```

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/action/index.tsx
+++ b/packages/botonic-plugin-flow-builder/src/action/index.tsx
@@ -6,6 +6,7 @@ import { SOURCE_INFO_SEPARATOR } from '../constants'
 import { FlowContent, FlowHandoff } from '../content-fields'
 import { HtNodeWithContent } from '../content-fields/hubtype-fields'
 import { getFlowBuilderPlugin } from '../helpers'
+import { createKnowledgeNode } from './knowledge-bases'
 import { EventName, trackEvent } from './tracking'
 import { getNodeByUserInput } from './user-input'
 
@@ -87,6 +88,14 @@ async function getFallbackNode(cmsApi: FlowBuilderApi, request: ActionRequest) {
   if (!request.session.user.extra_data) {
     request.session.user.extra_data = { isFirstFallbackOption: true }
   }
+
+  const knowladgeNode = await createKnowledgeNode(cmsApi, request)
+  if (knowladgeNode) {
+    // TODO: add event
+    // await trackEvent(request, 'KnowledgeBase response')
+    return knowladgeNode
+  }
+
   const isFirstFallbackOption =
     !!request.session.user.extra_data.isFirstFallbackOption
   const fallbackNode = cmsApi.getFallbackNode(isFirstFallbackOption)
--- a/packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
+++ b/packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
@@ -1,0 +1,37 @@
+import axios, { AxiosResponse } from 'axios'
+
+const DEFAULT_TIMEOUT = 10000
+
+export class HubtypeApiService {
+  private knowledgeBaseId: string
+  private host: string
+  private timeout: number
+
+  constructor(knowledgeBaseId: string, host: string, timeout?: number) {
+    this.knowledgeBaseId = knowledgeBaseId
+    this.host = host
+    this.timeout = timeout || DEFAULT_TIMEOUT
+  }
+
+  async inference(
+    authToken: string,
+    chatId: string
+  ): Promise<
+    AxiosResponse<{
+      ai: string
+      has_knowledge: boolean
+      sources: string[]
+    }>
+  > {
+    return await axios({
+      method: 'POST',
+      url: `${this.host}v1/ai/knowledge_bases/${this.knowledgeBaseId}/inference/`,
+      headers: {
+        Authorization: `Bearer ${authToken}`,
+        'Content-Type': 'application/json',
+      },
+      data: { chat_id: chatId },
+      timeout: this.timeout,
+    })
+  }
+}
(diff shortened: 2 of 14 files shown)


packages/botonic-plugin-knowledge-bases/src/index.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.authToken = options.authToken || ''
+  }
+
+  async pre(_request: PluginPreRequest): Promise<void> {
+    return
+  }
+
+  async getIaResponse(session: HubtypeSession): Promise<KnowledgebaseResponse> {
+    try {
+      console.log({ isProd })
```
MarcosCA, line 27: all these console logs are needed?

MarcosCA (approved):

packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -87,6 +88,14 @@ async function getFallbackNode(cmsApi: FlowBuilderApi, request: ActionRequest) {
   if (!request.session.user.extra_data) {
     request.session.user.extra_data = { isFirstFallbackOption: true }
   }
+
+  const knowladgeNode = await createKnowledgeNode(cmsApi, request)
```
vanbasten17, line 92: typo: rename them to knowledge instead of knowladge

packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
```diff
@@ -0,0 +1,37 @@
+import axios, { AxiosResponse } from 'axios'
+
+const DEFAULT_TIMEOUT = 10000
```
vanbasten17, line 3: are 10 seconds acceptable for our use case? which are the timings approximately of the calls to the knowledge api?

packages/botonic-plugin-knowledge-bases/src/index.ts
```diff
@@ -0,0 +1,50 @@
...
+  constructor(options: PluginKnowledgeBaseOptions) {
+    this.apiService = new HubtypeApiService(
+      options.knowledgeBaseId,
+      options.host,
+      options.timeout
+    )
+    this.authToken = options.authToken || ''
+  }
+
+  async pre(_request: PluginPreRequest): Promise<void> {
```
vanbasten17, line 21: with latests updates of Plugin interface no need to define empty `pre` or `post` calls

vanbasten17 (approved):

packages/botonic-plugin-knowledge-bases/src/index.ts
Iru89: nop, I will remove them

packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
Iru89: the api is slow, it takes between 6 and 10 seconds, at user level it is slow when the knowledge base responds.

PR #2611, opened 2023-09-06
Botonic-dx: update some packages version to stop using * #SER-2611

alexgimenez02:
## Description
Some packages (does common with `botonic-esling-config` were added with the `*` value in the version field. Additionally bumped the version to `0.21.1-alpha.1`.

## Context
Some bots that used botonic DX started to fail when running the linters and the github actions for those bots.

Checking the versions of `botonic-eslint-config` and adding the same version that those packages had.

## Testing

The pull request has no tests


Iru89 (approved):

vanbasten17 (approved):

alexgimenez02:
This PR is discarded to not being the optimal solution to the problem, but just a temporary solution found. The PR #2615 solves the same issue correctly

PR #2618, opened 2023-09-18
Plugin Flow Builder: track knowledge base event

Iru89:
## Description
Plugin-flow-builder and plugin-knowledge-bases:
- Change the type of sources in the knowledge base inference response (it was not typed correctly). The attribute sources is an array of objects, not a string array as it was until now.

Plugin-flow-builder:
- Adds botAiKnowladgeBase event tracking when the bot response is obtained from the inference with the knowledge base
- Refactor: a small problem with the suffix on payloads. When adding a suffix to all payloads when creating a button in flow builder, in the routes we don't always have the expected payload. To solve this I remove this suffix in the pre function of the plugin, this way in the routes we have the "clean" payload. 

## Testing

The pull request...

- has unit tests
- has integration tests


elozano98 (changes requested):
Replace all Knowledge* naming prefixes by KnowledgeBase*

Iru89:
> Replace all Knowledge* naming prefixes by KnowledgeBase*

I have also renamed inside the plugin-knowledge-bases the function getIaResponse to getInference. 
This way I think it's clearer to use-it will look like `pluginKnowledgeBases.getInference(session)`

elozano98 (approved):

MarcosCA (approved):

PR #2620, opened 2023-09-22
Feature: deploy bot in prod/qa/dev

Iru89:
## Description
- Deploy bot in prod/qa/dev with with environment variables set hubtype/infrastructure repository deploy_bot_etc functions.
- Makes a fix to the isBrowser function to solve a problem detected a few days ago in some websites where process.env existed inside the browser.

## Context
The getWebpackEnvVar function did not work correctly and ended up always putting the production url and the production pusher key

instead of using the getWebpackEnvVar function, I read directly process.env and if this is undefined, the HUBTYPE_API_URL or PUSHER_KEY of the production will be used.

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-core/src/hubtype-service.ts
+++ b/packages/botonic-core/src/hubtype-service.ts
@@ -96,10 +96,6 @@ export class HubtypeService {
       // TODO recover user & appId somehow
       return Promise.reject('No User or appId. Clear cache and reload')
     }
-    console.log('Initializing Pusher', {
-      _WEBCHAT_PUSHER_KEY_: WEBCHAT_PUSHER_KEY,
-      _HUBTYPE_API_URL_: HUBTYPE_API_URL,
-    })
     this.pusher = new Pusher(WEBCHAT_PUSHER_KEY, {
       cluster: 'eu',
       authEndpoint: `${HUBTYPE_API_URL}/v1/provider_accounts/webhooks/webchat/${this.appId}/auth/`,
--- a/packages/botonic-core/src/hubtype-service.ts
+++ b/packages/botonic-core/src/hubtype-service.ts
@@ -30,7 +30,7 @@ interface HubtypeServiceArgs {
 }
 
 const WEBCHAT_PUSHER_KEY =
-  process.env.WEBCHAT_PUSHER_KEY || '434ca667c8e6cb3f641c'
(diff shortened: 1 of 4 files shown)


packages/botonic-core/src/hubtype-service.ts
```diff
@@ -30,18 +29,10 @@ interface HubtypeServiceArgs {
...
-  '434ca667c8e6cb3f641c' // pragma: allowlist secret
-)
-const _HUBTYPE_API_URL_ = getWebpackEnvVar(
-  // @ts-ignore
-  typeof HUBTYPE_API_URL !== 'undefined' && HUBTYPE_API_URL,
-  'HUBTYPE_API_URL',
-  'https://api.hubtype.com'
-)
+const WEBCHAT_PUSHER_KEY =
+  process.env.WEBCHAT_PUSHER_KEY || '434ca667c8e6cb3f641c'
```
asastre, line 33: Doubt: Should we still add the `// pragma: allowlist secret` comment with this pusher key to avoid security warnings?

packages/botonic-core/src/hubtype-service.ts
```diff
@@ -105,9 +96,13 @@ export class HubtypeService {
       // TODO recover user & appId somehow
       return Promise.reject('No User or appId. Clear cache and reload')
     }
-    this.pusher = new Pusher(_WEBCHAT_PUSHER_KEY_, {
+    console.log('Initializing Pusher', {
+      _WEBCHAT_PUSHER_KEY_: WEBCHAT_PUSHER_KEY,
+      _HUBTYPE_API_URL_: HUBTYPE_API_URL,
+    })
```
asastre, line 102: Remove log

packages/botonic-plugin-flow-builder/src/functions/conditional-queue-status.ts
```diff
@@ -37,8 +29,11 @@ interface AvailabilityData {
 export async function getQueueAvailability(
   queueId: string
 ): Promise<AvailabilityData> {
+  console.log('plugin-flow-builder getQueueAvailability', {
+    _HUBTYPE_API_URL_: HUBTYPE_API_URL,
+  })
```
asastre, line 34: Same here, remove log

asastre (approved):

PR #2621, opened 2023-09-26
Plugin knowledge base: use external api

Iru89:
## Description
plugin-knowledge-bases:
- Use the end point of the external api
- In the response the attribute ai has been renamed to answer
plugin-flow-builder:
- In the knowledgeBaseResponse attribute ai has been renamed to answer

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
+++ b/packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
@@ -4,11 +4,11 @@ import axios, { AxiosResponse } from 'axios'
 const DEFAULT_TIMEOUT = 10000
 
 export class HubtypeApiService {
-  private knowledgeBaseId: string
   private host: string
+  private knowledgeBaseId: string
   private timeout: number
 
-  constructor(knowledgeBaseId: string, host: string, timeout?: number) {
+  constructor(host: string, knowledgeBaseId: string, timeout?: number) {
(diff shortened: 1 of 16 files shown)


packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
```diff
@@ -19,7 +19,7 @@ export class HubtypeApiService {
     chatId: string
   ): Promise<
     AxiosResponse<{
-      ai: string
+      answer: string
       has_knowledge: boolean
       sources: {
         knowledge_source_id: string
```
asastre, line 25: Doubt: Why not creating a `type` or `interface` with the API response the same way it's done in the Flow Builder plugin? 
In fact, I think that the interface should be declared and exported in the Knowledge base plugin and used it in the Flow Builder plugin.

asastre (commented):
Just one question before approving, the rest LGTM

packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
Iru89: I have not created this interface because it is in snake_case and it is only typed here like this because when you use the plugin everything is written in camelCase and does the change internally.
You can't do this import either because the  plugin-flow-builder does not have the plugin-knowledge-bases dependency.

asastre (approved):

PR #1063, opened 2020-11-04
Fix/ fix react warnings, log deprecated props and refactor references

vanbasten17:
☢️ **Please, review it commit by commit for your ease.**

## Description
**Fix warnings:** 
* `Warning: Received 'true' for non-boolean attribute '...'`
* `Warning: React does not recognize the 'enableTimestamps' prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase 'enabletimestamps' instead.` (also for `blobWidth`)

**Fix missing call to function:** 
* Add missing `stringifyWithRegexs` call when saving webchat state (affecting `WebchatSettings` feature)

**Refactored:**
* Refactored constants for `WEBCHAT.CUSTOM_PROPERTIES`.
* Refactored constants using `SENDER` object.
* Refactored constants using `ROLES` object.

**Add a way to warn about deprecated props** 
Screenshot 2020-11-04 at 16 36 28

## Context
Several annoying warnings when running tests. Take profit to do some refactor.

## To document / Usage example
When using customMessage wrapper, encourage use of `enabletimestamps` and `imagestyle`.

## Testing

The pull request...

- has unit tests
- has integration tests
- doesn't need tests because... **[provide a description]**

PR #2666, opened 2023-11-13
Change message image when speaking with the bot or the agent

Iru89:
## Description
To be able to define a different image for the messages sent by the agent

## Context
Until now we could not define a different image for the messages sent by the agent, what we did was to change all the images of the chat while the user was talking with an agent, but when the handoff was finished all the messages had the image of the bot again.

- an image can be defined in the custom property `agentMessageImage`
- this is used when a message arrives with the sentBy = agent
- the css class of the timestamp will be `botanic-timestamp-agent` to be able to define another css or another image in that div (for now this is done with css, in the future it would be interesting to be a custom component)


packages/botonic-react/src/components/message/timestamps.jsx
```diff
@@ -0,0 +1,62 @@
...
+    getFormattedTimestamp,
+    timestampStyle,
+  }
+}
+
+export const MessageTimestamp = ({
+  timestamp,
+  style,
+  isSentByUser,
+  sentBy,
```
AinaVendrell, line 48: do we need this to prop? can we not get `isSentByUser` ny using the `sentBy` prop

packages/botonic-react/src/components/message/message-image.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { useContext } from 'react'
+
+import { WEBCHAT } from '../../constants'
+import { WebchatContext } from '../../contexts'
+import { resolveImage } from '../../util/environment'
+import { BotMessageImageContainer } from './styles'
+
+interface MessageImageProps {
+  imagestyle: Record<string, unknown>
```
AlbertGom, line 9: camelCase: imageStyle?

packages/botonic-react/src/components/message/message-image.tsx
```diff
@@ -0,0 +1,52 @@
...
+  isSentByUser: boolean
+  isSentByAgent: boolean
```
AlbertGom, line 11: I think it's better to use a variable called `sentBy` that is an enum that can have values 'agent', 'bot' and 'user'?

packages/botonic-react/src/components/message/timestamps.jsx
AlbertGom: Agree with Aina

AlbertGom (approved):

AlbertGom:
@Iru89 Try to add tests for this new behaviour!

AlbertGom:
Looks good :clap:

packages/botonic-react/src/components/message/timestamps.jsx
Iru89: ok I will remove the isSentByUser

packages/botonic-react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/react",
-  "version": "0.24.0-alpha.2",
+  "version": "0.24.0-alpha.4",
```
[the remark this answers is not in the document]
Iru89: The alpha.3 already existed

packages/botonic-react/src/components/message/index.jsx
```diff
@@ -241,20 +195,15 @@ export const Message = props => {
...
-    )
     const animationsEnabled = getThemeProperty(
       WEBCHAT.CUSTOM_PROPERTIES.enableAnimations,
       true
     )
 
     const resolveCustomTypeName = () =>
-      isSentByBot && type === INPUT.CUSTOM ? ` ${m.customTypeName}` : ''
+      isSentByBot && type === INPUT.CUSTOM
+        ? ` ${messageJSON.customTypeName}`
```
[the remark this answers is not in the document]
Iru89: This space was already there before and I think it is necessary to show two classes of css, in the 208 line

packages/botonic-react/src/components/message/message-image.tsx
Iru89: No the correct prop is imagestyle. There is an old [PR#1063](https://github.com/hubtype/botonic/pull/1063) in the description of this PR that says it is better to use imagestyle. 
There is also a code that makes a log saying that legacy is imageStyle and the new imagestyle (src/util/logs.js).

packages/botonic-react/src/components/message/message-image.tsx
Iru89: Yes, I will refactor it to pass only sentBy and do the check inside the component.

PR #2667, opened 2023-11-13
Add css class in RepliesCopntainer

Iru89:
## Description
Add the replies-container css class to the replies container div. 
With this class we can modify in a simpler way the styles of the replies container.


packages/botonic-react/src/webchat/replies.jsx
```diff
@@ -49,11 +49,9 @@ export const WebchatReplies = props => {
       autoHide={scrollbarOptions.autoHide}
     >
       <RepliesContainer
+        className='replies-container'
         justify={justifyContent}
         wrap={flexWrap}
-        style={{
-          ...props.style,
-        }}
```
AinaVendrell, line 56: is this not used by any user?

AinaVendrell (approved):

packages/botonic-react/src/webchat/replies.jsx
Iru89: these props.style are never passed to the WebchatReplies component. You can see it in the other modified file in this PR

PR #2671, opened 2023-11-22
feat(plugin-flow-builder): make optional the flowUrl config param and calculate it based on new FLOWBUILDER_API_URL env var and isStagingIntegration session attribute

gbarba:
## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/index.ts
+++ b/packages/botonic-plugin-flow-builder/src/index.ts
@@ -23,9 +23,16 @@ import { DEFAULT_FUNCTIONS } from './functions'
 import { BotonicPluginFlowBuilderOptions, KnowledgeBaseResponse } from './types'
 import { resolveGetAccessToken } from './utils'
 
+function getFlowUrl(request: PluginPreRequest) {
+  const baseFlowUrl = process.env.FLOWBUILDER_API_URL
+  return request.session.isStagingIntegration
+  ? `${baseFlowUrl}/latest`
+  : `${baseFlowUrl}/draft`
+}
+
 export default class BotonicPluginFlowBuilder implements Plugin {
   public cmsApi: FlowBuilderApi
-  private flowUrl: string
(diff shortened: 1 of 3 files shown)


packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -23,9 +23,16 @@ import { DEFAULT_FUNCTIONS } from './functions'
 import { BotonicPluginFlowBuilderOptions, KnowledgeBaseResponse } from './types'
 import { resolveGetAccessToken } from './utils'
 
+function getFlowUrl(request: PluginPreRequest) {
+  const baseFlowUrl = process.env.FLOWBUILDER_API_URL
+  return request.session.isStagingIntegration
+  ? `${baseFlowUrl}/latest`
+  : `${baseFlowUrl}/draft`
```
asastre, line 30: Shouldn't this be the other way around? `draft` version when `isStagingIntegration` is `true` and `latest` otherwise?

asastre (changes requested):

packages/botonic-plugin-flow-builder/src/index.ts
Iru89: yes

packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -23,9 +23,16 @@ import { DEFAULT_FUNCTIONS } from './functions'
 import { BotonicPluginFlowBuilderOptions, KnowledgeBaseResponse } from './types'
 import { resolveGetAccessToken } from './utils'
 
+function getFlowUrl(request: PluginPreRequest) {
+  const baseFlowUrl = process.env.FLOWBUILDER_API_URL
```
gbarba, line 27: this is `FLOW_BUILDER_API_URL`

PR #2686, opened 2023-12-04
Plugin-flow-builder: remove start_up_node

Iru89:
## Description
- Remove the start_up_node. Always use the start_node_id to retrieve the content used as start node.
- Add a getStartContents function

## Context
Until now we had one node (different from all) for which text content was the start_up_node. With this change the JSON that has all the flow has an attribute where the id of the content that is used as start is indicated (it doesn't have to be a text content).


package.json
```diff
@@ -47,5 +47,6 @@
   "workspaces": [
     "packages/*"
   ],
-  "dependencies": {}
+  "dependencies": {},
+  "version": "0.23.5-alpha.0"
```
AinaVendrell, line 51: why is this version here?

Iru89: This is the version I will release when I have the approve. for testing in the QA environment.

AinaVendrell (approved):

PR #2695, opened 2023-12-14
Fix: className in CustomMessage, avoid render undefined in ScrollButton and BannerUnreadMessages

Iru89:
## Description
This PR fixes two issues introduced in version 0.24.0 of @botonic/react.

1) Custom components did not have the customMessage name as className.

2) The ScrollButton and BannerUnreadMessages components rendered nothing if this feature was not used. 

## Context
1) Custom components did not have the customMessage name as className. This happened because the deserialize function of the Custom component did not pass the new sentBy and isUnread attributes. 
As these attributes were not serialised in the Message component, a check was made to define the className that verified that the CustomMessage was sent by the bot. Prior to version 0.24.0 if from (sentBy now) was not defined then SENDERS.bot was set by default.

2) The ScrollButton and BannerUnreadMessages components rendered nothing if this feature was not used. 
If the feature is enabled by default or using custom components to display these components the error did not occur but if the feature was not enabled the webchat would break. Now when this feature is not active it returns null.

### DIFF
--- a/packages/botonic-react/src/components/custom-message.jsx
+++ b/packages/botonic-react/src/components/custom-message.jsx
@@ -64,6 +64,8 @@ export const customMessage = ({
           children: childrenWithoutReplies,
           customTypeName: name,
         }}
+        sentBy={props.sentBy}
+        isUnread={props.isUnread}
       >
         <ErrorBoundary key={'errorBoundary'} {...customMessageProps}>
           <CustomMessageComponent {...customMessageProps}>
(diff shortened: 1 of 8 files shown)


packages/botonic-react/src/components/custom-message.jsx
```diff
@@ -77,7 +79,14 @@ export const customMessage = ({
   WrappedComponent.customTypeName = name
   // eslint-disable-next-line react/display-name
   WrappedComponent.deserialize = msg => (
-    <WrappedComponent id={msg.id} key={msg.key} json={msg.data} {...msg.data} />
+    <WrappedComponent
+      id={msg.id}
+      key={msg.key}
+      json={msg.data}
```
asastre, line 85: Observation: I know this prop was alredy passed into this component, but taking a look at the `WrappedComponent`, is the `json` prop really needed here?

asastre (approved):

packages/botonic-react/src/components/custom-message.jsx
Iru89: Yes it is strange how this is passed here and it is recreated inside the CustomMessage. In a log you can see how one of the props is json and inside there is the same json with less attributes. 
Captura de pantalla 2023-12-14 a las 15 18 53

For now I don't want to touch it but I will check it in the future in another PR.

AinaVendrell (approved):

PR #2754, opened 2024-01-09
plugin-dynamoDb: implement plugin to comply core types

Iru89:
## Description
 implements Plugin to comply with @botonic/core types and avoid a type error
@botonic/core is as a devDependency because it is only needed to import the Plugin interface

## Context
This error is encountered when you create a bot action extending from FlowBuilderAction that uses `request: ActionRequest` where plugins are declared as `[id:string]: Plugin`


AlbertGom (approved):

AinaVendrell (approved):

Iru89:
I close this PR because this is applied in the node-v20 branch

PR #2769, opened 2024-01-19
plugin-flow-builder: handle bot action on the flow builder plugin #BLT-483

Iru89:
## Description

- adds the functionality to use BotActionNode created in the FlowBuilder frontend.
- add a function to getPayloadParams in JSON format when receiving a payload.
- add a base type for payloadParams 

## Context

A BotAction node can target 4 types of nodes: Button, Handoff, Keyword and Intent

When the target is a Button (from Text, Carousel or WhatsappButtonList) or a Handoff in the botonicInit function of the plugin-flow-builder we must obtain the payload defined in the BotAction node and replace it. This way when the user clicks on a button or executes a handoff the payload will be the correct one.

With Keywords and Intents it is different, what we want is that when the bot executes the routes, the payload is already in the request so that the route of the action indicated by the payload is executed. To do this we have to make the logic of building the payload in the pre function of the plugin-flow-builder. This function is executed before the routes.
**IMPORTANT -> For the case where an intent points directly to a nod BotAction to work, the plugin-hubtype-babel must be placed before the plugin-flow-builder.**

Example of how a bot action would be executed by receiving a payload defined with a BotActionNode.

Here we use the function getPaylodParams, one of the parameters will be followUpId if the BotAcitonNode has a follow up defined. With this Id we retrieve the contents for rendering.

```typescript
import {
  FlowBuilderActionProps,
  FlowBuilderMultichannelAction,
} from '@botonic/plugin-flow-builder'
import { PayloadParamsBase } from '@botonic/plugin-flow-builder/lib/esm'
import { RequestContext } from '@botonic/react'
import React from 'react'

import { BotRequest } from '../types'
import { getRequestData } from '../utils/actions'

export class TestAction extends FlowBuilderMultichannelAction {
  static async botonicInit(
    request: BotRequest
    const { cmsPlugin, botContext, payload } = getRequestData(request)
    const payloadParams = cmsPlugin.getPayloadParams<PayloadParamsBase>(payload)

    console.log('Do some logic with payloadParams', { payloadParams })

    const contents = await cmsPlugin.getContentsById(
      payloadParams.followUpId,
      botContext.locale
    )
    return { contents }
  }
}
```

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/api.ts
+++ b/packages/botonic-plugin-flow-builder/src/api.ts
@@ -1,7 +1,9 @@
 import { Input, PluginPreRequest } from '@botonic/core'
 import axios from 'axios'
 
+import { SEPARATOR } from './constants'
 import {
+  HtBotActionNode,
   HtFallbackNode,
   HtFlowBuilderData,
   HtIntentNode,
@@ -11,6 +13,7 @@ import {
   HtNodeWithContent,
   HtNodeWithContentType,
   HtNodeWithoutContentType,
+  HtPayloadNode,
 } from './content-fields/hubtype-fields'
 import { FlowBuilderApiOptions } from './types'
 
@@ -175,5 +178,28 @@ export class FlowBuilderApi {
 
   private containsAnyKeywords(input: string, keywords: string[]): boolean {
     return keywords.some(keyword => input.includes(keyword))
+  }
+
+  getPayloadFromBotActionNodeId(id?: string): string | undefined {
+    if (id) {
+      const botActionNode = this.getNodeById(id)
+      if (botActionNode.type === HtNodeWithoutContentType.BOT_ACTION) {
+        return this.createPayloadWithParams(botActionNode)
+      }
+    }
+    return undefined
+  }
+
+  private createPayloadWithParams(botActionNode: HtBotActionNode): string {
+    const payloadId = botActionNode.content.payload_id
+    const payloadNode = this.getNodeById<HtPayloadNode>(payloadId)
+    const customParams = JSON.parse(
+      botActionNode.content.payload_params || '{}'
+    )
+    const payloadJson = JSON.stringify({
+      ...customParams,
+      followUpId: botActionNode.follow_up?.id,
+    })
+    return `${payloadNode.content.payload}${SEPARATOR}${payloadJson}`
   }
 }
--- a/packages/botonic-plugin-flow-builder/src/api.ts
+++ b/packages/botonic-plugin-flow-builder/src/api.ts
@@ -145,18 +145,16 @@ export class FlowBuilderApi {
   getNodeByKeyword(
     userInput: string,
     locale: string
-  ): HtNodeWithContent | undefined {
+  ): HtKeywordNode | undefined {
     try {
       const keywordNodes = this.flow.nodes.filter(
-        node => node.type == HtNodeWithContentType.KEYWORD
+        node => node.type === HtNodeWithContentType.KEYWORD
       ) as HtKeywordNode[]
       const matchedKeywordNodes = keywordNodes.filter(node =>
         this.matchKeywords(node, userInput, locale)
       )
       if (matchedKeywordNodes.length > 0 && matchedKeywordNodes[0].target) {
-        return this.getNodeById<HtNodeWithContent>(
-          matchedKeywordNodes[0].target.id
-        )
+        return matchedKeywordNodes[0]
       }
     } catch (error) {
       console.error(`Error getting node by keyword '${userInput}': `, error)
@@ -180,12 +178,16 @@ export class FlowBuilderApi {
     return keywords.some(keyword => input.includes(keyword))
   }
 
-  getPayloadFromBotActionNodeId(id?: string): string | undefined {
-    if (id) {
-      const botActionNode = this.getNodeById(id)
-      if (botActionNode.type === HtNodeWithoutContentType.BOT_ACTION) {
+  getPayload(target?: HtNodeLink): string | undefined {
+    if (target) {
+      console.log('getPayload', { target })
+      if (target.type === HtNodeWithoutContentType.BOT_ACTION) {
+        console.log('BOT_ACTION')
+        const botActionNode = this.getNodeById<HtBotActionNode>(target.id)
         return this.createPayloadWithParams(botActionNode)
       }
+      const targetNode = this.getNodeById(target.id)
+      return targetNode.id
     }
     return undefined
   }
--- a/packages/botonic-plugin-flow-builder/src/api.ts
+++ b/packages/botonic-plugin-flow-builder/src/api.ts
@@ -179,15 +179,16 @@ export class FlowBuilderApi {
   }
 
   getPayload(target?: HtNodeLink): string | undefined {
-    if (target) {
-      if (target.type === HtNodeWithoutContentType.BOT_ACTION) {
-        const botActionNode = this.getNodeById<HtBotActionNode>(target.id)
-        return this.createPayloadWithParams(botActionNode)
-      }
-      const targetNode = this.getNodeById(target.id)
-      return targetNode.id
+    if (!target) {
+      return undefined
     }
-    return undefined
+
+    if (target.type === HtNodeWithoutContentType.BOT_ACTION) {
+      const botActionNode = this.getNodeById<HtBotActionNode>(target.id)
+      return this.createPayloadWithParams(botActionNode)
+    }
+
+    return target.id
   }
 
   private createPayloadWithParams(botActionNode: HtBotActionNode): string {
--- a/packages/botonic-plugin-flow-builder/src/api.ts
+++ b/packages/botonic-plugin-flow-builder/src/api.ts
@@ -180,9 +180,7 @@ export class FlowBuilderApi {
 
   getPayload(target?: HtNodeLink): string | undefined {
     if (target) {
-      console.log('getPayload', { target })
       if (target.type === HtNodeWithoutContentType.BOT_ACTION) {
-        console.log('BOT_ACTION')
         const botActionNode = this.getNodeById<HtBotActionNode>(target.id)
         return this.createPayloadWithParams(botActionNode)
       }
--- a/packages/botonic-plugin-flow-builder/src/content-fields/flow-button.tsx
+++ b/packages/botonic-plugin-flow-builder/src/content-fields/flow-button.tsx
@@ -27,12 +27,9 @@ export class FlowButton extends ContentFieldsBase {
     const newButton = new FlowButton(cmsButton.id)
     newButton.text = this.getTextByLocale(locale, cmsButton.text)
     if (cmsButton.target) {
-      const payloadFromBotActionNode = cmsApi.getPayloadFromBotActionNodeId(
-        cmsButton.target.id
-      )
-      const payload = payloadFromBotActionNode || cmsButton.target.id
-      newButton.payload = payload
+      newButton.payload = cmsApi.getPayload(cmsButton.target)
     }
+
     // OLD PAYLOAD
     if (cmsButton.payload && payloadId) {
       const payloadNode = cmsApi.getNodeById<HtPayloadNode>(payloadId)
--- a/packages/botonic-plugin-flow-builder/src/content-fields/flow-button.tsx
+++ b/packages/botonic-plugin-flow-builder/src/content-fields/flow-button.tsx
@@ -27,14 +27,19 @@ export class FlowButton extends ContentFieldsBase {
     const newButton = new FlowButton(cmsButton.id)
     newButton.text = this.getTextByLocale(locale, cmsButton.text)
     if (cmsButton.target) {
-      newButton.payload = cmsButton.target.id
+      const payloadFromBotActionNode = cmsApi.getPayloadFromBotActionNodeId(
+        cmsButton.target.id
+      )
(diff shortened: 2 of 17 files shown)


packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -165,8 +186,13 @@ export default class BotonicPluginFlowBuilder implements Plugin {
     }
     return result.target.id
   }
+
+  getPayloadParams<T>(payload: string): T & PayloadParamsBase {
+    const payloadParams = JSON.parse(payload.split(SEPARATOR)[1] || '{}')
+    return payloadParams
+  }
```
asastre, line 193: Doubt: Is this method used anywhere? Why you can assure that the response will implement the `PayloadParamsBase` interface?

asastre (commented):
@Iru89 I have some doubts that I would like to discuss before approving the PR

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -176,4 +179,27 @@ export class FlowBuilderApi {
...
+    }
+    return undefined
+  }
+
+  private createPayloadWithParams(botActionNode: HtBotActionNode): string {
+    const payloadId = botActionNode.content.payload_id
+    const payloadNode = this.getNodeById<HtPayloadNode>(payloadId)
+    const customParams = JSON.parse(
+      botActionNode.content.payload_params || '{}'
+    )
```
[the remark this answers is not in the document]
AinaVendrell: in the flow builder we display an error if the params are not a valid JSON, but if the user doesn't fix the error, it can fail here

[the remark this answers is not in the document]
Iru89: As Aina says, the flow builder front-end does not allow saving or publishing if the format is not correct

packages/botonic-plugin-flow-builder/src/content-fields/flow-button.tsx
```diff
@@ -27,14 +27,19 @@ export class FlowButton extends ContentFieldsBase {
     const newButton = new FlowButton(cmsButton.id)
     newButton.text = this.getTextByLocale(locale, cmsButton.text)
     if (cmsButton.target) {
-      newButton.payload = cmsButton.target.id
+      const payloadFromBotActionNode = cmsApi.getPayloadFromBotActionNodeId(
+        cmsButton.target.id
+      )
+      const payload = payloadFromBotActionNode || cmsButton.target.id
```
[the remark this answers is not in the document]
Iru89: That's just what line 33 does.  The function getPayloadFromBotActionNodeId returns the payload defined in a BotAction if the id that you pass is a node is of type bot-action, if it is not of type undefined it returns undefined so that cmsButton.target.id will be executed.

packages/botonic-plugin-flow-builder/src/content-fields/hubtype-fields/bot-action.ts
```diff
@@ -0,0 +1,11 @@
+import { HtBaseNode } from './common'
+import { HtNodeWithoutContentType } from './node-types'
+
+export interface HtBotActionNode extends HtBaseNode {
```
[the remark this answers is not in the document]
Iru89: I don't know if I understand you correctly. 
But HtNodeLink.target.type is HtNodeWithContentType or HtNodeWithoutContentType.  
BotAction is defined in HtNodeWithoutContentType

`
export interface HtNodeLink {
  id: string
  type: HtNodeWithContentType | HtNodeWithoutContentType
}

export enum HtNodeWithoutContentType {
  URL = 'url',
  PAYLOAD = 'payload',
  GO_TO_FLOW = 'go-to-flow',
  BOT_ACTION = 'bot-action',
}`

packages/botonic-plugin-flow-builder/src/index.ts
Iru89: I would like to discuss this with you. This method is for use in a bot action that runs when you receive a payload defined in a BotActionNode. I've been thinking about it and I think it needs to be typed like this

``typescript
  getPayloadParams<T extends PayloadParamsBase>(payload: string): T {
    const payloadParams = JSON.parse(payload.split(SEPARATOR)[1] || '{}')
    return payloadParams
  }
´´´

packages/botonic-plugin-flow-builder/src/api.ts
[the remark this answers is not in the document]
Iru89: What I said above is not entirely correct. You can save but not publish

asastre (approved):

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -176,4 +179,27 @@ export class FlowBuilderApi {
...
+
+  getPayloadFromBotActionNodeId(id?: string): string | undefined {
+    if (id) {
+      const botActionNode = this.getNodeById(id)
+      if (botActionNode.type === HtNodeWithoutContentType.BOT_ACTION) {
+        return this.createPayloadWithParams(botActionNode)
+      }
+    }
+    return undefined
+  }
```
AlbertGom, line 191: Why are you defining the `id` param as optional and then returning undefined if it is not defined?

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -176,4 +179,27 @@ export class FlowBuilderApi {
   private containsAnyKeywords(input: string, keywords: string[]): boolean {
     return keywords.some(keyword => input.includes(keyword))
   }
+
+  getPayloadFromBotActionNodeId(id?: string): string | undefined {
```
AlbertGom, line 183: Why are you creating this method? The `getNodeById` method should do what this method does, no?
Then when having the node, you should access `node.payload` to get the payload.

packages/botonic-plugin-flow-builder/src/content-fields/flow-button.tsx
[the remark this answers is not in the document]
AlbertGom: I agree with Adria

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -176,4 +179,27 @@ export class FlowBuilderApi {
...
+    if (id) {
+      const botActionNode = this.getNodeById(id)
```
AlbertGom, line 185: I think the name of this variable does not represents what it is as this node could be of another type, not a bot action one...

packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -59,6 +64,22 @@ export default class BotonicPluginFlowBuilder implements Plugin {
       accessToken: this.getAccessToken(request.session),
       request: this.currentRequest,
     })
+
+    if (
+      request.input.data &&
+      !request.input.payload &&
+      !request.session.is_first_interaction
```
AlbertGom, line 71: nit: I would encapsulate this into a function with a descriptive name to make the code easier to read.

packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -165,8 +186,13 @@ export default class BotonicPluginFlowBuilder implements Plugin {
     }
     return result.target.id
   }
+
+  getPayloadParams<T>(payload: string): T & PayloadParamsBase {
```
AlbertGom, line 190: I think this function should be placed in another file like utils or something like that.

Iru89: I need to export this function from the plugin's index to be able to do the import in a simple way as

`import { PayloadParamsBase } from '@botonic/plugin-flow-builder'`

packages/botonic-plugin-flow-builder/src/api.ts
Iru89: I have refactored this as we have discussed with Adria it will now be a function called getPayload and will return the payload. It will create the payload from a BotAction node or return the id of the text, image, video, carousel etc that is connected to a button, handoff, intent or keyword.

packages/botonic-plugin-flow-builder/src/api.ts
Iru89: I have refactored and changed a lot this function but what you say I would say that it is not like that if you want tomorrow we will talk and I will explain you how it is done.

packages/botonic-plugin-flow-builder/src/index.ts
Iru89: I want it inside the plugin for you to use by doing `cmsPlugin.getPaylodParams<PayloadParamsBase>(request.input.payload)` as seen in the description of this PR. It can be another type extending from PayloadParamsBase.

packages/botonic-plugin-flow-builder/src/content-fields/flow-button.tsx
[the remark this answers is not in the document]
Iru89: refactored

packages/botonic-plugin-flow-builder/src/index.ts
Iru89: applied

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -176,4 +177,29 @@ export class FlowBuilderApi {
...
+    if (target) {
+      if (target.type === HtNodeWithoutContentType.BOT_ACTION) {
+        const botActionNode = this.getNodeById<HtBotActionNode>(target.id)
+        return this.createPayloadWithParams(botActionNode)
+      }
+      const targetNode = this.getNodeById(target.id)
+      return targetNode.id
+    }
+    return undefined
+  }
```
AlbertGom, line 191: As Adrià said I think the best approach would be to implement an abstract class called Target and create two subclasses (BotActionTarget and ContentTarget) which each of this has a method that implements an asbtract method called getPayload. If you want to keep this function I think some improvements can be done:

- Use early returns: `if(!target){ return undefined }`
- You can avoid one getNodeById call: Line 187 is not necessary since you have the target received by param and you can access target.id.

packages/botonic-plugin-flow-builder/src/api.ts
AlbertGom: Yes I was confused, sorry for the noise

packages/botonic-plugin-flow-builder/src/api.ts
AlbertGom: Yes, I was confused, sorry for the noise

packages/botonic-plugin-flow-builder/src/index.ts
AlbertGom: To me, makes sense that the methods declared in a class do operations with elements of the class. In my opinion and following the SRP this method is independent from the class and should be placed into a utils file and be exported from there. But this is an opinion, if you think it makes more sense having it there, keep it.

packages/botonic-plugin-flow-builder/src/api.ts
AlbertGom: What I was trying to explain here is that you were getting the node and then after that there was an `if` checking if the node is of type BotAction, so it meant that it couod be the possibility that it was not of type BotAction.

AlbertGom:
Btw, good job! This will improve a lot the bots development 👏🫶

packages/botonic-plugin-flow-builder/src/api.ts
asastre: I agree with the early return, this way you will remove the long `if(target)` that encapsulates the whole code inside the function.

I also agree with the other point, you could return directly `target.id` and you could avoid the `getNodeById` call.

Regarding the classes, if you do it, do it in a later PR, this one has plenty of stuff already 😉

packages/botonic-plugin-flow-builder/src/content-fields/flow-button.tsx
```diff
@@ -27,14 +27,16 @@ export class FlowButton extends ContentFieldsBase {
     const newButton = new FlowButton(cmsButton.id)
     newButton.text = this.getTextByLocale(locale, cmsButton.text)
     if (cmsButton.target) {
-      newButton.payload = cmsButton.target.id
+      newButton.payload = cmsApi.getPayload(cmsButton.target)
```
asastre, line 30: Observation: Comming back to the suggestion of creating a class for the target objects, this could be called this way:
```suggestion
      newButton.payload = cmsButton.target.getPayload()
```
or even this way if you create a `get` method.
```suggestion
      newButton.payload = cmsButton.target.payload
```

packages/botonic-plugin-flow-builder/src/api.ts
Iru89: Yeah I don't know why I changed this, I think before I applied this refactor I was debugging the target.id 🤷🏻‍♂️
I will apply the early return and avoid the getNodeById by returning the target.id directly

Iru89: > As Adrià said I think the best approach would be to implement an abstract class with an abstract method getPayload and create two classes (BotActionTarget and ContentTarget) which implement the first one. If you want to keep this function I think some improvements can be done:
> 
> * Use early returns: `if(!target){ return undefined }`
> * You can avoid one getNodeById call: Line 187 is not necessary since you have the target received by param and you can access target.id.

I will keep it in mind for a future refactor to create a Target class that has the logic to create the payloads but right now I won't apply it because it would imply a lot of changes.

PR #2773, opened 2024-01-22
plugin-flow-builder: fix trackEvent when intent does not have enough confidence

Iru89:
## Description
Call only the track event indicating that an intent has failed when the intent does not have sufficient confidence

## Context

This function was being called whenever the bot received an input from the user and if this input did not match any created intent the server response is 500 because it expects the intent and the confidence in the POST call object.

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/action/intent.ts
+++ b/packages/botonic-plugin-flow-builder/src/action/intent.ts
@@ -1,17 +1,14 @@
 import { ActionRequest } from '@botonic/react'
 
 import { FlowBuilderApi } from '../api'
-import {
-  HtIntentNode,
-  HtNodeWithContent,
-} from '../content-fields/hubtype-fields'
+import { HtIntentNode } from '../content-fields/hubtype-fields'
 import { EventName, trackEvent } from './tracking'
 
-export async function getNodeByIntent(
+export async function getIntentNodeByInput(
   cmsApi: FlowBuilderApi,
   locale: string,
   request: ActionRequest
-): Promise<HtNodeWithContent | undefined> {
+): Promise<HtIntentNode | undefined> {
   const intentNode = cmsApi.getIntentNode(request.input, locale)
   const eventArgs = {
     intent: request.input.intent as string,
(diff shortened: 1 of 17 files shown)


packages/botonic-plugin-flow-builder/src/action/intent.ts
```diff
@@ -15,14 +15,17 @@ export async function getIntentNodeByInput(
     confidence: request.input.confidence as number,
     confidence_successful: true,
   }
-  if (isIntentValid(intentNode, request, cmsApi) && intentNode?.target?.id) {
-    await trackEvent(request, EventName.botAiModel, eventArgs)
-    return intentNode
-  } else {
+  if (request.input.confidence && request.input.intent) {
+    if (isIntentValid(intentNode, request, cmsApi) && intentNode?.target?.id) {
```
AlbertGom, line 19: You can add this `&& intentNode?.target?.id` inside `isIntentValid` method

AlbertGom (approved):

packages/botonic-plugin-flow-builder/src/action/intent.ts
Iru89: This intentNode.target.id was here because of the logic I changed in the previous PR, it is not needed now. What would be logical is that if there is no IntentNode for the received request.input.intent then don't create event for botAiModel.

```
if (request.input.confidence && request.input.intent && intentNode) {
    if (isIntentValid(intentNode, request, cmsApi)) {
```

packages/botonic-plugin-flow-builder/src/action/intent.ts
```diff
@@ -15,23 +15,25 @@ export async function getIntentNodeByInput(
...
 
 function isIntentValid(
-  intentNode: HtIntentNode | undefined,
+  intentNode: HtIntentNode,
   request: ActionRequest,
   cmsApi: FlowBuilderApi
 ) {
   return (
-    intentNode &&
     request.input.confidence &&
```
AlbertGom, line 37: You can also remove this line

AlbertGom (approved):

AinaVendrell (approved):

PR #2782, opened 2024-02-28
plugin-flow-builder: check if agents are available in conditional queue 

Iru89:
## Description

Now in the FlowBuilder frontend we can have queue conditionals, which check if there are available agents with a check box. This generates a conditional queue with 3 outputs.

## Context

The node with 3 outputs will look like this
```typescript
{
      id: 'b0b61028-b683-4ea6-ba7d-b79ca160e438',
      code: 'CONDITIONAL_QUEUE',
      meta: { x: 358.20617669934023, y: 352.83848058278346 },
      follow_up: null,
      target: null,
      flow_id: '8d527e7d-ea6d-5422-b810-5b4c8be7657b',
      type: 'function',
      content: {
        action: 'check-queue-status',
        arguments: [
          {
            type: 'boolean',
            name: 'check_available_agents',
            value: true,
          },
          {
            locale: 'en',
            values: [
              {
                type: 'string',
                name: 'queue_id',
                value: 'e7a2304d-f73c-409d-b272-239a9b8a9e0e',
              },
              { type: 'string', name: 'queue_name', value: 'General' },
            ],
          },
        ],
        result_mapping: [
          {
            result: 'open',
            target: {
              id: 'de7a4af4-0c50-44ee-9f6c-86796c6d4c53',
              type: 'text',
            },
          },
          {
            result: 'open-without-agents',
            target: {
              id: '503cfe1a-8cd4-4c9c-be42-dc00b3d43806',
              type: 'text',
            },
          },
          {
            result: 'closed',
            target: {
              id: 'd2ec4c3a-bdc2-4e22-8b1c-ec0b473feadd',
              type: 'text',
            },
          },
        ],
      },
    },
```

The arguments of a node of type function can have arguments that depend on the locale and arguments that do not depend on the locale (for example argument with name check_available_agents)

After processing the arguments we get the following arrays:

```typescript 
functionArguments: [
    { type: 'boolean', name: 'check_available_agents', value: true },
    {
      type: 'string',
      name: 'queue_id',
      value: 'e7a2304d-f73c-409d-b272-239a9b8a9e0e'
    },
    { type: 'string', name: 'queue_name', value: 'General' }
  ],
  nameValues: [
    { check_available_agents: true },
    { queue_id: 'e7a2304d-f73c-409d-b272-239a9b8a9e0e' },
    { queue_name: 'General' }
  ]
```

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/functions/conditional-queue-status.ts
+++ b/packages/botonic-plugin-flow-builder/src/functions/conditional-queue-status.ts
@@ -10,15 +10,21 @@ type ConditionalQueueStatusArgs = {
   check_available_agents: boolean
 }
 
+enum QueueStatusResult {
+  open = 'open',
+  closed = 'closed',
+  openWithoutAgents = 'open-without-agents',
+}
+
 export async function conditionalQueueStatus({
   queue_id,
   check_available_agents,
-}: ConditionalQueueStatusArgs): Promise<string> {
+}: ConditionalQueueStatusArgs): Promise<QueueStatusResult> {
   const data = await getQueueAvailability(queue_id, check_available_agents)
   if (check_available_agents && data.open && data.available_agents === 0) {
-    return 'open_without_agents'
+    return QueueStatusResult.openWithoutAgents
   }
-  return data.open ? 'open' : 'closed'
+  return data.open ? QueueStatusResult.open : QueueStatusResult.closed
 }
 
 interface AvailabilityData {
--- a/packages/botonic-plugin-flow-builder/src/functions/conditional-queue-status.ts
+++ b/packages/botonic-plugin-flow-builder/src/functions/conditional-queue-status.ts
@@ -11,9 +11,9 @@ type ConditionalQueueStatusArgs = {
 }
 
 enum QueueStatusResult {
-  open = 'open',
-  closed = 'closed',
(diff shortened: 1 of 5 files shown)


packages/botonic-plugin-flow-builder/src/functions/conditional-queue-status.ts
```diff
@@ -7,14 +7,24 @@ type ConditionalQueueStatusArgs = {
   request: ActionRequest
   queue_id: string
   queue_name: string
+  check_available_agents: boolean
+}
+
+enum QueueStatusResult {
+  open = 'open',
```
AinaVendrell, line 14: don't we use upper case normally for enum key?
```
enum QueueStatusResult {
  OPEN = 'open',
  ...
}
```

AinaVendrell (approved):

packages/botonic-plugin-flow-builder/src/functions/conditional-queue-status.ts
Iru89: yes you are right

Iru89: I will apply it to follow the same as we have throughout the code, even though the lint rule indicates a warn

Captura de pantalla 2024-03-04 a las 12 01 58

PR #2790, opened 2024-03-11
plugin-flow-builder: use only content id outside plugin

Iru89:
## Description

This PR for the moment is just a Draft to see how we could change the functions to get the contents without using the UUID (in the code it is called id) and use only the ContentID (in the code it is called code).

The idea is to have ContentID instead of UUID hardcoded in the bot code, this way if the client deletes and recreates a node in the frontend of the flow builder but puts the same ContentID the bot will still work the same.

## Context

Use the `getContentsByContentID` function if you need to get content in an action written in the bot code, instead of using the `getContentsById`.  The function `getContentsByContentID` can be useful if you want to render a content in a custom message in webchat but not in other channels like WhatsApp. For example in a rating action in webchat you want a custom message to select the stars but in Whatsapp you want it to be a normal message with buttons.

Add the function `getUUIDByContentID` that will allow to get the UUID of a node using the ContentID. This is useful in case you need to pass the UUID as a payload to the botonicInit function of the plugin-flow-builder to do handoff, tracking, etc.

Make the `getContentsById` function private so that it cannot be called from outside the plugin.

You can no longer use the function getContentsById in the bot code, this function becomes private, instead of this function you should use getContentsByContentID. 

If the BotAction node has a followUp instead of passing its followUpId (UUID) as a payload parameter, the followUpContentID will be added and using the getContentsByContentID function you can get the contents to render.

Example using a BotAction with followUp:  
```typescript
import {
  FlowBuilderActionProps,
  FlowBuilderMultichannelAction,
} from '@botonic/plugin-flow-builder'
import { RequestContext } from '@botonic/react'
import React from 'react'

import { BotRequest } from '../types'
import { getRequestData } from '../utils/actions'

interface RatingParms {
  value: number
  followUpContentID: string
}

export class TestRatingAction extends FlowBuilderMultichannelAction {
  static async botonicInit(
    request: BotRequest
    const { cmsPlugin, botContext, payload } = getRequestData(request)
    const payloadParams = cmsPlugin.getPayloadParams<RatingParms>(payload)

    console.log('Do some logic with payloadParams', { payloadParams })

    const contents = await cmsPlugin.getContentsByContentID(
          payloadParams.followUpContentID,
          botContext.locale
        )

    return { contents }
  }
}

```


packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -90,16 +90,21 @@ export default class BotonicPluginFlowBuilder implements Plugin {
...
     prevContents?: FlowContent[]
   ): Promise<FlowContent[]> {
-    const node = this.cmsApi.getNodeByCode(code) as HtNodeWithContent
+    const node = this.cmsApi.getNodeByContentID(contentID) as HtNodeWithContent
     return await this.getContentsByNode(node, locale, prevContents)
   }
 
-  async getContentsById(
+  async getUUIDByContentID(contentID: string): Promise<string> {
+    const node = this.cmsApi.getNodeByContentID(contentID)
```
Iru89, line 103: Is it necessary to use an await?

vanbasten17 (approved):

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -56,11 +57,11 @@ export class FlowBuilderApi {
     return node as T
   }
 
-  getNodeByCode(code: string): HtNodeComponent {
+  getNodeByContentID(contentID: string): HtNodeComponent {
```
AinaVendrell, line 60: why using uppercase `D` instead of `Id` as we do with other variables?

AinaVendrell (approved):

packages/botonic-plugin-flow-builder/src/api.ts
Iru89: I have used ContentID because that is how it is written in the flow builder frontend

PR #2816, opened 2024-05-03
Botonic-React: add url icon theme property

AlbertGom:
## Description
Added a property (`urlIcon`) to the webchat theme to be able to show or not an url icon to url buttons. If the property is enabled, a default icon will be shown but there is the possibility to set a custom icon.

Url button with custom icon set:

Captura de pantalla de 2024-05-03 15-15-40

Url button with default icon:

Captura de pantalla de 2024-05-03 16-26-52

## Context
With this change the user will have a better experience since it will easily identify which buttons take him to an external web page.

The urlIcon property can have a enable prop and a image prop.

- If the enable prop is set to false, it will not show the icon (even if an image prop is set)
- If the enable prop is set to true and no image set, then the default icon will be shown.
- If the enable prop is set to true and image set, then the custom icon will be shown.
- If no enable prop, but image set, then the icon will be shown.
- If no enable nor image prop set, then no icon will be shown.

With default icon:

```
button: {
      urlIcon: { enable: true },
      messageType: 'text',
    },
```

With custom icon:
```

 button: {
      urlIcon: { image: UrlIcon },
      messageType: 'text',
    },
```

Without url icon:
```

  button: {
      messageType: 'text',
    },
```


elozano98 (approved):

packages/botonic-react/src/components/button.jsx
```diff
@@ -137,6 +138,22 @@ export const Button = props => {
...
+    const urlIconImage = getThemeProperty(
+      WEBCHAT.CUSTOM_PROPERTIES.urlIconImage
+    )
+
+    const urlIcon = urlIconEnabled
+      ? urlIconImage ?? WEBCHAT.DEFAULTS.URL_ICON
+      : propertyDisabled(urlIconEnabled)
+        ? undefined
+        : urlIconImage
+
```
Iru89, line 156: You are not using the urlIconProp anywhere. I think this could be simpler, the getThemeProperty function accepts a second parameter which is used in case the first one is undefined.
```suggestion
    const urlIconEnabled = getThemeProperty(
      WEBCHAT.CUSTOM_PROPERTIES.urlIconEnabled,
      false
    )

    const urlCustomIconImage = getThemeProperty(
      WEBCHAT.CUSTOM_PROPERTIES.urlIconImage,
      WEBCHAT.DEFAULTS.URL_ICON
    )

    const urlIcon =
      urlIconEnabled || urlCustomIconImage ? urlCustomIconImage : undefined
```

packages/botonic-react/src/components/button.jsx
```diff
@@ -149,14 +166,26 @@ export const Button = props => {
...
+          />
+        )}
       </StyledButton>
     )
   }
 
+  const propertyDisabled = value => {
+    return value === false
+  }
+
```
Iru89, line 188: remove this function

packages/botonic-react/src/components/button.jsx
```diff
@@ -149,14 +166,26 @@ export const Button = props => {
...
           ...(props.disabled && autoDisable && disabledStyle),
+          position: 'relative',
         }}
         bottom={props.bottomRadius}
       >
         {props.children}
+        {props.url && urlIcon && (
+          <img
+            className='botonic-url-icon'
+            style={{ width: '20px', position: 'absolute', right: '12px' }}
```
Iru89, line 177: If you set the prop position absolute, the image is above the button text

Iru89 (changes requested):

packages/botonic-react/src/components/button.jsx
Iru89: I wouldn't use inline styles, these styles can't be easily overwritten and are the ones we end up putting an !important. 
Better to use css or styles in a styled component

Iru89 (changes requested):

packages/botonic-react/src/components/button.jsx
AlbertGom: With your solution, the condition will always be true and the url image will always be shown. That's why I'm not using the default value in getThemeProperty function (the second parameter).

packages/botonic-react/src/components/button.jsx
AlbertGom: I'm using this function to check if the property is set to false and even if there is an image, not showing the image (but maybe this can be simplified an just let the image be shown if there is image and enable rop is set to false.)

Iru89: I think it's messed up, you're using this function to check for false, and when it returns true in the ternary where it's used you return undefined...

packages/botonic-react/src/components/button.jsx
Iru89: You are right the condition always returns true and it should not be like that. Keep in mind that maybe you should have a const with the urlCustomIcon and another with the urlDefaultIcon.
Something like the following
```
const urlIconEnabled = getThemeProperty(
      WEBCHAT.CUSTOM_PROPERTIES.urlIconEnabled,
      false
    )
    const urlCustomIconImage = getThemeProperty(
      WEBCHAT.CUSTOM_PROPERTIES.urlIconImage
    )
    const urlDefaulIconImage = WEBCHAT.DEFAULTS.URL_ICON
    const Icon = urlCustomIconImage ? urlCustomIconImage : urlDefaulIconImage
    const urlIcon =
      urlIconEnabled || urlCustomIconImage ? Icon : undefined
```

Iru89 (approved):

PR #2817, opened 2024-05-06
plugin-flow-builder: bot action limit characters

Iru89:
## Description

When using a Bot Action node with parameters that is connected to a follow up. It is easy to exceed the character limit allowed by Telegram (64bytes = 64 characters).

## Context

When connecting a BotActionNode to a button the payload will be ba|BotActionNodeUUID|source_x

In the pre function of the plugin: first the |source_x is removed, then if a payload starting with ba| is detected the BotActionNode is obtained using the botActionNodeUUID and the payload is replaced by the payload plus the parameters defined in the BotActionNode.

When we get to the bot routes we have the payload with the parameters.

In the action we can continue using the getPayloadParams function as we did before.

## Testing

Add config for jest and github action
Add test for first interaction 
Add test for bot action node

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/index.ts
+++ b/packages/botonic-plugin-flow-builder/src/index.ts
@@ -3,7 +3,7 @@ import { ActionRequest } from '@botonic/react'
 
 import { FlowBuilderApi } from './api'
 import {
-  BOT_ACTION_PAYLOAD_SEPARATOR,
+  BOT_ACTION_PAYLOAD_PREFIX,
   FLOW_BUILDER_API_URL_PROD,
   SEPARATOR,
   SOURCE_INFO_SEPARATOR,
@@ -101,22 +101,23 @@ export default class BotonicPluginFlowBuilder implements Plugin {
       request.input.payload = this.removeSourceSeparatorFromPayload(
         request.input.payload
       )
-      request.input.payload = this.updateBotActionPayload(request.input.payload)
+
+      if (request.input.payload.startsWith(BOT_ACTION_PAYLOAD_PREFIX)) {
+        request.input.payload = this.updateBotActionPayload(
+          request.input.payload
+        )
+      }
     }
   }
 
   private removeSourceSeparatorFromPayload(payload: string): string {
-    return payload?.split(SOURCE_INFO_SEPARATOR)[0]
+    return payload.split(SOURCE_INFO_SEPARATOR)[0]
   }
 
   private updateBotActionPayload(payload: string): string {
-    if (payload.startsWith(BOT_ACTION_PAYLOAD_SEPARATOR)) {
-      const botActionId = payload.split(SEPARATOR)[1]
-      const botActionNode =
-        this.cmsApi.getNodeById<HtBotActionNode>(botActionId)
-      return this.cmsApi.createPayloadWithParams(botActionNode)
-    }
-    return payload
+    const botActionId = payload.split(SEPARATOR)[1]
+    const botActionNode = this.cmsApi.getNodeById<HtBotActionNode>(botActionId)
+    return this.cmsApi.createPayloadWithParams(botActionNode)
   }
 
   async getContentsByContentID(
--- a/packages/botonic-plugin-flow-builder/src/index.ts
+++ b/packages/botonic-plugin-flow-builder/src/index.ts
@@ -3,6 +3,7 @@ import { ActionRequest } from '@botonic/react'
 
 import { FlowBuilderApi } from './api'
 import {
+  BOT_ACTION_PAYLOAD_SEPARATOR,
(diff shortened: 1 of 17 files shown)


vanbasten17 (approved):

packages/botonic-plugin-flow-builder/tests/helpers/flows.ts
```diff
@@ -0,0 +1,307 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+export const testFlow = {
+  version: 'draft',
+  name: 'Test data',
+  comments: null,
```
AinaVendrell, line 5: would be nice to make this flow multi

packages/botonic-plugin-flow-builder/tests/helpers/flows.ts
```diff
@@ -0,0 +1,307 @@
...
+  published_by: null,
+  published_on: null,
+  locales: ['es'],
```
AinaVendrell, line 8: would be nice to make this flow multi locale in the future

packages/botonic-plugin-flow-builder/tests/bot-action.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    const request = createRequest({
+      input: {
+        type: INPUT.POSTBACK,
+        payload: `${BOT_ACTION_PAYLOAD_SEPARATOR}${botActionUuid}`,
+      },
+      plugins: {
+        // @ts-ignore
+        flowBuilderPlugin,
+      },
+    })
```
AinaVendrell, line 69: maybe we cold have a helper file with this logic encapsulated in methods

AinaVendrell (approved):

packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -91,11 +93,30 @@ export default class BotonicPluginFlowBuilder implements Plugin {
...
+  }
+
+  private updateBotActionPayload(payload: string): string {
+    if (payload.startsWith(BOT_ACTION_PAYLOAD_SEPARATOR)) {
+      const botActionId = payload.split(SEPARATOR)[1]
+      const botActionNode =
+        this.cmsApi.getNodeById<HtBotActionNode>(botActionId)
+      return this.cmsApi.createPayloadWithParams(botActionNode)
     }
+    return payload
```
asastre, line 119: Suggestion: I would move the condition that checks if the payload starts with the `BOT_ACTION_PAYLOAD_SEPARATOR` outside the function so the function will be executed only if we detect a bot action payload and I think the code will be clearer.

packages/botonic-plugin-flow-builder/tests/bot-action.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  createFlowBuilderPlugin,
+  createRequest,
+  getActionRequest,
+} from './helpers/utils'
+
+describe('Check the contents returned by the plugin', () => {
+  process.env.NODE_ENV = ProcessEnvNodeEnvs.PRODUCTION
+  const flowBuilderPlugin = createFlowBuilderPlugin(testFlow)
+
+  test('The user does the first interaction', async () => {
```
asastre, line 18: Observation/Change: Try to add a more descriptive names for the test to explain what you're checking with this test.
For example, in this case, I would change it to something like this: "The starting content is displayed on the first interaction"

packages/botonic-plugin-flow-builder/tests/bot-action.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      input: { data: 'Hola', type: INPUT.TEXT },
+      isFirstInteraction: true,
+      plugins: {
+        // @ts-ignore
+        flowBuilderPlugin,
+      },
+    })
+    await flowBuilderPlugin.pre(request)
+    const actionRequest = getActionRequest(request)
+    const { contents } = await FlowBuilderAction.botonicInit(actionRequest)
```
asastre, line 29: Observation: As @AinaVendrell said with another part, maybe this 3 lines can be encapsulated also within a function because you will have to execute them in every test

asastre (approved):
Everything should be working as expected, but I would like you @Iru89 to review my comments before merging 😉

packages/botonic-plugin-flow-builder/tests/bot-action.test.ts
Iru89: This is already a function inside the /helpers/utils.ts file
As Adria says in another comment I have encapsulated this logic that was repeated twice in a function.
`
await flowBuilderPlugin.pre(request)
    const actionRequest = getActionRequest(request)
    const { contents } = await FlowBuilderAction.botonicInit(actionRequest)
`

PR #2820, opened 2024-05-12
plugin-flow-builder: fallback , keyword, smart-intent tests

Iru89:
## Description

**Many lines added (292 in one file and 486 in the other) are from the JSON that represents the flow used for testing**

- Split the bot-action and first-interaction tests into two different files. 
- Add tests for fallback, keyword and smart intent. 
- Refactor the function getSmartIntentNodeByInput into a SmartIntentApi class and add a mock for the private function getInference of this class

## Context

We need to create tests to keep adding new features and refactor the plugin without changing the behaviour

For the smart intent test I have created a separate flow so that I don't always have to use the mock (we can discuss if this is the best way or if all tests should go in the same flow)

## Testing

This PR only contains changes to add tests and refactors the getSmartIntentNodeByInput function into a class and then creates a mock that modifies only one function of this class.

Add tests for fallback, keyword and smart intent


packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -98,23 +98,19 @@ export default class BotonicPluginFlowBuilder implements Plugin {
...
     }
   }
 
-  private removeSourceSeparatorFromPayload(payload: string): string {
+  private removeSourceSufix(payload: string): string {
     return payload.split(SOURCE_INFO_SEPARATOR)[0]
   }
 
-  private updateBotActionPayload(payload: string): string {
+  private changePayload(payload: string): string {
```
vanbasten17, line 113: i would name this with a more understandable name of what is doing exactly. I prefer the previous name or something that includes "replace". Also, would it make sense to have a specific class/set of functions to do extra logic for payloads?

vanbasten17 (approved):

PR #2839, opened 2024-05-27
plugin-knowledge-bases: new inference endpoint without knowledge base id

Iru89:
## Description

Use the new inference endpoint without the need for knowledge base id
With this new endpoint it is necessary to pass the userInput and the sources you want to use to get the response


elozano98 (approved):
Great job! I approve it despite I don't see any tests 🤔

vanbasten17 (approved):

PR #2852, opened 2024-06-13
fix(plugin-flow-builder): add is meaningful node tracking #874

vanbasten17:
## Description
* Updated plugin flow builder to use is_meaningful node


packages/botonic-plugin-flow-builder/src/content-fields/hubtype-fields/common.ts
```diff
@@ -33,6 +33,7 @@ export interface HtBaseNode {
   follow_up?: HtNodeLink
   target?: HtNodeLink
   flow_id: string // TODO: Review if this field is necessary in all HtBaseNode
+  is_meaningful: boolean
```
AinaVendrell, line 36: this should be optional

AinaVendrell (approved):

packages/botonic-plugin-flow-builder/src/content-fields/hubtype-fields/common.ts
vanbasten17: agree

PR #2853, opened 2024-06-13
plugin-flow-builder: resolve keywords, smart intents and intents in first interaction

Iru89:
## Description

If the user sends a text that matches a keyword, smart intent or intent in the first interaction, the bot's response is the contents of the start conversation followed by the contents of the match.

If the bot has an active knowledge base after trying to match the above and not getting a match, it will attempt to respond using the knowledge base. If it has sufficient knowledge and the response is faithful the bot will display the start contents followed by the response generated using the knowledge base.

## Context

We want the bot in platforms that are not proactive (such as Whatsapp) where the user is the first to write, if the user asks something in the first inbox, the bot will solve it the first time. But the initial contents have to be shown because they usually contain the GDPR and the AI disclaimer. 

I have created a function that replicates the logic used in the plugin's pre function to resolve the user input. 
In the pre function if it is the first interaction the user input is not resolved because we always want to show the first contents of the flow.
Refactor the ways to get the content in the action by splitting into different files how the content is obtained depending on whether it is the first interaction, has a payload in the input, uses a knowledge base or is the fallback. In the first interaction file, the other methods for resolving the follow ups are used. 

## Testing

Add test to check first interaction with:
- an input that no match any keyword, smart-intent, intent or knowledge base.
- an input that match with a keyword
- an input that match with a keyword that point to start contents
- an input that match with a keyword that point to a content 
- an input that match with a smart-intent that point to a content 
- an input that match with an intent that point to a content 
- an input that generate a knowledge response content

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/action/first-interaction.ts
+++ b/packages/botonic-plugin-flow-builder/src/action/first-interaction.ts
@@ -2,9 +2,9 @@ import { INPUT } from '@botonic/core'
 
 import { BOT_ACTION_PAYLOAD_PREFIX } from '../constants'
 import { FlowContent } from '../content-fields'
-import { HtNodeWithContent } from '../content-fields/hubtype-fields'
 import { getNodeByUserInput } from '../user-input'
 import { FlowBuilderContext } from './index'
+import { getContentsByPayload } from './payload'
 
 export async function getContentsByFirstInteraction({
   cmsApi,
@@ -23,9 +23,7 @@ export async function getContentsByFirstInteraction({
       resolvedLocale,
     })
 
-    if (contentsByUserInput) {
-      return [...firstInteractionContents, ...contentsByUserInput]
-    }
+    return [...firstInteractionContents, ...contentsByUserInput]
   }
 
   return firstInteractionContents
@@ -36,7 +34,7 @@ async function getContentsByUserInput({
   flowBuilderPlugin,
   request,
   resolvedLocale,
-}: FlowBuilderContext): Promise<FlowContent[] | undefined> {
+}: FlowBuilderContext): Promise<FlowContent[]> {
   const nodeByUserInput = await getNodeByUserInput(
     cmsApi,
     resolvedLocale,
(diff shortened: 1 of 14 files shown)


packages/botonic-plugin-flow-builder/src/action/first-interaction.ts
```diff
@@ -0,0 +1,78 @@
...
+export async function getContentsByFirstInteraction({
+  cmsApi,
+  flowBuilderPlugin,
+  request,
+  resolvedLocale,
+}: FlowBuilderContext): Promise<FlowContent[]> {
+  const firstInteractionContents =
+    await flowBuilderPlugin.getStartContents(resolvedLocale)
+
+  if (request.input.data && request.input.type === INPUT.TEXT) {
```
vanbasten17, line 19: i'm wondering if it could make sense to have a util function for this like `textInputHasData` or similar, as it is a check used in many places

packages/botonic-plugin-flow-builder/src/action/first-interaction.ts
```diff
@@ -0,0 +1,78 @@
...
+  const conversationStartId = cmsApi.flow.flows.find(
+    flow => flow.name === MAIN_FLOW_NAME
+  )?.start_node_id
+
+  if (request.input.payload === conversationStartId) {
+    return []
+  }
+
+  if (request.input.payload?.startsWith(BOT_ACTION_PAYLOAD_PREFIX)) {
+    request.input.payload = flowBuilderPlugin.replaceBotActionPayload(
```
vanbasten17, line 56: it seems strange to me to call the plugin itself if we are in botonic-plugin-flow-builder, should this be a separate function?

packages/botonic-plugin-flow-builder/src/action/first-interaction.ts
```diff
@@ -0,0 +1,78 @@
...
+  const contentsByKeywordsOrIntents = await getContentsByPayload({
+    cmsApi,
+    flowBuilderPlugin,
+    request,
+    resolvedLocale,
+  })
+
+  if (contentsByKeywordsOrIntents.length > 0) {
+    return contentsByKeywordsOrIntents
+  }
```
vanbasten17, line 70: suggestion for future: what if for functions like `getContentsBy{identifier}` we have a function signature that returns as well if it is empty or not? maybe we could save extra checks

vanbasten17 (approved):

packages/botonic-plugin-flow-builder/src/action/first-interaction.ts
Iru89: inputWithTextData?

packages/botonic-plugin-flow-builder/src/action/first-interaction.ts
Iru89: It must be a function of the plugin because it needs the flow (JSON that has been downloaded in the pre function of the plugin)

AinaVendrell (approved):

PR #2854, opened 2024-06-14
Botonic - React: scroll to bottom when typing

AlbertGom:
## Description
Fixed how scroll to bottom function was executed when typing changed its state.

## Context
The issue was that when a big message was rendered (not the entire message was visible in the screen), when I tried to scroll up, it was scrolling down automatically, don't letting me to scroll up.

The issue was happening because we were checking the isLastMessageVisible and isTyping at the same time.
The desired behaviour is that when the bot is typing then we want to scroll down so the user can see the last message shown. I think checking the isLastMessageVisible is not necessary.

Added a timeout so the scroll down action is smooth.

Edited (Oriol), check commit by commit:
- [commit](4181e4f7a6089b702a1bc8a9359231e850063312): I created the hook useNotifications to reuse the logic between UnreadMessagesBanner, ScrollButton, WebchatMessageList components. 

- [commit](1b9e08c9b42c556d1ce5820d764e79094d8f1c05): I have removed the react-reveal dependency that was used to have the Fade animation and was not working very well with the intersection observer. I've created a component that does the same animation using styled-components

- [commit](e648e03c9d00a87ddb2e21e181b31c2f5b32d28a): I have simplified the useEffect so that one runs when notifications are turned off and one runs when notifications are turned on, scrolling is always done using the message ref or the banner ref

- [commit](aba99947a00b7d61cd9e357c69b8e67b59410918): I have to observe the bottom of the message so that only the scroll down button is displayed while the whole message is not displayed. Before, the beginning of the message was observed and when the message was very long, the button was displayed if the beginning of the message was not inside the screen but the end of the message was

## Testing
Tested in WizzAir bot that had big messages rendered.
Edited (Oriol): Tested on a backline-bot with knowledge base without follow up, with very long messages, with notifications on and off.

### DIFF
--- a/packages/botonic-react/src/webchat/message-list/index.tsx
+++ b/packages/botonic-react/src/webchat/message-list/index.tsx
@@ -58,13 +58,10 @@ export const WebchatMessageList = props => {
   }, [webchatState.messagesComponents])
 
   useEffect(() => {
-    if (webchatState.isLastMessageVisible && webchatState.typing) {
+    setTimeout(() => {
       scrollToBottom({ host: props.host })
-    }
-    if (webchatState.isLastMessageVisible) {
-      scrollToBottom({ host: props.host })
-    }
-  }, [webchatState.typing, webchatState.isLastMessageVisible])
+    }, 100)
+  }, [webchatState.typing])
 
   useEffect(() => {
     if (firstUnreadMessageId) {
--- a/packages/botonic-react/src/webchat/message-list/index.tsx
+++ b/packages/botonic-react/src/webchat/message-list/index.tsx
@@ -9,8 +9,9 @@ import { IntroMessage } from './intro-message'
 import { ScrollButton } from './scroll-button'
 import { ContainerMessage } from './styles'
 import { UnreadMessagesBanner } from './unread-messages-banner'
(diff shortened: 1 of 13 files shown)


AlbertGom:
@Iru89 We should check that it still works with a bot that have notification enabled (since a scroll down behaviour is happenning there), but this change added in this PR should not affect that part.

Iru89 (approved):

Iru89 (approved):

vanbasten17 (approved):

packages/botonic-react/src/webchat/message-list/index.tsx
```diff
@@ -97,9 +112,7 @@ export const WebchatMessageList = props => {
         scrollbar={scrollbarOptions}
         autoHide={scrollbarOptions.autoHide}
         ismessagescontainer={true.toString()}
-        style={{
-          ...props.style,
-        }}
+        style={{ flex: 1 }}
```
vanbasten17, line 115: why not passing this flex directly to the styled component?

vanbasten17:
Beware updating the changelog for these changes: @AlbertGom @Iru89

PR #2855, opened 2024-06-14
Feat/handle preview for is_test_integration apps

vanbasten17:
## Description
* Type `is_test_integration` in `Session`.
* Avoid calling botonic-plugin-hubtype-analytics when bot has `session.is_test_integration` set to true.
* Do call smart intents inference with use_latest parameter set to false if bot has `session.is_test_integration` set to true.
* Flow Builder plugin to fetch only draft version of the flow if bot has `session.is_test_integration` set to true.
* FB Plugin and core coordinated to being able to execute a first input with a simulation of the handoff, and a preceding input to emulate the on finish payload after this faked handoff.

## Context
This PR contains the needed implementations in the botonic side to avoid test integration bots to produce not expected data and pollute metrics, this is, not calling analytics, faking handoff, ... if bots are of this type.

## Testing
 
- Add botonic-core: handoff with is_test_integration and core-bot tests
- Fix plugin-flow-builder tests
- Fix plugin-hubtype-analytics tests

### DIFF
--- a/packages/botonic-core/src/core-bot.ts
+++ b/packages/botonic-core/src/core-bot.ts
@@ -85,11 +85,63 @@ export class CoreBot {
     session.__locale = locale
   }
 
-  async input({
-    input,
-    session,
-    lastRoutePath,
-  }: BotRequest): Promise<BotResponse> {
+  async input({ input, session, lastRoutePath }: any): Promise<BotResponse> {
+    const output = await this.runInput({ input, session, lastRoutePath })
+    if (!session.is_test_integration) {
+      return output
+    }
+    if (!session._botonic_action?.startsWith('create_test_integration_case:')) {
+      return output
+    }
+
+    const [_, onFinishPayloadId] = session._botonic_action.split(
+      'create_test_integration_case:'
+    )
+
+    const newInput = {
+      ...input,
+      payload: onFinishPayloadId,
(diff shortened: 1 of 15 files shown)


packages/botonic-core/tests/core-bot/core-bot.test.ts
Iru89, line 1: one question what are we currently using the __locale and __retries attributes of the session for?
I think we are not using it in our bots.

packages/botonic-core/src/core-bot.ts
```diff
@@ -81,61 +81,33 @@ export class CoreBot {
...
-    session._botonic_action = undefined
-
-    const newOutput = (this.router as Router).processInput(
-      newInput,
-      output.session,
-      output.lastRoutePath
-    )
-
-    const newRequest = {
+  createRequestFromOutput({ input, session, lastRoutePath }, output) {
```
Iru89, line 84: Now that we have this function, it would be a good time to type this BotRequest and use this type in all the points of the bot (pre and post functions of the plugin, routes and botonicInit of the actions). 
Maybe it's better to do it in another PR to avoid mixing concepts

Iru89 (approved):

AinaVendrell (approved):

PR #2861, opened 2024-06-26
Botonic-React: change staticAsset function + add unit tests

AlbertGom:
## Description
Changed how `staticAsset` function is searching for the script to resolve the assets path.

## Context
The function is only searching for scripts that end with "webchat.botonic.js" so if the webchat file has another name it won't work.

Change the query done in the function to check that the script contains "webchat.botonic.js"

## Testing

Added unit tests to test the function.

### DIFF
--- a/packages/botonic-react/tests/utils.test.js
+++ b/packages/botonic-react/tests/utils.test.js
@@ -1,5 +1,8 @@
+/**
+ * @jest-environment jsdom
+ */
+import { isURL, staticAsset } from '../src/util/environment'
 import { deserializeRegex, stringifyWithRegexs } from '../src/util/regexs'
-
 describe('Regex serialization / deserialization', () => {
   const regexsItems = [
     { regex: /regex/i, string: '"/regex/i"', samples: ['Regex', 'regex'] },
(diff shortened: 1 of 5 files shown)


packages/botonic-react/tests/utils.test.js
```diff
@@ -26,3 +29,100 @@ describe('Regex serialization / deserialization', () => {
...
+}
+
+describe('staticAsset function', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  test('should return the path if it is a URL', () => {
+    //arrange
+    isURL.mockReturnValue(true)
```
vanbasten17, line 57: why mocking isURL? in this case it is a url in fact, right?

packages/botonic-react/tests/utils.test.js
```diff
@@ -26,3 +29,100 @@ describe('Regex serialization / deserialization', () => {
...
+    //act
+    const result = staticAsset(url)
+
+    //assert
+    expect(result).toBe(url)
+  })
+
+  test('should resolve the static asset path if it is not a URL and file name is webchat.botonic.js', () => {
+    //arrange
+    isURL.mockReturnValue(false)
```
vanbasten17, line 69: same here

packages/botonic-react/tests/utils.test.js
AlbertGom: Mocking it ensures that what I am testing is the behaviour of the `staticAsset` function without relying on the correctness of `isURL` method. In fact, the isURL function should have a unit test itself.

vanbasten17: yes, this was exactly what I meant..

vanbasten17 (approved):

Iru89 (approved):

PR #2863, opened 2024-06-27
plugin-flow-builder: create contents with map

Iru89:
## Description

- The functions `getTextContent` and `getImageSrc` return an empty string if it does not find the value.

- The `useWebviewContents` hook accept an object mapContents. This object relates the names the developer wants for each content to the contentID of each content. Using this object the hook returns a contents object with all the contents. 

## Context

The hook useWebviewContents now returns a contents object which is like the mapContents passed as a parameter but with the relation `key: fetchedContent`
An important point is to have the `contents` object typed the same as the `mapContents`

To have the `contents` typed like the `mapContents` it is necessary create `MyWebviewContentsContext` outside the webview component and export it to be able to use it inside the components. 
To create this `MyWebviewContentsContext` we use the `createWebviewContentsContext` function that allows to pass the type of the used in `mapContents`, by creating the type of this object or by passing `typeof mapContents`. 
This way when using `useContext(MyWebviewContentsContext) `we get the contents typed as the `mapContents`.  

```typescript
  type WebviewContents = {
    textIntro: string
    image2: string
    headerWebview: string
    image: string
  }

export const MyWebviewContentsContext = createWebviewContentsContext<WebviewContents>()

export const FlowBuilderWebview = () => {
  const webviewRequestContext = useContext(WebviewRequestContext)

  const { isLoading, error, webviewContentsContext } = useWebviewContents({
    apiUrl: FLOW_BUILDER_API_URL,
    version: FlowBuilderJSONVersion.LATEST,
    orgId: webviewRequestContext.session.organization_id,
    botId: webviewRequestContext.session.bot.id,
    webviewId: WEBVIEW_ID,
    locale: 'es',
    mapContents: {
      textIntro: 'TEXT_INTRO',
      image2: 'IMAGE_2',
      headerWebview: 'HEADER_WEBVIEW',
      image: 'IMAGE',
    },
  })

  if (error) {
    return <div>Error</div>
  }

  return (
    <MyWebviewContentsContext.Provider value={webviewContentsContext}>
      // your components
      <Step1 />
    </MyWebviewContentsContext.Provider>
  )
}
```

In Step1 component you can read the contents from context like 

```typescript
export const Step1 = () => {
  const webviewContentsContext = useContext(MyWebviewContentsContext)

  const { contents } = webviewContentsContext

  return (
    <div>
      <h1>Step2</h1>

      <div>
        <p>{contents.textIntro}</p>
        <img style={{ width: '350px' }} src={contents.image2} />
      </div>
    </div>
  )
}
```

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/webview/contents-context.ts
+++ b/packages/botonic-plugin-flow-builder/src/webview/contents-context.ts
@@ -1,6 +1,6 @@
 import { createContext } from 'react'
 
-import { WebviewContentsContextType } from './types'
+import { MapContentsType, WebviewContentsContextType } from './types'
 
 /* 
    Define a generic type parameter for your context
@@ -26,7 +26,7 @@ import { WebviewContentsContextType } from './types'
 
     const { contents } = useContext(MyWebviewContentsContext)
 */
-export const WebviewContentsContext = <T>() =>
+export const createWebviewContentsContext = <T extends MapContentsType>() =>
   createContext<WebviewContentsContextType<T>>({
     getTextContent: () => '',
     getImageSrc: () => '',
--- a/packages/botonic-plugin-flow-builder/src/webview/contents-context.ts
+++ b/packages/botonic-plugin-flow-builder/src/webview/contents-context.ts
@@ -4,8 +4,9 @@ import { WebviewContentsContextType } from './types'
 
 export const WebviewContentsContext = createContext<WebviewContentsContextType>(
   {
-    getTextContent: () => undefined,
-    getImageSrc: () => undefined,
+    getTextContent: () => '',
+    getImageSrc: () => '',
     setCurrentLocale: () => undefined,
+    contents: {},
   }
 )
--- a/packages/botonic-plugin-flow-builder/src/webview/contents-context.ts
+++ b/packages/botonic-plugin-flow-builder/src/webview/contents-context.ts
@@ -2,11 +2,34 @@ import { createContext } from 'react'
 
 import { WebviewContentsContextType } from './types'
 
-export const WebviewContentsContext = createContext<
-  WebviewContentsContextType<unknown>
->({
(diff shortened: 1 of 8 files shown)


packages/botonic-plugin-flow-builder/src/webview/use-webview-contents.ts
```diff
@@ -19,16 +19,43 @@ export function useWebviewContents({
...
+        ?.content.image.find(image => image.locale === currentLocale)?.file ||
+      ''
+    )
+  }
+
+  const createContentsObject = () => {
+    const contentsObject = {}
+    for (const [key, value] of Object.entries(mapContents)) {
+      contentsObject[key] = getTextContent(value)
+    }
```
asastre, line 53: Change: You're missing the image sources to be obtanined also.
```suggestion
    for (const [key, contentId] of Object.entries(mapContents)) {
     
      contentsObject[key] = imageContents?.find(imageContent => imageContent.code === contentId) ? getImageSrc(contentId) :  getTextContent(contentId)
    }
```

asastre (changes requested):
Just a 1 change, the rest LGTM

packages/botonic-plugin-flow-builder/src/webview/contents-context.ts
```diff
@@ -2,10 +2,34 @@ import { createContext } from 'react'
...
+
+    Then you can use it in your component like this:
+    <MyWebviewContentsContext.Provider value={webviewContentsContext}>
+    ...// your components
+    </MyWebviewContentsContext.Provider>
+
+    In your components you can use the context to read contents like this:
+
+    const { contents } = useContext(MyWebviewContentsContext)
+*/
```
vanbasten17, line 28: better to have this documented in the readme, instead of here

vanbasten17 (approved):
nice feature!

PR #2874, opened 2024-07-15
Botonic: upgrade typescript to 5.5.3

Iru89:
## Description

Upgrade typescript used in botonic to [v5.5.3](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/)

## Context

Remove a type in the Contentful plugin as it is not typed correctly and when inferring the type automatically with the new version of typescript, an error occurs.

Removes the use of as T (in several flow builder plugin files) after using a filter on an array as the new version of typescript correctly infers the type.


packages/botonic-react/package.json
```diff
@@ -56,8 +56,7 @@
     "copyfiles": "^2.4.1",
     "identity-obj-proxy": "^3.0.0",
     "intersection-observer": "^0.12.2",
-    "react-test-renderer": "^16.14.0",
-    "typescript": "5.0.4"
```
vanbasten17, line 60: nice catch

tsconfig.esm.base.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
-    "module": "ES2022",
+    "module": "Node16",
```
vanbasten17, line 4: beware with this change. Here https://www.typescriptlang.org/tsconfig/#module is stated that this changes the moduleResolution. I think we have to be sure that the compiled code is the one we expect

vanbasten17 (commented):
Looking good ATM, but I would like to have this tested and be sure we are not breaking the code we are compiling and distributing

tsconfig.esm.base.json
Iru89: Yes you are right I had to change this to make the build but now it is failing some tests in core related to regexp. I have to check better how the build is made

PR #2881, opened 2024-08-01
Blt 974 update flow builder plugin

Iru89:
## Description

Use knowledge base with sourceIds instead of sourceNames

## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-plugin-flow-builder/src/content-fields/hubtype-fields/knowledge-base.ts
```diff
@@ -5,6 +5,9 @@ export interface HtKnowledgeBaseNode extends HtBaseNode {
   type: HtNodeWithContentType.KNOWLEDGE_BASE
   content: {
     feedback_enabled: boolean
-    sources: string[]
+    sources_data: {
```
vanbasten17, line 8: do you think it's worth to have an interface `SourceData`? is this type going to change a lot? if not it's ok like this, but maybe it's better to have it defined in an interface

vanbasten17 (approved):

AinaVendrell (approved):

PR #2883, opened 2024-08-02
botonic-react: create WebchatInputPanel component

Iru89:
## Description

Create WebchatInputPanel to group all components for user input (text input, emoji picker, send attachment and send button)

## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-react/src/webchat/webchat-input-panel/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+  return (
+    <UserInputContainer
+      style={{
+        ...getThemeProperty(WEBCHAT.CUSTOM_PROPERTIES.userInputStyle),
+      }}
+      className='user-input-container'
+    >
+      {webchatState.isEmojiPickerOpen && (
+        <OpenedEmojiPicker
+          // height={webchatState.theme.style.height} // Revisar si es necessari
```
vanbasten17, line 86: i think it is needed when webchat is displayed in phone-like screens

packages/botonic-react/src/webchat/webchat-input-panel/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+      <EmojiPicker
+        enableEmojiPicker={enableEmojiPicker}
+        onClick={handleEmojiClick}
+      />
+
+      <Attachment
+        enableAttachments={enableAttachments}
+        onChange={handleAttachment}
+        accept={getFullMimeWhitelist().join(',')}
```
vanbasten17, line 110: not for this PR, but I'm not a fan of having a join here, I think we can think of a more nicer approach

PR #2885, opened 2024-08-07
cli: use async await insted of then in deploy

Iru89:
## Description

Refactor the deploy file to always use async await instead of chaining asyncorn functions with callbacks inside the then function.

## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -120,43 +120,41 @@ Deploying to AWS...
...
   }
 
-  signupFlow(): Promise<void> {
+  async signupFlow(): Promise<void> {
     const choices = [
       'No, I need to create a new one (Signup)',
       'Yes, I do. (Login)',
     ]
-    return prompt([
+    const inp = await prompt([
```
vanbasten17, line 144: const userInput

vanbasten17: const {signupConfirmation} = await ...

packages/botonic-cli/src/commands/deploy.ts
```diff
@@ -120,43 +120,41 @@ Deploying to AWS...
...
         message:
           'You need to login before deploying your bot.\nDo you have a Hubtype account already?',
         choices: choices,
       },
-    ]).then((inp: any) => {
-      if (inp.signupConfirmation == choices[1]) return this.askLogin()
-      else return this.askSignup()
-    })
+    ])
+    if (inp.signupConfirmation == choices[1]) return this.askLogin()
```
vanbasten17, line 153: ===

vanbasten17 (approved):

PR #2894, opened 2024-08-26
botonic-cli: use axios.interceptor for update refreshToken

Iru89:
## Description

Use axios interceptors to update token using refreshToken when response status is 401

## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-cli/src/botonic-api-service.ts
```diff
@@ -30,20 +50,50 @@ export class BotonicAPIService {
   oauth?: OAuth
   me?: Me
   analytics: AnalyticsInfo
-  bot: BotInfo | null
-  headers: Record<string, string> | null = null
+  bot: BotDetail | null
+  headers: AxiosHeaders
+  apiClient: AxiosInstance
```
MarcosCA, line 55: axiosClient better no?

MarcosCA (approved):

packages/botonic-cli/src/botonic-api-service.ts
```diff
@@ -1,4 +1,10 @@
-import axios, { AxiosPromise, AxiosResponse } from 'axios'
+/* eslint-disable @typescript-eslint/naming-convention */
```
vanbasten17, line 1: why is complaining here eslint?

packages/botonic-cli/src/botonic-api-service.ts
```diff
@@ -243,39 +257,55 @@ export class BotonicAPIService {
...
     resp.data.results.map(b => bots.push(b))
     nextBots = resp.data.next
+
     return this.getMoreBots(bots, nextBots)
   }
 
   async getProviders(): Promise<AxiosPromise> {
     return this.apiGet({
-      path: '/v1/provider_accounts/',
+      path: 'provider_accounts/',
```
vanbasten17, line 308: i'm wondering if it will be helpful to have a kind of URL builder or helper. I think that this trailing slash `/` sometimes makes the request to work or not, so having something to avoid this kind of problems would be nice. Up to you

vanbasten17 (approved):

packages/botonic-cli/src/botonic-api-service.ts
Iru89: This is to avoid seeing in many lines (e.g. when creating headers) that the attributes have to be with camelCase.

packages/botonic-cli/src/botonic-api-service.ts
Iru89: This seems to be because of the configuration of django in the backend

PR #2895, opened 2024-08-27
botonic-cli: no read/write bot.config.json

Iru89:
## Description

- Not read or write the bot.config.json file, just get the bot config values to send them in the post when deploying.
- Use the ora spinner to give feedback to the user: Getting bot config...

### DIFF
--- a/packages/botonic-cli/src/util/bot-config.ts
+++ b/packages/botonic-cli/src/util/bot-config.ts
@@ -1,9 +1,9 @@
 import childProcess from 'child_process'
 import { promises as fs } from 'fs'
+import ora from 'ora'
 import path from 'path'
 import * as util from 'util'
 
-const OUTPUT_JSON_NAME = 'bot.config.json'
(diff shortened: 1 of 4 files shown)


packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -12,42 +12,46 @@ const NPM_DEPTH_0 = 0
...
+  packages: BotonicDependencies
 }
 export class BotConfig {
-  constructor(public appDirectory: string) {}
-
-  async createJson(): Promise<BotConfigJson> {
-    const packages = await this.readPackageJson()
+  static async get(appDirectory: string): Promise<BotConfigJSON> {
+    const spinner = ora({
+      text: 'Getting bot config...',
```
vanbasten17, line 28: i would remove this message, I think it is not needed for the user to know this is happening in the background

vanbasten17 (approved):

PR #2898, opened 2024-09-04
plugin-flow-builder: bot action as follow up, first node in a flow, after conditionals etc

Iru89:
## Description

Be able to use bot-action nodes at any point in the flow builder. For example as follow ups, after a conditional, at the start of a flow etc

The request.input should be immutable. To avoid modifying the request.input in the lambda execution, FlowBuilderAction.botonicInit function accepts as a second parameter the contentID to indicate which content has to be rendered. This contentID is the unique text that is defined in the flow builder frontend. This way you don't need to get the UUID of the content and modify the request.input.payload to indicate to the plugin which content to render.

## Context

Bot-action nodes could only be used connected to a button or after a handoff.
With this solution it will be possible to make flow in a simpler way. There are several use cases in which the flow is greatly simplified by being able to put bot actions after a conditional or at the beginning of a flow.

Flow example:
Captura de pantalla 2024-09-03 a las 17 11 40

The third commit makes the core bot run again (prePlugins, routes, render, postPlugins). I think this should be the backend in charge of re-executing the lambda of the bot to avoid timeouts if the actions are asynchronous and more than one is chained. This way we would also avoid modifying the request.input.payload in a lambda execution.

IMPORTANT
With this change in the custom actions written in the flow it is necessary to obtain the following contents using the function FlowBuilderAction.botonicInit(request) since this is the one in charge of putting the variable in the session so that if there is another bot action connected to the flow, the bot will continue executing. 

WelcomeAction example
```ts
export class WelcomeAction extends FlowBuilderMultichannelAction {
  static async botonicInit(
    request: BotRequest
    const { cmsPlugin, botContext } = getRequestData(request)
    const contents = await cmsPlugin.getStartContents(botContext.locale)

     return FlowBuilderMultichannelAction.botonicInit(request, contents[0].code)
  }
}
```

## Testing

For now I have not checked the tests, I am aware that they are failing.

### DIFF
--- a/packages/botonic-core/src/core-bot.ts
+++ b/packages/botonic-core/src/core-bot.ts
@@ -275,8 +275,13 @@ export class CoreBot {
 
   private async runRedirectAction(
     previousResponse: BotResponse,
-    { input, session, lastRoutePath }: BotRequest
+    { input, session, lastRoutePath }: BotRequest,
+    numOfRedirects: number = 0
   ) {
+    if (numOfRedirects > 10) {
+      throw new Error('Maximum BotAction recursive calls reached (10)')
+    }
+
     const nextPayload = session._botonic_action?.split(
       BotonicAction.Redirect
     )[1]
@@ -306,11 +311,15 @@ export class CoreBot {
 
     // Recursive call to handle multiple bot actions in a row
     if (response.session._botonic_action?.startsWith(BotonicAction.Redirect)) {
-      return await this.runRedirectAction(response, {
-        input,
-        session,
-        lastRoutePath,
-      })
+      return await this.runRedirectAction(
+        response,
+        {
+          input,
+          session,
+          lastRoutePath,
+        },
+        numOfRedirects + 1
+      )
     }
 
     return response
--- a/packages/botonic-core/src/core-bot.ts
+++ b/packages/botonic-core/src/core-bot.ts
@@ -322,6 +322,16 @@ export class CoreBot {
       )
     }
 
+    if (
+      response.session._botonic_action?.startsWith(BotonicAction.CreateTestCase)
+    ) {
+      return await this.runFollowUpTestIntegrationInput(response, {
+        input,
+        session,
+        lastRoutePath,
+      })
+    }
+
     return response
   }
 }
--- a/packages/botonic-core/src/core-bot.ts
+++ b/packages/botonic-core/src/core-bot.ts
@@ -283,7 +283,7 @@ export class CoreBot {
     }
 
     const nextPayload = session._botonic_action?.split(
-      BotonicAction.Redirect
+      `${BotonicAction.Redirect}:`
     )[1]
 
     const inputWithBotActionPayload: Input = {
--- a/packages/botonic-core/src/core-bot.ts
+++ b/packages/botonic-core/src/core-bot.ts
@@ -90,6 +90,15 @@ export class CoreBot {
     lastRoutePath,
   }: BotRequest): Promise<BotResponse> {
     const output = await this.runInput({ input, session, lastRoutePath })
+
+    if (session._botonic_action?.startsWith('flow_builder_bot_action:')) {
+      return await this.runFollowUpFlowBuilderBotAction(output, {
+        input,
+        session,
+        lastRoutePath,
+      })
+    }
+
     if (session._botonic_action?.startsWith('create_test_integration_case:')) {
       return await this.runFollowUpTestIntegrationInput(output, {
         input,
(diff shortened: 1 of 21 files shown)


packages/botonic-plugin-flow-builder/src/content-fields/flow-bot-action.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  static fromHubtypeCMS(
+    cmsBotAction: HtBotActionNode,
+    _locale: string,
+    cmsApi: FlowBuilderApi
+  ): FlowBotAction {
+    const newBotAction = new FlowBotAction(cmsBotAction.id)
+    newBotAction.code = cmsBotAction.code
+    newBotAction.payloadId = cmsBotAction.content.payload_id
+    newBotAction.payloadParams = cmsBotAction.content.payload_params
```
Iru89, line 21: It is not necessary to store these two values in FlowBotAction

packages/botonic-core/src/core-bot.ts
```diff
@@ -90,6 +90,15 @@ export class CoreBot {
     lastRoutePath,
   }: BotRequest): Promise<BotResponse> {
     const output = await this.runInput({ input, session, lastRoutePath })
+
+    if (session._botonic_action?.startsWith('flow_builder_bot_action:')) {
+      return await this.runFollowUpFlowBuilderBotAction(output, {
```
ericmarcos, line 95: Let's change these names to something generic and not related to "Flow Builder". The core package should not know that the flow builder exists, that's why we have a separate package for that. Otherwise is like introducing a circular dependency. I propose to call it `redirect_action`, the same with the function name.

packages/botonic-core/src/core-bot.ts
```diff
@@ -261,4 +270,49 @@ export class CoreBot {
...
+      response: previousResponse.response.concat(followUp.response),
+      session,
+      lastRoutePath: followUp.lastRoutePath,
+    }
+
+    // Recursive call to handle multiple bot actions in a row
+    if (
+      response.session._botonic_action?.startsWith('flow_builder_bot_action:')
+    ) {
+      return await this.runFollowUpFlowBuilderBotAction(response, {
```
ericmarcos, line 309: Be careful when using recursive functions. We need to ensure it's not possible to run this infinitely. In the current implementation we're not protected against this (imagine a bot action that is connected to itself, or a chain of bot action 1 -> bot action 2 -> bot action 1). If this happens it's very hard to debug because you just see the process running out of memory or other strange errors.
I suggest adding an additional parameter "n" (by default 0) and in this call doing `return await this.runFollowUpFlowBuilderBotAction(..., n=n+1)` then adding a conditional in the function like `if(n>10) throw new Error('Maximum BotAction recursive calls reached (10)');`

packages/botonic-core/src/models/legacy-types.ts
```diff
@@ -316,3 +316,9 @@ export type MatchingProp =
   | 'request'
 
 export type Matcher = string | RegExp | ((args) => boolean)
+
+export enum BotonicAction {
+  Redirect = 'redirect_action:',
+  CreateCase = 'create_case:',
+  CreateTestCase = 'create_test_integration_case:',
+}
```
vanbasten17, line 324: i would change this a little bit in favour of something like:
```

type BotonicActionSeparator = ':';

enum BotonicActionType {
  Redirect = 'redirect_action',
  CreateCase = 'create_case',
  CreateTestCase = 'create_test_integration_case',
}

type BotonicAction = `${BotonicActionType}${BotonicActionSeparator}${string}`;

```

To be more precise, here an example: 
const action:BotonicAction = `${BotonicActionType.Redirect}:blablabla`
And this will make tsc to fail: 
const action:BotonicAction = `${BotonicActionType.Redirect}--blablabla`

Also in this way we can have this value in the Session correctly typed

packages/botonic-plugin-flow-builder/src/index.ts
```diff
@@ -179,8 +165,7 @@ export default class BotonicPluginFlowBuilder implements Plugin {
       contents.push(content)
     }
     // TODO: prevent infinite recursive calls
-
-    if (node.follow_up) {
+    if (node.follow_up && !(content instanceof FlowBotAction)) {
```
vanbasten17, line 168: store the condition in a variable which represents clear this case, or create a private/auxiliary function to encapsulate this

vanbasten17 (approved):
Add at least a test for this new scenario 🙏

ericmarcos (approved):

packages/botonic-plugin-flow-builder/src/index.ts
Iru89: applied

PR #2900, opened 2024-09-06
refactor: get rid of Simplebar (vanilla js based scrollbar)

vanbasten17:
This PR include this other one [PR#2903](https://github.com/hubtype/botonic/pull/2903)

## Work done so far
1. [Getting rid of `scrollbar` webchat property customization](https://github.com/hubtype/botonic/pull/2900/commits/b8b676e02660e29c554be3bf3cc0c2c635f62159): We don't want to support this feature anymore. This was done before in order to have "nice" styles in Internet Explorer and similar browsers. As IE is not supported anymore, we can remove this.
2. [Having ids for relevant botonic containers](https://github.com/hubtype/botonic/pull/2900/commits/79a2bf878df196e4ee87d1855416fa813cbe0fef): In this way we can have a better way to access the values of these well differentiated areas of the webchat.
3. [Ensuring elements within webchat are spaced correctly](https://github.com/hubtype/botonic/pull/2900/commits/cd37e6ff30e22011c9bf68f29e246f172e10c7b3): This is, spacing elements between in webchat for every container to push the other siblings with same distances.
4. [Remove Simplebar in favour of a new component created by us](https://github.com/hubtype/botonic/pull/2900/commits/e1a937c67bea4672391a0aa0daa04d02aed72644): Simplebar was giving lots of headaches to us, so we can change it by containers with the ids created in point 2.
5. [Fix replies container not scrolling](https://github.com/hubtype/botonic/pull/2900/commits/72f3eeb6a6b634ff8984d9738d595b62e6b4da2d)
6. [chatArea container to englobe messages list and replies containers](https://github.com/hubtype/botonic/pull/2900/commits/7dcafa0446a000c2c6e34481540f657a8cae5ad7)
7. [Recalculate chatarea size when iOS safari virtual keyboard is shown](https://github.com/hubtype/botonic/pull/2900/commits/383ff18701c262d308af293afb9999ab5db55250): As we have modified the containers which where important to calculate when the webchat has to be resized in iOS, we needed to adapt and update a little bit this logic.

## Current state

As per the list described above, the functionality should be the same as before. The important thing to notice here is that **WE DON'T RELY ANYMORE** on simplebar selectors `'.simplebar-content'` and `'.simplebar-content-wrapper'`, so bots using them should be adapted.

## Next steps
* Create reusable hooks to do calculations over DOM elements or use [Resizer observer API](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to have the code up to date with new standards and good code practices.
* Remove related logic not needed anymore after removing simplebar dependency (possibly tricky parts of the code related to ShadowDOM).


packages/botonic-react/src/util/dom.js
```diff
@@ -24,7 +21,8 @@ export const scrollToBottom = ({
 } = {}) => {
   const webchatElement = getWebchatElement(host)
   if (!webchatElement) return
-  const frame = getScrollableArea(webchatElement).visible
+  const frame = getScrollableArea(webchatElement).full
```
Iru89, line 24: why did you change the frame from visible to full?

packages/botonic-react/src/webchat/devices/scrollbar-controller.js
```diff
@@ -51,8 +51,10 @@ export class ScrollbarController {
 
   hasScrollbar() {
     const scrollableArea = getScrollableArea(this.webchat)
+
     const isScrollable =
-      scrollableArea.visible.clientHeight - scrollableArea.full.clientHeight < 0
+      scrollableArea.full.scrollHeight > scrollableArea.visible.clientHeight
```
Iru89, line 56: I understand that it has to be the same result but why didn't you keep the same operation?

packages/botonic-react/src/webchat/constants.ts
Iru89, line 1: Maybe it would be better to use ref in the components instead of ids

packages/botonic-react/src/components/carousel.jsx
```diff
@@ -77,6 +73,7 @@ export const Carousel = props => {
     WEBCHAT.CUSTOM_PROPERTIES.customCarouselRightArrow,
     undefined
   )
+  // TODO: Investigate why when set to false, scroll is enabled via dragging the bar but not hovering the carousel elements
```
Iru89, line 76: What happens in the carousel?

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -680,6 +693,21 @@ export const Webchat = forwardRef((props, ref) => {
...
+    const headerHeight = document.getElementById(
+      BotonicContainerId.Header
+    )?.clientHeight
+    const inputPanelHeight = document.getElementById(
+      BotonicContainerId.InputPanel
+    )?.clientHeight
+    if (webchatHeight && headerHeight && inputPanelHeight) {
+      setChatAreaHeight(webchatHeight - headerHeight - inputPanelHeight)
+    }
+  }, [webchatState.isWebchatOpen])
```
Iru89, line 710: We should encapsulate this logic in a custom hook

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -735,13 +764,22 @@ export const Webchat = forwardRef((props, ref) => {
...
-            )}
+          <div
+            id={BotonicContainerId.ChatArea}
+            style={{
+              display: 'inherit',
+              flexDirection: 'inherit',
+              height: chatAreaHeight,
+              width: 'inherit',
+              overflow: 'inherit',
+            }}
```
Iru89, line 775: These styles should be in a styled component. 
I think there is no need to add a div here. We could do this inside the WebchatMessageList component.

packages/botonic-react/src/webchat/webchat.jsx
```diff
@@ -735,13 +764,22 @@ export const Webchat = forwardRef((props, ref) => {
...
+          >
+            <WebchatMessageList />
+            {webchatState.replies &&
+              Object.keys(webchatState.replies).length > 0 && (
+                <WebchatReplies replies={webchatState.replies} />
+              )}
```
Iru89, line 781: This should be a component within the WebchatMessageList

packages/botonic-react/src/components/carousel.jsx
vanbasten17: fixed in next PR

packages/botonic-react/src/util/dom.js
vanbasten17: now that the content is pushed in the other direction, this condition changes (this was due to how simplebar was creating the layout)

packages/botonic-react/src/webchat/devices/scrollbar-controller.js
vanbasten17: i removed full and visible properties form object, now they are get directly from `ScrollableMessagesListRef` and `ChatAreaRef`

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: yep, done in the next PR

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: done in the next PR

packages/botonic-react/src/webchat/webchat.jsx
vanbasten17: leave it for next refactors 👌

Iru89 (approved):

PR #2913, opened 2024-10-02
improve flowbuilder webview contents #BLT-1131 

Iru89:
## Description

- Resolve locale as a language when get webview contents. For example if locale is en-GB and your contents are only in en
- set organzation_id and bot_id in session when bot is in local development.

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/api.ts
+++ b/packages/botonic-plugin-flow-builder/src/api.ts
@@ -17,10 +17,11 @@ import {
   HtPayloadNode,
 } from './content-fields/hubtype-fields'
 import { HtSmartIntentNode } from './content-fields/hubtype-fields/smart-intent'
-import { FlowBuilderApiOptions } from './types'
+import { FlowBuilderApiOptions, ProcessEnvNodeEnvs } from './types'
 
 export class FlowBuilderApi {
   url: string
+  flowUrl: string
   flow: HtFlowBuilderData
   request: PluginPreRequest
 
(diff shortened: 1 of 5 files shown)


packages/botonic-plugin-flow-builder/src/webview/use-webview-contents.ts
```diff
@@ -31,6 +31,28 @@ export function useWebviewContents<T extends MapContentsType>({
...
+    if (textContents?.[0].content.text.some(text => text.locale === language)) {
+      setCurrentLocale(language)
+      return
+    }
+
+    console.error(
+      `locale: ${currentLocale} cannot be resolved with: ${textContents?.[0].content.text.map(text => text.locale).join(', ')}`
+    )
+    return
+  }
```
vanbasten17, line 54: Do happy path and early returns
```suggestion
const updateCurrentLocale = (textContents?: WebviewTextContent[]) => {
  if (!textContents || textContents.length === 0) {
    return;
  }

  const textContents = textContents[0].content.text;
  
  if (textContents.some(text => text.locale === currentLocale)) {
    setCurrentLocale(currentLocale);
    return;
  }
  
  const language = currentLocale.split('-')[0];
  if (textContents.some(text => text.locale === language)) {
    setCurrentLocale(language);
    return;
  }

  console.error(
    `locale: ${currentLocale} cannot be resolved with: ${textArray.map(text => text.locale).join(', ')}`
  );
};
```

vanbasten17 (approved):

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -30,19 +31,33 @@ export class FlowBuilderApi {
...
       headers: { Authorization: `Bearer ${token}` },
     })
     return data as HtFlowBuilderData
   }
 
+  private async updateSessionWithUserInfo(token: string) {
+    const url = `${this.url}/get-user-info`
+    const response = await axios.get(url, {
+      headers: { Authorization: `Bearer ${token}` },
+    })
```
AlbertGom, line 56: I know it will only be executed in development env, but I recommend to encapsulate calls in a try catch block to avoid the bot "crashing".

packages/botonic-plugin-flow-builder/src/webview/use-webview-contents.ts
Iru89: applied

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -30,19 +31,33 @@ export class FlowBuilderApi {
...
+    this.request.session.organization_id = response.data.organization_id
+    this.request.session.bot.id = response.data.bot_id
```
AlbertGom, line 58: :clap:

AlbertGom (approved):

packages/botonic-plugin-flow-builder/src/webview/use-webview-contents.ts
```diff
@@ -31,6 +31,31 @@ export function useWebviewContents<T extends MapContentsType>({
...
   const [error, setError] = useState(false)
 
+  const updateCurrentLocale = (textContents?: WebviewTextContent[]) => {
+    if (!textContents || textContents.length === 0) {
+      console.log('There is no text contents to check the locale')
+      return
+    }
+
+    if (
+      textContents[0].content.text.some(text => text.locale === currentLocale)
```
AlbertGom, line 41: I would extract this line into a function to make the code more readable and use it also in line 48.

AlbertGom (approved):

packages/botonic-plugin-flow-builder/src/api.ts
Iru89: I have it in mind for a next refactoring, but if this call fails, the one to get the flow will also fail and the bot will have no content to respond

PR #2931, opened 2024-11-05
botonic-react: fix mixed markdowns used in the same text #BLT-1177 

Iru89:
## Description

Refactor to simplify the code used to apply markdown in Whatsapp and Facebook 

## Context

A sentence with bold and italic at the same time is not rendered correctly. For example in the following sentence the bold was not applied on the word `agentes`
`_Recuerda que si es por una incidencia en la entrega o recepción, puedes contactar con nuestros __agentes__ para que te ayuden lo antes posible_`

Instead of normalizing directly with the result * or _ now use the words &%BOLD%& and &%ITALIC%&
I have added the &% symbols at the beginning and end to make it less probable that this is written in the bot message.

## Testing

Adds tests for several markdown use cases

### DIFF
--- a/packages/botonic-react/src/components/multichannel/whatsapp/markdown.ts
+++ b/packages/botonic-react/src/components/multichannel/whatsapp/markdown.ts
@@ -1,105 +1,66 @@
-const MARKDOWN_BOLD_OPTION_1 = '**'
-const MARKDOWN_BOLD_OPTION_2 = '__'
-const MARKDOWN_WHATSAPP_BOLD = '*'
+const REGEX_MARKDOWN_BOLD = /(\*\*|__)(.*?)\1/g
+const REGEX_MARKDOWN_ITALIC = /(\*|_)(.*?)\1/g
+const REGEX_MARKDOWN_LINK = /\[([^\]]+)\]\(([^)]+)\)/g
 
-const MARKDOWN_ITALIC_OPTION_1 = '*'
-const MARKDOWN_WHATSAPP_ITALIC = '_'
+const NORMALIZED_BOLD = '&%BOLD%&'
+const REGEX_NORMALIZED_BOLD = new RegExp(
+  `${NORMALIZED_BOLD}(.*?)${NORMALIZED_BOLD}`,
+  'g'
+)
+const NORMALIZED_ITALIC = '&%ITALIC%&'
+const REGEX_NORMALIZED_ITALIC = new RegExp(
+  `${NORMALIZED_ITALIC}(.*?)${NORMALIZED_ITALIC}`,
+  'g'
+)
 
-const MARKDOWN_BOLD_OR_ITALIC_REGEX = /(\*\*|__)(.*?)\1|(\*|_)(.*?)\3/g
+const WHATSAPP_BOLD = '*'
+const WHATSAPP_ITALIC = '_'
 
-const MARKDOWN_NORMALIZED_BOLD_ITALIC_OPEN = '**_'
-const MARKDOWN_NORMALIZED_BOLD_ITALIC_CLOSE = '_**'
-
-const MARKDOWN_BOLD_AND_ITALIC_OPTION1 = /(_\*\*)(.*?)(\*\*_)/g
-const MARKDOWN_BOLD_AND_ITALIC_OPTION2 = /(\*__)(.*?)(__\*)/g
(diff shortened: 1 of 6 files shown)


packages/botonic-react/src/components/multichannel/whatsapp/markdown.ts
```diff
@@ -1,105 +1,66 @@
...
+  // convert &%BOLD%&text&%BOLD%& to *text*
+  text = text.replace(
+    REGEX_NORMALIZED_BOLD,
+    `${WHATSAPP_BOLD}$1${WHATSAPP_BOLD}`
+  )
+  // convert &%ITALIC%&text&%ITALIC%& to _text_
+  text = text.replace(
+    REGEX_NORMALIZED_ITALIC,
+    `${WHATSAPP_ITALIC}$1${WHATSAPP_ITALIC}`
   )
```
vanbasten17, line 58: while I understand this is working correctly, I would do some changes to make it more readable for us in the future, just a possible one suggested by IA:
```
// Constants for Markdown and Normalized/WhatsApp symbols
const REGEX_MARKDOWN_BOLD = /\*\*(.+?)\*\*/g;
const REGEX_MARKDOWN_ITALIC = /_(.+?)_/g;
const NORMALIZED_BOLD = "&%BOLD%&";
const NORMALIZED_ITALIC = "&%ITALIC%&";

const REGEX_NORMALIZED_BOLD = /&%BOLD%&(.+?)&%BOLD%&/g;
const REGEX_NORMALIZED_ITALIC = /&%ITALIC%&(.+?)&%ITALIC%&/g;
const WHATSAPP_BOLD = "*";
const WHATSAPP_ITALIC = "_";

// Helper function for formatting conversion
function applyFormatting(
  text: string,
  regex: RegExp,
  openingSymbol: string,
  closingSymbol: string = openingSymbol
): string {
  return text.replace(regex, `${openingSymbol}$1${closingSymbol}`);
}

// Normalize Markdown to custom format
function normalizeMarkdown(text: string): string {
  text = applyFormatting(text, REGEX_MARKDOWN_BOLD, NORMALIZED_BOLD);
  text = applyFormatting(text, REGEX_MARKDOWN_ITALIC, NORMALIZED_ITALIC);
  return text;
}

// Convert custom format to WhatsApp format
function convertNormalizedToWhatsApp(text: string): string {
  text = applyFormatting(text, REGEX_NORMALIZED_BOLD, WHATSAPP_BOLD);
  text = applyFormatting(text, REGEX_NORMALIZED_ITALIC, WHATSAPP_ITALIC);
  return text;
}
```

packages/botonic-react/src/components/multichannel/whatsapp/markdown.ts
```diff
@@ -1,105 +1,66 @@
...
 }
 
-const REGEX_MARKDOWN_LINK = /\[([^\]]+)\]\(([^)]+)\)/g
-
-function replaceMarkdownLinks(markdown: string) {
-  return markdown.replace(REGEX_MARKDOWN_LINK, '$1: $2')
+function replaceMarkdownLinks(text: string) {
+  // $1 = textUrl, $2 = linkUrl
+  return text.replace(REGEX_MARKDOWN_LINK, '$1: $2')
 }
```
vanbasten17, line 66: I have looked about captured groups in regexs and you should be able to do this (it can be also applied in the other PR)
```
const REGEX_MARKDOWN_LINK = /\[(?<textUrl>[^\]]+)\]\((?<linkUrl>[^)]+)\)/g;

function replaceMarkdownLinks(text: string) {
  return text.replace(REGEX_MARKDOWN_LINK, '$<textUrl>: $<linkUrl>');
}
```

vanbasten17 (approved):

packages/botonic-react/src/components/multichannel/whatsapp/markdown.ts
Iru89: Cannot be used with the configuration we have for the CommonJS build

packages/botonic-react/src/components/multichannel/whatsapp/markdown.ts
Iru89: It cannot be as you propose because in the applyFormatting function it would need another input parameter for the $1 that goes between the openingSymbol and the closingSymbol.

PR #2932, opened 2024-11-06
botonic-core: export interface HubtypeHandoffParams

Iru89:
## Description

export interface HubtypeHandoffParams


AlbertGom (approved):
:100:

PR #2934, opened 2024-11-08
dx-bundler-rspack: create a dx package to build with rspack #BLT-1197 

Iru89:
## Description

Create a new dx package that has the dependencies to build a project using as bundler rspack
It also contains a file rspack.config.ts. As explained in the README you have to copy this file to the root of the botonic project.

## Context

In the README you can also see the scripts to use rspack to build the project locally or to make a builder for production.

### DIFF
--- a/packages/botonic-dx-bundler-rspack/README.md
+++ b/packages/botonic-dx-bundler-rspack/README.md
@@ -4,6 +4,10 @@
 
 This package simplifies the configuration of a rspack bundler to build a Botonic project.
 
+## IMPORTANT
+
+- Webviews imported inside the webviews/index.ts must be classes or functions but cannot be arrow functions.
+
 ## Setup
 
 - Install this package.
--- a/packages/botonic-dx-bundler-rspack/README.md
+++ b/packages/botonic-dx-bundler-rspack/README.md
@@ -1,0 +1,27 @@
+# Botonic bundler Rspack
+
+## What Does This Package Do?
+
+This package simplifies the configuration of a rspack bundler to build a Botonic project.
+
+## Setup
+
+- Install this package.
+
+```
+npm install -D @botonic/dx-bundler-rspack
+```
+
+- Copy `baseline/rspack.config.js` file to the root of your project.
+
+- Add this script to your package.json to build for production and start for development in local.
+
+```
+...
+"scipts": {
+    ...
+    "build_production": "ENVIRONMENT=production NODE_ENV=production rspack build --env target=all --mode=production",
+    "start": "ENVIRONMENT=local NODE_ENV=development rspack serve --env target=dev --mode=development",
+    ...
+}
+```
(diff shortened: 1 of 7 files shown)


asastre (approved):

packages/botonic-dx-bundler-rspack/README.md
```diff
@@ -0,0 +1,31 @@
+# Botonic bundler Rspack
+
+## What Does This Package Do?
+
+This package simplifies the configuration of a rspack bundler to build a Botonic project.
+
+## IMPORTANT
+
+- Webviews imported inside the webviews/index.ts must be classes or functions but cannot be arrow functions.
```
vanbasten17, line 9: I would put an example and explain the why briefly.
For example:
Do:
```
export function MyWebview(){
...
return
}
```
instead of:
```
export const MyWebview = () => {
...
return
}
```

This is because the bundler does not preserve correctly Webview names on transpilation process.

packages/botonic-dx-bundler-rspack/baseline/rspack.config.ts
```diff
@@ -0,0 +1,380 @@
...
+}
+
+const CONFIG_ENVIRONMENTS = ['production', 'staging', 'local']
+
+const resolveConfig = {
+  extensions: ['.*', '.js', '.jsx', '.ts', '.tsx'],
+  alias: {
+    BotonicProject: ROOT_PATH,
+    react: path.resolve('./node_modules/react'),
+    // 'styled-components': path.resolve('./node_modules/styled-components'),
```
vanbasten17, line 66: remove this line? otherwise, if this is left intentionally add a comment next to specify why

packages/botonic-dx-bundler-rspack/baseline/rspack.config.ts
```diff
@@ -0,0 +1,380 @@
...
+    '@botonic/react': BOTONIC_PATH,
+  },
+}
+
+const isDev = String(process.env.ENVIRONMENT) === 'local'
+
+const typescriptLoaderConfig =
+  // exclude: /node_modules[/\\](?!(@botonic\/(core|react))[/\\])/,
```
vanbasten17, line 74: same here

packages/botonic-dx-bundler-rspack/baseline/rspack.config.ts
```diff
@@ -0,0 +1,380 @@
...
+      exclude: [/node_modules/],
+      use: {
+        loader: 'builtin:swc-loader',
+        options: {
+          jsc: {
+            parser: {
+              syntax: 'typescript',
+              tsx: true,
+            },
+            keepClassNames: true,
```
vanbasten17, line 87: Just if you want in a last try, or later improvement, here it talks abouts specifying mangle with keep keepFnNames, which I think is what we want. I don't remember if we already tried it
https://swc.rs/docs/configuration/minification#jscminifymangle

packages/botonic-dx-bundler-rspack/baseline/rspack.config.ts
```diff
@@ -0,0 +1,380 @@
...
+const minimizerPlugin = new rspack.SwcJsMinimizerRspackPlugin({
+  minimizerOptions: {
+    minify: true,
+    compress: {
+      keep_classnames: true,
+      keep_fargs: false,
+      keep_fnames: true,
+    },
+    mangle: {
+      keep_fnames: true,
```
vanbasten17, line 218: ok, i see we have already put it

PR #2936, opened 2024-11-12
botonic-react: pass Button code to typescript

Iru89:
## Description

Review commit by commit. 
In the first commit only the file extension changes, in the following ones you can see the applied changes.

## Context

Add the ClientInput because the input that is sent from the client (webchat just like whatsapp, telegram etc) it is not necessary to put the message_id or the bot_interaction_id. These values are set by the backend 

Remove the bottom prop from the Button as it is not used anywhere and the StyledButton used top and bootom but the top did not exist. The button styles can be edited with a css stylesheet or by passing the style in the Webchat theme.button.style.


vanbasten17 (approved):

packages/botonic-react/src/components/button.tsx
```diff
@@ -187,43 +197,62 @@ export const Button = props => {
...
         </button>
       )
-    } else if (props.path) {
+    }
+
+    if (props.path) {
       const payload = `__PATH_PAYLOAD__${props.path}`
       return (
+        // @ts-ignore
+        // eslint-disable-next-line react/no-unknown-property
```
alexgimenez02, line 214: Nit: since you are adding this rule disable in multiple lines, you could add it on top of the file as `/* eslint-disable react/no-unknown-property */`

alexgimenez02 (approved):

packages/botonic-react/src/components/button.tsx
Iru89: Yes, I know, but I only want it for these lines

PR #2938, opened 2024-11-14
botonic-react: fix cover component in botonic react #BLT-1223

Iru89:
## Description

Create a separate file for the cover component (using typescript and reducing the size of the webchat file).
Add the toggleCoverComponent function in the WebchatContext to avoid having to pass the function using props

## Context

Removing the scroll in a previous PR was not rendering the cover component correctly

Use a conditional render to render the cover component or the webchat. It is not necessary to render the messages under the cover component.

### DIFF
--- a/packages/botonic-react/src/webchat/cover-component/index.tsx
+++ b/packages/botonic-react/src/webchat/cover-component/index.tsx
@@ -23,11 +23,9 @@ export const CoverComponent = ({ component, componentProps }: Props) => {
     toggleCoverComponent(false)
   }
 
-  if (!Cover) {
+  if (!Cover || !webchatState.isCoverComponentOpen) {
     return null
   }
 
-  return Cover && webchatState.isCoverComponentOpen ? (
-    <Cover closeComponent={closeCoverComponent} {...coverComponentProps} />
-  ) : null
+  return <Cover closeComponent={closeCoverComponent} {...coverComponentProps} />
 }
--- a/packages/botonic-react/src/webchat/cover-component/index.tsx
+++ b/packages/botonic-react/src/webchat/cover-component/index.tsx
@@ -1,0 +1,33 @@
+import React, { useContext } from 'react'
+
+import { WEBCHAT } from '../../constants'
+import { WebchatContext } from '../../contexts'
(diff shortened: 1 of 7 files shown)


packages/botonic-react/src/webchat/cover-component/index.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useContext } from 'react'
+
+import { WEBCHAT } from '../../constants'
+import { WebchatContext } from '../../contexts'
+
+interface Props {
```
vanbasten17, line 6: rename to `CoverComponentProps`

packages/botonic-react/src/webchat/cover-component/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  const closeCoverComponent = () => {
+    toggleCoverComponent(false)
+  }
+
+  if (!Cover) {
+    return null
+  }
+
+  return Cover && webchatState.isCoverComponentOpen ? (
```
vanbasten17, line 30: no need to check for `Cover` again if already checked above.

just return `webchatState.isCoverComponentOpen && <Cover ...`
and before `if (!Cover || !webchatState.isCoverComponentOpen`

vanbasten17 (approved):
Nice it is a starting point for having this in TS, I think we can still improve it in next iterations

packages/botonic-react/src/webchat/cover-component/index.tsx
Iru89: CoverComponentProps already exists elsewhere in the code

packages/botonic-react/src/webchat/cover-component/index.tsx
Iru89: applied

packages/botonic-react/src/webchat/cover-component/index.tsx
vanbasten17: so we'll need to refactor this in the future, we should not have it twice. Let's do it in next iterations as we simplify the code

PR #2942, opened 2024-11-26
botonic-react: ensure typing is visible by scrolling when typing is active #BLT-1243

vanbasten17:
## Description
Review commit by commit

* Refactor TypingIndicator to have a ref and a container. A container is needed to remove margin and add a padding because scrollIntoView with `end` option not scroll margin pixels 
* Fixed typing indicator to be visible when there is a typing effect.

* Add isInputFocus in webchatState to show UnreadMessagesBanner only when user is not focus in 

* Change the position to sticky so that it is always above the InputPanel (when the user types several lines of text the InputPanel can grow and the scroll down button should always be above it).

## Context
When sending a user input the webchat is scrolled to bottom, but as we have more messages, when we send an input and webchat scrolls to bottom and we have a typing effect, it is not visible because is under the scrolled area.
In this PR, we are making sure that when typing is set to true, the webchat is scrolled again to bottom to make it visible.

TypingIndicator

https://github.com/user-attachments/assets/4945c1c1-9816-4851-a37a-8bfbce3e1ef7

UnreadMessagesBanner

https://github.com/user-attachments/assets/1a1f725a-1cf0-4fd8-94f4-2b835c14b0da


Iru89 (approved):

MarcosCA (commented):
If someone is reviewing the beggining of the conversation and the agent starts typing you will lose the point of what were you checking right?

donihubtype (approved):

Iru89 (approved):

PR #2943, opened 2024-11-28
 Remove strings that refer components #BLT-1258

Iru89:
## Description

@botonic/react
Remove string that refer to @botonic/react components as Text, Button, MultichannelText etc 

@botonic/dx-bundler-rspack
Remove comment
Update README

### DIFF
--- a/packages/botonic-react/src/components/multichannel/multichannel.jsx
+++ b/packages/botonic-react/src/components/multichannel/multichannel.jsx
@@ -1,16 +1,18 @@
 import { isFacebook, isWhatsapp } from '@botonic/core'
 import React, { useContext } from 'react'
 
-import { COMPONENT_TYPE } from '../../constants'
 import { RequestContext } from '../../contexts'
 import { deepMapWithIndex } from '../../util/react'
+import { Button } from '../button'
+import { Carousel } from '../carousel'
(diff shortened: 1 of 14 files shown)


packages/botonic-react/src/components/multichannel/multichannel.jsx
```diff
@@ -22,7 +24,7 @@ export const Multichannel = props => {
   }
   if (isFacebook(requestContext.session)) {
     const newChildren = deepMapWithIndex(props.children, child => {
-      if (child && child.type && child.type.name === COMPONENT_TYPE.TEXT) {
+      if (child && child.type && child.type.name === Text.name) {
```
asastre, line 27: Refactor: You could change this (and all the other conditions below) to:
```suggestion
      if (isText(child)) {
```
This `isText` function doesn't exist now, but should be the same as `isPic`, `isButton`, etc.

asastre (approved):

alexgimenez02 (approved):

AinaVendrell (approved):

packages/botonic-react/src/components/multichannel/multichannel.jsx
Iru89: applied

PR #2947, opened 2024-12-16
botonic-react: Webchat component with typescript

Iru89:
## Description

Use typescript in the webchat file, there are still not all the correct types, there are some any and some @ts-ignore that will have to be checked in the future.

Check commit by commit, the first commit only changes the extension of the webchat file from jsx to tsx so in the following commits you can see the changes applied.

In the following PRs more components and files are passed to typescript


packages/botonic-react/src/index-types.ts
```diff
@@ -51,32 +54,91 @@ export interface RequestContextInterface extends ActionRequest {
...
+  closeWebchat: () => void
+  toggleWebchat: () => void
+  openCoverComponent: () => void
+  closeCoverComponent: () => void
+  toggleCoverComponent: () => void
+  renderCustomComponent: (customComponent: any) => void
+  unmountCustomComponent: () => void
+  isOnline: () => boolean
+  setOnline: (online: boolean) => void
+  getMessages: () => { id: string; ack: number; unsentInput: CoreInput }[] // TODO: define MessagesJSON
```
vanbasten17, line 76: do still need to be defined?

packages/botonic-react/src/index-types.ts
```diff
@@ -169,9 +239,9 @@ export interface WebchatContextProps {
   updateReplies: (replies: boolean) => void
   updateUser: (user: Partial<CoreSessionUser>) => void
   updateWebchatDevSettings: (settings: WebchatSettingsProps) => void
+  trackEvent?: TrackEventFunction
   webchatState: WebchatState
-  trackEvent: TrackEventFunction
-  webchatRef: React.MutableRefObject<HTMLDivElement | null>
+  webchatRef: React.MutableRefObject<HTMLDivElement | null> // Rename this to webchatContainerRef
```
vanbasten17, line 244: yes, i would rename it also

packages/botonic-react/src/webchat-app.tsx
```diff
@@ -248,21 +187,35 @@ export class WebchatApp {
     this.onClose && this.onClose(this, ...args)
   }
 
-  async onUserInput({ user, input }) {
+  async onUserInput({ user, input }: OnUserInputArgs): Promise<void> {
+    if (!user) return
```
vanbasten17, line 191: if `Botonic.addUserText()` is not called with user and the function goes through onUserInput, maybe when calling it it won't be executed. Can you review it?

packages/botonic-react/src/webchat-app.tsx
```diff
@@ -248,21 +187,35 @@ export class WebchatApp {
...
     this.onMessage &&
       this.onMessage(this, {
         ...input,
         sentBy: SENDERS.user,
         isUnread: false,
-      })
-    return this.hubtypeService.postMessage(user, input)
+      } as unknown as WebchatMessage)
+    this.hubtypeService.postMessage(user, input)
+    return
```
vanbasten17, line 200: optional: if it is returning a void promise, I would leave it as it was

vanbasten17 (approved):

packages/botonic-react/src/webchat-app.tsx
Iru89: Botonic.addUserText() -> Botonic.addUserMessage() -> webchatRef.current?.addUserMessage(message) -> sendInput() -> sendUserInput() -> onUserInput()

packages/botonic-react/src/index-types.ts
Iru89: In a futur PR I will try to create the right type for the messages.

packages/botonic-react/src/index-types.ts
Iru89: I will apply it to a future PR

packages/botonic-react/src/webchat-app.tsx
Iru89: It is the same and as it was before, the linter was complaining

packages/botonic-react/src/contexts.tsx
```diff
@@ -27,23 +27,20 @@ export interface CloseWebviewOptions {
...
   closeWebview: (options?: CloseWebviewOptions) => Promise<void>
-  getString: (stringId: string) => string
+  getString?: (stringId: string) => string
   params: Record<string, any>
-  session: CoreSession
+  session: Partial<CoreSession>
 }>({
   closeWebview: async () => undefined,
-  getString: () => '',
+  getString: undefined,
```
AinaVendrell, line 35: what is this for?

AinaVendrell (approved):

PR #2951, opened 2024-12-19
chore: update blank-typescript example with rspack

vanbasten17:
## Description
Updating blank-typescript to use rspack


examples/blank-typescript/package.json
```diff
@@ -12,6 +12,7 @@
     "@botonic/react": "^0.30.0"
   },
   "devDependencies": {
-    "@botonic/dx": "^0.30.0"
+    "@botonic/dx": "^0.30.0",
+    "@botonic/dx-bundler-rspack": "^0.30.0-alpha.0"
```
Iru89, line 16: we have released version 0.30.0-alpha.4 but by default it installs alpha.0 because it is marked as latest

Iru89 (approved):

PR #2952, opened 2024-12-19
botonic-react: WebchatReplies and Reply with typescript

Iru89:
## Description

Using typescript in Reply and WebchatReplies components

In the WebchatReplies component I have to put a @ts-ignore because replies is of type Reply[] but typescript does not detect that the component is already “rendered with the props” (TODO investigate in the future how to pass the props and render Reply components at this point).


packages/botonic-react/src/webchat/chat-area/index.tsx
```diff
@@ -27,9 +27,7 @@ export const ChatArea = () => {
       height={chatAreaHeight}
     >
       <WebchatMessageList />
-      {replies && Object.keys(replies).length > 0 && (
-        <WebchatReplies replies={replies} />
-      )}
+      {replies && replies.length > 0 && <WebchatReplies />}
```
AinaVendrell, line 30:
```suggestion
      {replies?.length > 0 && <WebchatReplies />}
```

AinaVendrell (approved):

PR #2958, opened 2025-01-09
Core: added types for whatsapp order

AlbertGom:
## Description
Added types for [whatsapp order](https://developers.facebook.com/docs/whatsapp/cloud-api/guides/sell-products-and-services/receive-responses) input type.

## Context
This input will be received from the backend when the user presses the button "Make order".
Captura de pantalla de 2025-01-14 12-00-44

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21317,6 +21317,50 @@
         "node": ">=0.10.0"
       }
     },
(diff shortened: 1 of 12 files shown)


vanbasten17 (approved):
Approving but just in case you keep consistency with camelCase

PR #2967, opened 2025-01-20
botonic-react: create webchat theme folder

Iru89:
## Description

This PR does not change any logic
- Create folder theme with a types.ts file. This file contains ThemeProps without CustomProp, EnableProp, ImageProp and StyleProp
- Update all imports
- Add TODO comments for a future refactor removing context.theme and use only context.webchatState.theme


packages/botonic-react/src/components/custom-message.tsx
```diff
@@ -6,7 +6,7 @@ import { SENDERS } from '../index-types'
 import { createErrorBoundary } from '../util/error-boundary'
 import { warnDeprecatedProps } from '../util/logs'
 import { mapObjectNonBooleanValues } from '../util/react'
-import { CustomMessageType } from './index-types'
+import { CustomMessageType } from '../webchat/theme/types'
```
vanbasten17, line 9: I think that this type should not go within webchat/theme/types but in another file. But ok for this moment

vanbasten17 (approved):

packages/botonic-react/src/components/custom-message.tsx
Iru89: It is only used in the Theme that's why I brought it to this file but I agree that it doesn't make much sense. I'll move it to src/components/index-types and we'll see where it needs to be later.

PR #2972, opened 2025-01-27
tech-feat: adapt botonic code to use new flow builder backend endpoints

vanbasten17:
## Description
* Update bot flow endpoints to point to new hubtype-backend endpoints

## Context
* Part of process of migrating flow builder to backend

## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -31,14 +31,17 @@ export class FlowBuilderApi {
...
     if (process.env.NODE_ENV === ProcessEnvNodeEnvs.DEVELOPMENT) {
       await newApi.updateSessionWithUserInfo(options.accessToken)
     }
-
+    const updatedRequest = newApi.request
+    newApi.flowUrl = options.flowUrl.replace(
+      '{bot_id}',
+      updatedRequest.session.bot.id
+    )
+    newApi.flow = options.flow ?? (await newApi.getFlow(options.accessToken))
```
vanbasten17, line 44: to make it work correctly in local, we should retrieve botId from userInfo, not from incoming session.bot.user.id, because it will be the default for webchat dev ('00000')
This should be refactored at some point

vanbasten17: added a TODO for refactoring

Iru89 (approved):

PR #2975, opened 2025-01-31
feat: subscribe to initial queue position in botonic

vanbasten17:
## Description
* Allow subscribe to new helpdesk event `Initial Queue Position`.

// routes
```
export const routes = (request) => {
  console.log("REQUEST", request);
  return [
    {
      path: "initial_queue_position",
      type: "case_event_initial_queue_position",
      action: InitialQueuePosition,
    },
    { path: "agent", text: /^handoff$/i, action: TransferAgent },
    { path: "thanks-for-contacting", action: Thanks },
  ];
};
```

// action

```
  static async botonicInit(request) {
    const input = request.input;
    console.log("INPUT", input);
    return {
      input,
    };
  }
  render() {
    return (
      <Text>
        QUEUE POSITION: {"\n"}
        Current: {this.props.input.current_queue_position} {"\n"}
        Total waiting: {this.props.input.total_queue_waiting_cases_number}
      </Text>
    );
  }
}

```
## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -83,6 +84,10 @@ async function getContents(
     return await getContentsByFirstInteraction(context)
   }
 
+  if (request.input.type === INPUT.EVENT_INITIAL_QUEUE_POSITION) {
+    // TODO: Think how we will get this content
+  }
```
AinaVendrell, line 89: are we not gonna implement this now?

vanbasten17: yes, I'm leaving this after I can try with the whole integration backend/fbuilder app and this

AinaVendrell (approved):

PR #2976, opened 2025-01-31
botonic-react: create a WebchatWebview component to render a webview inside the webchat

Iru89:
## Description

This PR does not change any logic.
Create a component to separate the logic of rendering the webview inside the webchat.

### DIFF
--- a/packages/botonic-react/src/webchat/context/index.tsx
+++ b/packages/botonic-react/src/webchat/context/index.tsx
@@ -13,6 +13,9 @@ export const WebchatContext = createContext<WebchatContextProps>({
   getThemeProperty: () => {
     return
   }, // used to retrieve a specific property of the theme defined by the developer in his 'webchat/index.js'
+  closeWebview: async () => {
+    return
+  },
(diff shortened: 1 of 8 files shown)


vanbasten17 (approved):

packages/botonic-react/src/webchat/webview/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  useEffect(() => {
+    window.addEventListener('message', close, false)
+
+    return () => window.removeEventListener('message', close, false)
+  }, [])
+
+  const isProduction = typeof Webview === 'string'
```
vanbasten17, line 54: in my opinion, unless this is what happen internally this is not a "sufficient condition" to determine that this is a production environment. I think it would make sense to be named as:
`isUrlToWebview` or similar, but I would add a TODO above this line to refactor this in the future. Add a task in tech debt so we can rethink properly how we determine different envs in the whole application later.

vanbasten17 (approved):

PR #2984, opened 2025-02-24
Blt 1408 update knowledge base plugin to use new inference endpoint

Iru89:
## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/content-fields/flow-knowledge-base.tsx
+++ b/packages/botonic-plugin-flow-builder/src/content-fields/flow-knowledge-base.tsx
@@ -11,12 +11,18 @@ export class FlowKnowledgeBase extends ContentFieldsBase {
   public text: string = ''
   public inferenceId?: string
   public sourcesData: { id: string; name: string }[] = []
+  public instructions: string
+  public hasMemory: boolean
+  public memoryLength?: number
 
   static fromHubtypeCMS(component: HtKnowledgeBaseNode): FlowKnowledgeBase {
     const newKnowledgeBase = new FlowKnowledgeBase(component.id)
     newKnowledgeBase.code = component.code
     newKnowledgeBase.feedbackEnabled = component.content.feedback_enabled
     newKnowledgeBase.sourcesData = component.content.sources_data
+    newKnowledgeBase.instructions = component.content.instructions
+    newKnowledgeBase.hasMemory = component.content.has_memory || false
+    newKnowledgeBase.memoryLength = component.content.memory_length
 
     return newKnowledgeBase
   }
--- a/packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
+++ b/packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
@@ -1,6 +1,8 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import axios, { AxiosResponse } from 'axios'
 
+import { Chunk } from './types'
+
 const DEFAULT_TIMEOUT = 10000
 
 export class HubtypeApiService {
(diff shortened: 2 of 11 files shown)


packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
```diff
@@ -14,28 +16,33 @@ export class HubtypeApiService {
...
+    memoryLength: number
   ): Promise<
     AxiosResponse<{
       inference_id: string
-      question: string
-      answer: string
+      query: string
+      chunks: Chunk[]
       has_knowledge: boolean
+      // raw_answer: string
```
vanbasten17, line 29: remove commented

packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
```diff
@@ -14,28 +16,33 @@ export class HubtypeApiService {
...
   async inference(
     authToken: string,
-    question: string,
-    sources: string[]
+    sources: string[],
+    instructions: string,
+    messageId: string,
+    memoryLength: number
   ): Promise<
     AxiosResponse<{
```
vanbasten17, line 24: create an interface for `InferenceResponse` so it will be easier for us to maintain it

vanbasten17: maybe add it to `src/types`

vanbasten17 (approved):
Maybe for local development we can do right now something like:
```
function getFlowBuilderConfig(
  env: ENVIRONMENT
): BotonicPluginFlowBuilderOptions {
  if (process.ENV === 'development') return getKnowledgeBaseResponse: localDevFunction.
  return {what you defined}
  ```

File: packages/botonic-plugin-knowledge-bases/src/index.ts [hunk 3]
```diff
@@ -20,24 +20,27 @@ export default class BotonicPluginKnowledgeBases implements Plugin {
...
   async getInference(
     session: Session,
-    userInput: string,
-    sources: string[]
+    sources: string[],
+    instructions: string,
+    messageId: string,
+    memoryLength: number
   ): Promise<KnowledgeBaseResponse> {
     const authToken = isProd ? session._access_token : this.authToken
```
vanbasten17, line 29: I would add a safe check here (if process.env === 'development') return 
with a TODO for local version

packages/botonic-plugin-flow-builder/src/content-fields/flow-knowledge-base.tsx
```diff
@@ -11,12 +11,18 @@ export class FlowKnowledgeBase extends ContentFieldsBase {
...
+  public memoryLength?: number
 
   static fromHubtypeCMS(component: HtKnowledgeBaseNode): FlowKnowledgeBase {
     const newKnowledgeBase = new FlowKnowledgeBase(component.id)
     newKnowledgeBase.code = component.code
     newKnowledgeBase.feedbackEnabled = component.content.feedback_enabled
     newKnowledgeBase.sourcesData = component.content.sources_data
+    newKnowledgeBase.instructions = component.content.instructions
+    newKnowledgeBase.hasMemory = component.content.has_memory || false
+    newKnowledgeBase.memoryLength = component.content.memory_length
```
AinaVendrell, line 25: you should consider the case when `has_memory` is false and `memory_length` is defined. I think you are not using `hasMemory` in the rest of the code.

AinaVendrell (changes requested):

packages/botonic-plugin-knowledge-bases/src/hubtype-knowledge-api-service.ts
Iru89: But this is not the object that the plugin resolves since the attributes are in snake_case as the backend resolves them. In the types file there is the KnowledgeBaseResponse interface with the attributes in camelCase which is the object that the call is debugged when the plugin is used and the attributes that are used in the bot

packages/botonic-plugin-flow-builder/src/content-fields/flow-knowledge-base.tsx
```diff
@@ -4,19 +4,28 @@ import React from 'react'
 import { ContentFieldsBase } from './content-fields-base'
 import { HtKnowledgeBaseNode } from './hubtype-fields'
 
+export const DEFAULT_MEMORY_LENGTH = 1
+
```
AinaVendrell, line 8: maybe `DISABLED_MEMORY_LENGHT`? in flow builder we also have `DEFAULT_MEMORY_LENGTH` and it is set to 10 so this could be confusing

AinaVendrell (approved):

packages/botonic-plugin-knowledge-bases/src/index.ts
Iru89: I don't understand you, isProd is checking if `process.env === 'production'` which is the case when we want to use the token in session

elozano98 (approved):

PR #2993, opened 2025-03-20
feat: avoid apply twice markdownMeta function

Iru89:
## Description

Avoid apply twice markdownMeta when transform Text with one url button to WhatsappCTAURLBUtoon inside Multichannel

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25238,7 +25238,7 @@
     },
     "packages/botonic-react": {
       "name": "@botonic/react",
-      "version": "0.33.0",
+      "version": "0.33.1",
       "license": "MIT",
       "dependencies": {
         "@botonic/core": "^0.33.0",


packages/botonic-react/src/components/multichannel/multichannel-text.jsx
```diff
@@ -239,10 +239,14 @@ export const MultichannelText = props => {
     )
 
     if (postbackButtonElements.length === 0) {
+      // Use texts[0] instead of textElements[0] as a body to avoid apply the markdownMeta function twice.
+      // If the markdownMeta function is applied twice, bold is replaced by italics.
+      const body = texts[0]
```
vanbasten17, line 244: seeing that texts comes from getText() function previously, would it be safer to do something like `texts[0] ?? ''` ?

vanbasten17 (approved):

PR #2995, opened 2025-03-24
Refactor theme button style

Iru89:
## Description

Define ButtonStyle interface and  add button.style inside default-theme. This allow add styles in webchat/index.ts inside the bot code or edit this styles with a css file without use `!important`
Remove getThemeProperty used in Button component


packages/botonic-react/src/components/multichannel/multichannel-button.jsx
```diff
@@ -3,7 +3,7 @@ import React, { useContext } from 'react'
 
 import { RequestContext } from '../../contexts'
 import { truncateText } from '../../util'
-import { Button } from '../button'
+import { Button } from '../button/index'
```
vanbasten17, line 6: why we need to do import from /index? this should not be necessary

vanbasten17 (approved):
Add the entry in botonic/react changelog

AinaVendrell (approved):

packages/botonic-react/src/webchat/theme/default-theme.ts
```diff
@@ -44,6 +44,31 @@ export const defaultTheme: WebchatTheme = {
...
+      outline: '0',
+      border: ` 1px solid ${COLORS.SEASHELL_WHITE}`,
+      borderTopLeftRadius: '0px',
+      borderTopRightRadius: '0px',
+      borderBottomLeftRadius: '0px',
+      borderBottomRightRadius: '0px',
+      padding: '12px 32px',
+      overflow: 'hidden',
+      whiteSpace: 'pre-line',
+    },
```
asastre, line 67: Observation: From the original styles, these ones are missing:
- display: flex
- text-align, align-content, justify-content: center
- gap: 6px
- cursor: pointer

And these ones have been added:
- height: 'auto'
- whiteSpace: 'pre-line'

These changes may affect actual bots, why not keeping the exact same styles as before? Also, I think that the border-radius properties can be unified into `borderRadius: '0px'`, right?

packages/botonic-react/src/components/multichannel/multichannel-button.jsx
asastre: I agree, I think there's no need to add the`/index` in the import path.

packages/botonic-react/src/webchat/theme/default-theme.ts
Iru89: the first attributes (cursor, display text-align and justify-content) I think they are values that will always have the same value but you can add them in case you want to change them someday.

height: auto is the default value if you don't define it.

I will remove the whiteSpace

the borderRadius if we leave only one it will be the 4 equal borders but it seems to me well

Iru89: applied

packages/botonic-react/src/components/multichannel/multichannel-button.jsx
Iru89: removed

packages/botonic-react/src/webchat/theme/default-theme.ts
asastre: For the borderRadius prop, you can always indicate all 4 borders in the same line: `borderRadius: "0px 1px 2px 3px"`

PR #2998, opened 2025-03-31
feat(core & plugin-flow-builder): add nluResult in input

Iru89:
## Description

- add nluResult botonic-core in request.input
- set and unset nluResult in plugin-flow-builder

## Testing

Update keyword and smart-intents test to check that request.input.nluResult is set in after pre function and unset after post function

### DIFF
--- a/packages/botonic-core/src/models/legacy-types.ts
+++ b/packages/botonic-core/src/models/legacy-types.ts
@@ -174,9 +174,9 @@ export interface Input extends Partial<NluResult> {
   bot_interaction_id: string
   catalog_id?: string
   product_items?: ProductItem[]
-  nluResult?: {
+  nluResolution?: {
     type: NluType
-    match: string
+    matchedValue: string
   }
 }
 
--- a/packages/botonic-core/src/models/legacy-types.ts
+++ b/packages/botonic-core/src/models/legacy-types.ts
@@ -104,7 +104,7 @@ export interface ResolvedPlugin extends Plugin {
 }
 export type ResolvedPlugins = Record<string, ResolvedPlugin>
 
-type InputType =
+export type InputType =
   | INPUT.AUDIO
   | INPUT.BUTTON_MESSAGE
   | INPUT.CAROUSEL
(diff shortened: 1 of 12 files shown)


packages/botonic-core/src/models/legacy-types.ts
```diff
@@ -169,6 +174,10 @@ export interface Input extends Partial<NluResult> {
   bot_interaction_id: string
   catalog_id?: string
   product_items?: ProductItem[]
+  nluResult?: {
```
vanbasten17, line 177: `nluResolution` better?

packages/botonic-core/src/models/legacy-types.ts
```diff
@@ -169,6 +174,10 @@ export interface Input extends Partial<NluResult> {
...
+    type: NluType
+    match: string
```
vanbasten17, line 179: `value` or `matchedValue`

vanbasten17 (approved):

packages/botonic-core/src/models/legacy-types.ts
Iru89: matchedValue

packages/botonic-plugin-flow-builder/tests/keyword.test.ts
```diff
@@ -12,14 +12,25 @@ describe('Check the contents returned by the plugin using keywords', () => {
...
+          flowBuilderOptions: { flow: basicFlow },
+          requestArgs: {
+            input: { data: inputData, type: INPUT.TEXT },
+          },
+        })
 
       expect((contents[0] as FlowText).text).toBe('Welcome message')
+      expect(request.input.nluResolution).toEqual({
+        type: 'keyword',
+        matchedValue: inputData,
```
AlbertGom, line 26: The `matchedValue` should be the keyword not the input of the user, no?

vanbasten17: `matchingValue` maybe makes more sense

Iru89: the matchedValue will be the keyword not the entire user input. But in this test the user input is only the keyword

PR #3003, opened 2025-04-05
botonic-react: ReactBot class with typescript

Iru89:
## Description

Pass ReactBot class to typescript
Add RequestCoreContext type. This is the request that the CoreBot passes to the actions and is used from the RequestContext.
Create requestCoreContext with the values and functions that need RequestContext inside actions
Not only use this requestCoreContext for actions, use it for pre and post functions od the plugins and also in routes


packages/botonic-core/src/core-bot.ts
```diff
@@ -91,194 +91,160 @@ export class CoreBot {
     session,
     lastRoutePath,
   }: BotRequest): Promise<BotResponse> {
-    const output = await this.runInput({ input, session, lastRoutePath })
+    const requestCoreContext: RequestCoreContext = {
```
vanbasten17, line 94: I think this name is not accurate, maybe simply `BotContext`?

packages/botonic-core/src/core-bot.ts
```diff
@@ -91,194 +91,160 @@ export class CoreBot {
...
     if (this.plugins) {
-      await runPlugins(
-        this.plugins,
-        'pre',
-        input,
-        session,
-        lastRoutePath,
-        undefined
-      )
+      await runPlugins({ requestCoreContext, mode: 'pre' })
```
vanbasten17, line 157: we could have 'pre' and 'post' typed as `type PluginMode = 'pre' | 'post'`

packages/botonic-core/src/models/legacy-types.ts
```diff
@@ -100,7 +100,7 @@ export interface PluginConfig<T> {
 export interface ResolvedPlugin extends Plugin {
   id: string
   name: string
-  config: any
+  config?: any
```
vanbasten17, line 103: do we have any plugin which does not require config?

packages/botonic-core/src/plugins.ts
```diff
@@ -26,27 +31,22 @@ export function loadPlugins(plugins: PluginConfig<any>[]): ResolvedPlugins {
...
+  response = null,
+}: RunPluginArgs): Promise<void> {
+  const plugins = requestCoreContext.plugins
   for (const key in plugins) {
-    const p = await plugins[key]
+    const plugin: Plugin = await plugins[key]
     try {
-      if (mode === 'pre' && p.pre) {
-        await p.pre({ input, session, lastRoutePath, plugins })
+      if (mode === 'pre' && plugin.pre) {
```
vanbasten17, line 43: maybe we have to ensure pre is a callable function, otherwise, we should throw some kind of error like: plugin misconfigured

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -248,11 +248,17 @@ export class FlowBuilderApi {
     if (this.flow.locales.find(flowLocale => flowLocale === locale)) {
       return locale
     }
-    const language = locale.split('-')[0]
-    if (this.flow.locales.find(flowLocale => flowLocale === language)) {
+
+    const language = locale?.split('-')[0]
```
vanbasten17, line 252: you could also do:
`const [language, _] = locale?.split('-')` and you will already have it

packages/botonic-plugin-flow-builder/src/api.ts
```diff
@@ -248,11 +248,17 @@ export class FlowBuilderApi {
...
+
+    if (
+      language &&
+      this.flow.locales.find(flowLocale => flowLocale === language)
+    ) {
       console.log(`locale: ${locale} has been resolved as ${language}`)
       return language
     }
```
vanbasten17, line 260: I think we can do this more robust in terms of resolving the locale, also I would find a way to make this more readable and not "comparing" `flowLocale` to `language`

packages/botonic-plugin-flow-builder/src/helpers.ts
```diff
@@ -1,19 +1,25 @@
...
-  if (!flowBuilderPlugin)
-    throw new Error(
-      `You must include '@botonic/plugin-flow-builder' in your plugins file.`
-    )
+export function getFlowBuilderPlugin(
+  plugins?: ResolvedPlugins
+): BotonicPluginFlowBuilder {
+  const ERROR_MESSAGE = `You must include '@botonic/plugin-flow-builder' in your plugins file.`
+
+  if (!plugins) {
```
vanbasten17, line 12: not sure if needed

packages/botonic-plugin-flow-builder/src/helpers.ts
```diff
@@ -1,19 +1,25 @@
...
+    throw new Error(ERROR_MESSAGE)
+  }
+
+  const flowBuilderPlugin = Object.values(plugins).find(plugin =>
+    plugin.name.includes(FLOW_BUILDER_PLUGIN_NAME)
+  ) as unknown as BotonicPluginFlowBuilder
```
vanbasten17, line 18: future improvement: rethink how we resolve plugins

packages/botonic-react/src/react-bot.tsx
```diff
@@ -0,0 +1,50 @@
...
+      renderer: (args: RenderReactActionsArgs) => this.renderReactActions(args),
+      ...options,
+    })
+  }
+
+  async renderReactActions({ request, actions }: RenderReactActionsArgs) {
+    const renderedActions: JSX.Element[] = []
+
+    for (const Action of actions) {
+      if (Action) {
```
vanbasten17, line 36: not needed if

packages/botonic-core/src/core-bot.ts
Iru89: Yes but this is a type not an enum, if you try to pass a string that is neither `pre` nor `post` there is a type error. But you can not do PluginMode.pre

packages/botonic-plugin-flow-builder/src/api.ts
Iru89: Yes, in the next PRs I will add logic to use a defaultLocale and refactor this part.

packages/botonic-react/src/react-bot.tsx
Iru89: applied

Iru89: I have added the if again because without it it does not work fine...

packages/botonic-core/src/core-bot.ts
Iru89: applied

packages/botonic-core/src/core-bot.ts
```diff
@@ -91,194 +91,149 @@ export class CoreBot {
...
-        this.plugins,
-        'post',
-        input,
-        session,
-        lastRoutePath,
-        response
-      )
+      await runPlugins({
+        botContext,
+        mode: 'post',
```
vanbasten17, line 193: nit: it would be nice to have type/enum for this `PluginMode.Pre|Post`

packages/botonic-core/src/models/legacy-types.ts
```diff
@@ -252,11 +252,13 @@ export interface HubtypeCaseContactReason {
   project_id: string
 }
 
+export interface SessionBot {
```
vanbasten17, line 255: I'm wondering if this is more `InitialSessionBot` rather than `SessionBot` as later this is filled with more values

vanbasten17 (approved):

packages/botonic-core/src/core-bot.ts
Iru89: I'll apply it in another PR

PR #3004, opened 2025-04-08
plugin-flow-builder: nluResolution available in first interaction

Iru89:
## Description

In order to have available the detection of keywords, smart-intents and intents-babel in the first interaction, as long as the user input is of text type and there is no payload, a match with keywords, smart-intents or intents-babel will be searched. This information is stored in the request.inpu.nluResolution together with the payload pointing to the matched node.

In this way it is no longer necessary to look for the match within the action when it is the first interaction, since the necessary payload will be stored in the input.nluResolution.payload.

## Testing

Update tests

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/action/first-interaction.ts
+++ b/packages/botonic-plugin-flow-builder/src/action/first-interaction.ts
@@ -2,18 +2,15 @@ import { FlowBuilderApi } from '../api'
 import { MAIN_FLOW_NAME } from '../constants'
 import { FlowBotAction, FlowContent } from '../content-fields'
 import BotonicPluginFlowBuilder from '../index'
-import { getNodeByUserInput } from '../user-input'
 import { inputHasTextData } from '../utils'
 import { FlowBuilderContext } from './index'
 import { getContentsByKnowledgeBase } from './knowledge-bases'
 import { getContentsByPayload } from './payload'
 
-export async function getContentsByFirstInteraction({
-  cmsApi,
-  flowBuilderPlugin,
-  request,
-  resolvedLocale,
-}: FlowBuilderContext): Promise<FlowContent[]> {
+export async function getContentsByFirstInteraction(
+  context: FlowBuilderContext
+): Promise<FlowContent[]> {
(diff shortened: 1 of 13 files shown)


packages/botonic-plugin-flow-builder/src/user-input/intent.ts
```diff
@@ -15,6 +16,13 @@ export async function getIntentNodeByInput(
     await trackIntentEvent(request, intentNode)
 
     if (isIntentValid(intentNode, request, cmsApi)) {
+      const targetPayload = cmsApi.getPayload(intentNode.target)
+
+      request.input.nluResolution = {
+        type: NluType.Keyword,
```
AlbertGom, line 22: this should be `NluType.Intent`

AlbertGom (changes requested):

AlbertGom (approved):

packages/botonic-plugin-flow-builder/src/action/first-interaction.ts
```diff
@@ -43,39 +35,36 @@ async function getContentsByUserInput({
   request,
   resolvedLocale,
 }: FlowBuilderContext): Promise<FlowContent[]> {
-  const nodeByUserInput = await getNodeByUserInput(
-    cmsApi,
-    resolvedLocale,
-    request,
-    flowBuilderPlugin.smartIntentsConfig
-  )
+  const payloadByNlu = request.input.nluResolution?.payload
```
vanbasten17, line 38: payloadFromNluResolution ?

packages/botonic-plugin-flow-builder/tests/keyword.test.ts
```diff
@@ -25,10 +25,8 @@ describe('Check the contents returned by the plugin using keywords', () => {
         })
 
       expect((contents[0] as FlowText).text).toBe('Welcome message')
-      expect(request.input.nluResolution).toEqual({
-        type: 'keyword',
-        matchedValue: keyword,
```
vanbasten17, line 30: should we check also the `payload` you added in types?

PR #3005, opened 2025-04-09
botonic-react: add user locale and user country on init webchat #BLT-1562

Iru89:
## Description

1. Add fields country, locale and system_locale in session.
2. Add functions to set this new fields from RequestCoreContext to be able to set this fields in the bot lambda (plugins, routes and actions) and add the same functions in WebchatApp to be able to set this values from the browser using for example Botonic.setCountry
3. Webchat after execute onInit function check the session:
   - if session.user has the new values not update session
   -  if user.extra_data has a country and language use it to update the new user fields
   - if not has values in extra_data use browser API to get locale using `navigator.locale` and country using `Intl.DateTimeFormat().resolvedOptions().timeZone` and a JSON to map the country to a ISO 3166-1 alpha-2
4. When CoreBot.input recibes a session try to update the `session.user` fields (this is needed for a non webchat integrations)

## Testing

Add tests for CoreBot updateSession


packages/botonic-core/src/models/legacy-types.ts
```diff
@@ -245,7 +246,14 @@ export interface SessionUser {
   extra_data?: any
   imp_id?: string
   provider_id?: string
+  locale?: string
+  country?: string
+  system_locale?: string
+  // TODO: Rethink how declare a field to reuse after finish different flow actions
+  show_first_interaction?: boolean
+  show_content_after_flow?: string
```
vanbasten17, line 254: remove

Iru89: yes

PR #3006, opened 2025-04-15
botonic-core: infer type of plugins to bot context #BLT-1574

Iru89:
## Description

With the new `BotContext` the types in the plugins were not correct, because CoreBot received an array of plugins and initialized them but with the function `loadPlugins`, but this function added to the object an id, name and configuration.

Now it will be possible to infer a type for the plugins in `BotContext` so that when retrieving plugins from a `request` it will be typed correctly.

Now `CoreBot` receives the plugins initialized and not use the `loadPlugins` function

src/server/plugins.ts
```typescript
import BotonicPluginFlowBuilder from '@botonic/plugin-flow-builder'
import BotonicPluginHubtypeAnalytics from '@botonic/plugin-hubtype-analytics'
import BotonicPluginKnowledgeBases from '@botonic/plugin-knowledge-bases'

import { CONFIG } from './config'
import { getEnvironment } from './utils/env-utils'

const config = CONFIG[getEnvironment()]

export type BotPlugins = typeof plugins

export const plugins = {
  flowBuilder: new BotonicPluginFlowBuilder(config.flowBuilder),
  knowledgeBases: new BotonicPluginKnowledgeBases(config.knowledgeBases),
  hubtypeAnalytics: new BotonicPluginHubtypeAnalytics(),
}
```

Then when you have the request in routes, plugin pre or post function, and action botonicInit function you can type the request using the next type and have the request.plugin typed correctly

src/server/routes.ts
```typescript
export function routes(request: BotContext<BotPlugins>): Routes {
...
}
```


vanbasten17 (approved):
Review failing tests 🙏

Iru89:
> Review failing tests 🙏

I will fix it in future PRs, I need to apply the core changes in all plugins

PR #3007, opened 2025-04-22
botonic-core & botonic-react: BotContext with getUserLocale, geUserCountry and getSystemLocale functions 

Iru89:
## Description

Add functions in BotContext to getUserLocale, geUserCountry and getSystemLocale
Add the same functions in RequestContext to have this functions in botonic actions

## Context

In the same way that we have the functions to update the new session attributes from the BotContext we also have the functions to obtain these values

## Testing

Add new functions when check request as a BotContext
Fix tests that check request inside the response because request.response is only an array with the responses created by the actions


packages/botonic-core/tests/core-bot/core-bot.test.ts
```diff
@@ -209,29 +184,7 @@ describe('CoreBot', () => {
...
-          system_locale: LOCALE_EN,
-        },
-        is_first_interaction: false,
-      },
-      setUserLocale: expect.any(Function),
-      setUserCountry: expect.any(Function),
-      setSystemLocale: expect.any(Function),
-    })
-    expect(botResponse.response[0].actions).toEqual([null, 'Hi user!', null])
+    expect(botResponse.response).toEqual(['Hi user!'])
```
vanbasten17, line 187: I think that you should preserve as well `[null, 'Hi user!', null]` the null actions. Though they will not be rendered, should be there technically

packages/botonic-core/tests/core-bot/core-bot.test.ts
```diff
@@ -281,15 +234,9 @@ it('input returns two actions when first return a _botonic_action redirect', asy
     },
   })
   expect(botResponse.lastRoutePath).toEqual('after-rating-action')
-  expect(botResponse.response[0].actions).toEqual([
-    null,
+  expect(botResponse.response).toEqual([
```
vanbasten17, line 237: same, null actions should be preserved

packages/botonic-core/src/core-bot.ts
```diff
@@ -76,8 +76,9 @@ export class CoreBot {
           )
   }
 
+  // TODO: remove getString function?
   getString(id: string, session: Session): string {
-    return getString(this.locales, session.user.system_locale!, id)
+    return getString(this.locales, session.user.system_locale, id)
```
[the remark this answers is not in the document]
Iru89: I will discuss it with Delivery but I think we can eliminate this function

MarcosCA (approved):
a part from marc comments

packages/botonic-core/tests/core-bot/core-bot.test.ts
Iru89: applied

packages/botonic-core/tests/core-bot/core-bot.test.ts
Iru89: applied

vanbasten17 (approved):

PR #3011, opened 2025-04-24
core & react & plugin-flow-builder: remove get string function from request bot context #BLT-1591 

Iru89:
## Description

Remove getString function from BotContext

## Testing

Remove tests that check getString function


Iru89:
I have a doubt, should we also remove the locales attribute from the BotCore? This is related to the getString function.
cc: @marcrabat @asastre

vanbasten17 (approved):

vanbasten17:
i won't remove `__locale` at this moment as I it is still filled by the BotSessionContext. Let's see if we can get rid of it in next iterations once we have the "locale" api better defined

PR #3019, opened 2025-05-12
WIP: load initialState from localStorage

Iru89:
## Description

Load the `initialState` in the context using the `localStorage` data. 
Do not condition the call to the hook `useWebchat` inside the Webchat component. Always use the `prop.webchatHooks`. To achieve this I have added a component between the `WebchatApp` and the Webchat called `WebchatHooks` that uses the hook `useWebchat` and passes the `webchatHooks` as props.

## Testing

The pull request...

- has unit tests
- has integration tests


Iru89:
I'm closing the PR because it's very outdated and botonicV2 is being developed, where the webchat is completely new

PR #3020, opened 2025-05-13
chore: create gh action deploy botonic #PLA-2878

amerino84:
## Description

Bash script and gh action to publish a new version of botonic.
The gh action calls the bash script when needed

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/.github/workflows/botonic-publish.yml
+++ b/.github/workflows/botonic-publish.yml
@@ -11,7 +11,6 @@ on:
         options:
           - 'patch'
           - 'minor'
-          - 'major'
       packages:
         description: 'Select packages to deploy'
         required: true
--- a/.github/workflows/botonic-publish.yml
+++ b/.github/workflows/botonic-publish.yml
@@ -1,4 +1,4 @@
-name: Botonic Deploy
+name: Botonic Publish
 
 on:
   workflow_dispatch:
@@ -6,11 +6,12 @@ on:
       version_type:
         description: 'Type of version upgrade'
         required: true
-        default: 'minor'
+        default: 'patch'
         type: choice
         options:
-          - 'major'
+          - 'patch'
           - 'minor'
+          - 'major'
       packages:
         description: 'Select packages to deploy'
         required: true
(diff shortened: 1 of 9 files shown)


amerino84:
I have prepared the script and the github action for major versions upgrades too. Previously it was only prepared for patch and minor but with major and minor names. A bit confusing

.github/workflows/botonic-publish.yml
```diff
@@ -0,0 +1,372 @@
...
+    inputs:
+      version_type:
+        description: 'Type of version upgrade'
+        required: true
+        default: 'patch'
+        type: choice
+        options:
+          - 'patch'
+          - 'minor'
+          - 'major'
```
vanbasten17, line 14: I would remove 'major' (or leave it commented) at this moment as we are only publishing "minor" and patch versions so we don't mess publishing a 1.X.Y

scripts/ci/botonic-publish.sh
```diff
@@ -0,0 +1,189 @@
...
+find_highest_version() {
+  local highest_version="0.0.0"
+  
+  for pkg_dir in packages/*; do
+    if [ -d "$pkg_dir" ] && [ -f "$pkg_dir/package.json" ]; then
+      local pkg_version=$(cd $pkg_dir && node -p "require('./package.json').version")
+      
+      # Parse the version numbers
+      local highest_patch=$(echo $highest_version | cut -d. -f3)
+      local highest_major=$(echo $highest_version | cut -d. -f2)
```
vanbasten17, line 33: from previous comment, maybe we don't need to include the major logic, but it can be good in the future

vanbasten17 (approved):
💯 💯 💯 💯

Khaeshah (approved):

scripts/ci/botonic-publish.sh
amerino84: I think it is good that it is prepared for the future. It was a small change and makes sense I think

Iru89 (approved):

PR #3028, opened 2025-05-29
plugin-hubtype-analytics: hubtype analytics send new fields #BLT-1629

Iru89:
## Description

Add new `user_locale`, `user_country` and `system_locale` fields for events with format_version = 3.
Remove fields language and country used in format_version = 2

## Testing

Update all tests


vanbasten17 (approved):
Go! But I think we should briefly readapt backend code to make it simpler there

PR #3030, opened 2025-06-02
feat: add webchatSettingParams as an optional prop to use WebchatSett…

Iru89:
…ings component

## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -68,19 +75,30 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
...
-    const request = this.context as ActionRequest
+    const { contents, webchatSettingsParams } = this.props
+
     return (
       <Multichannel text={{ buttonsAsText: false }}>
-        {contents.map(content => content.toBotonic(content.id, request))}
+        <FlowBuilderAction
+          contents={contents}
+          webchatSettingsParams={webchatSettingsParams}
+        />
```
asastre: Nit: This could be reduce to:
```suggestion
        <FlowBuilderAction { ...this.props } />
```

asastre (approved):

packages/botonic-plugin-flow-builder/src/action/index.tsx
Iru89: applied

packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -68,19 +75,26 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
...
     const request = this.context as ActionRequest
-    return contents.map(content => content.toBotonic(content.id, request))
+
+    return (
+      <>
+        {(isWebchat(request.session) || isDev(request.session)) &&
+          !!webchatSettingsParams && (
+            <WebchatSettings {...webchatSettingsParams} />
+          )}
+        {contents.map(content => content.toBotonic(content.id, request))}
```
vanbasten17, line 87: i'm not really sure right now, but I think that WebchatSettings should be sent at the end 👀

Iru89: Yesterday we did some testing and it works fine

asastre: @vanbasten17 we're sending the `WebchatSettings` at the beginning to apply the changes in the webchat as soon as possible. If we send it after the contents, the changes will be applied after all contents are displayed in the webchat.

One thing we usually do is to disable the user input after a handoff, and if we send the `WebchatSettings` at the end, the user will be able to write something while the after handoff contents are being rendered, but we want the user input to be disabled as soon as possible after the case has been resolved.

PR #3031, opened 2025-06-04
feat: retrieve context for webviews from new external api endpoint for bot session context

vanbasten17:
## Description

WebviewApp to retrieve user_id and bot_id from new url generated by backend and using them to retrieve bot session via external api.

### DIFF
--- a/packages/botonic-react/src/webview-app.tsx
+++ b/packages/botonic-react/src/webview-app.tsx
@@ -1,5 +1,5 @@
 // @ts-nocheck
-import { params2queryString, PROVIDER } from '@botonic/core'
+import { params2queryString, PROVIDER, Session } from '@botonic/core'
 import axios from 'axios'
 import React from 'react'
 import { createRoot } from 'react-dom/client'
@@ -16,6 +16,11 @@ enum WebviewUrlParams {
 
 class App extends React.Component {
   private url: URL
+  private botId: string
+  private userId: string
+  private hubtypeApiUrl: string
+  private state: { session: Session; params: Record<string, string> }
+
   constructor(props) {
     super(props)
     this.url = new URL(window.location.href)
@@ -76,10 +81,9 @@ class App extends React.Component {
 
       const session = this.state.session
       try {
-        const baseUrl = session._hubtype_api || 'https://api.hubtype.com'
-        const url = `${baseUrl}/v1/bots/${session.bot.id}/send_postback/`
+        const url = `${this.hubtypeApiUrl}/v1/bots/${this.botId}/send_postback/`
         // eslint-disable-next-line @typescript-eslint/naming-convention
-        const data = { payload: payload, chat_id: session.user.id }
+        const data = { payload: payload, chat_id: this.userId }
         await axios.post(url, data)
       } catch (e) {
         console.log(e)
--- a/packages/botonic-react/src/webview-app.tsx
+++ b/packages/botonic-react/src/webview-app.tsx
@@ -8,18 +8,59 @@ import { BrowserRouter, Route } from 'react-router-dom'
 import { CloseWebviewOptions, WebviewRequestContext } from './contexts'
 
 class App extends React.Component {
+  private url: URL
   constructor(props) {
     super(props)
-    const url = new URL(window.location.href)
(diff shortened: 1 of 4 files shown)


packages/botonic-react/src/webview-app.tsx
```diff
@@ -8,18 +8,59 @@ import { BrowserRouter, Route } from 'react-router-dom'
...
+        this.getBotSessionContextFromUrl()
+      }
+    } catch (error) {
+      this.getBotSessionContextFromUrl()
+    }
+  }
+
+  async getBotSessionContextFromExternalApi(botId: string, chatId: string) {
+    // const baseUrl = session._hubtype_api || 'https://api.hubtype.com'
+    const baseUrl = 'https://api.hubtype.com' // TODO: How to retrieve here the api url to request?
```
Iru89: We have to think about how to solve this, otherwise in non-production environments it won't work

packages/botonic-react/src/webview-app.tsx
```diff
@@ -8,18 +8,59 @@ import { BrowserRouter, Route } from 'react-router-dom'
...
+    const url = `${baseUrl}/external/v2/conversational_apps/${botId}/users/${chatId}/context/`
+
+    const response = await axios.get(url)
+    const session = response.data
+
+    this.setState({
+      session,
+      // TODO: Are we currently using botonic webviews api for params?
```
Iru89: I think we have to keep the params since they are used in the WebviewRequestContext line(122)

packages/botonic-react/src/webview-app.tsx
```diff
@@ -8,18 +8,59 @@ import { BrowserRouter, Route } from 'react-router-dom'
...
+    }
+  }
+
+  async componentDidMount() {
+    try {
+      const botId = this.url.searchParams.get('bot_id')
+      const chatId = this.url.searchParams.get('chat_id')
+      if (botId && chatId) {
+        this.getBotSessionContextFromExternalApi(botId, chatId)
+      } else {
```
gbarba: I don't think we want the if/else in botonic because we know the back end has already [this](https://github.com/metis-ai/hubtype-backend/pull/2389/files) in place so we know we always receive the bo tid and user id

packages/botonic-react/src/webview-app.tsx
```diff
@@ -8,18 +8,59 @@ import { BrowserRouter, Route } from 'react-router-dom'
...
+    this.state = {
+      session: null,
+      params: {},
+    }
+  }
+
+  async componentDidMount() {
+    try {
+      const botId = this.url.searchParams.get('bot_id')
+      const chatId = this.url.searchParams.get('chat_id')
```
gbarba: this is the userId/user_id which is the naming the bots use

packages/botonic-react/src/webview-app.tsx
gbarba: this should be in the config of the bot (local/npm level), so get on "compilation time" (idea from an outsider of botonic, so maybe doesn't make sense).

or this is something we changed recently to try to avoid have this "env var/config" at compilatin time?
is it possible this is injected in some way (do a replace or add a file) when the bot (webview) is deployed?

Yep, probably worth do a meet to expose ideas and pros/cons

gbarba (changes requested):

packages/botonic-react/src/webview-app.tsx
vanbasten17: we have added a `hubtype_api_url` param to have this

packages/botonic-react/src/webview-app.tsx
vanbasten17: added 👌

packages/botonic-react/src/webview-app.tsx
vanbasten17: removed the old logic as this is how they are gonna work in the new version of botonic

packages/botonic-react/src/webview-app.tsx
vanbasten17: changed to get user_id instead of chat_id

Iru89 (approved):

gbarba (approved):

PR #3033, opened 2025-06-05
plugin-ai-agents:  use langGraph to manage agents and tools #BLT-1666 

Iru89:
## Description

This plugin is used to activate an AIAgent in the bot. It uses langGrah to create the model, agent and tools.
For now only Azure openAI models are supported. 
When deploying the bot it is necessary to have the environment variables: 
- AZURE_OPENAI_API_KEY
- AZURE_OPENAI_API_DEPLOYMENT_NAME
- AZURE_OPENAI_API_VERSION
- AZURE_OPENAI_API_BASE

The plugin adds by default some tools to stop the agent from continuing the conversation:
- handoffToHumanAgent
- outOfContext
- finishConversation

Updates the flow-builder plugin that receives the agent's response. Change the way the exit tool is handled so that the AIAgent does not respond.

### DIFF
--- a/packages/botonic-plugin-ai-agents/src/hubtype-client.ts
+++ b/packages/botonic-plugin-ai-agents/src/hubtype-client.ts
@@ -4,7 +4,7 @@ import axios from 'axios'
 import { HUBTYPE_API_URL } from './constants'
 import { AgenticInputMessage } from './types'
 
-export class HubtypeClient {
+export class HubtypeApiClient {
   private readonly authToken: string
 
   constructor(authToken: string) {
--- a/packages/botonic-plugin-ai-agents/src/index.ts
+++ b/packages/botonic-plugin-ai-agents/src/index.ts
@@ -1,7 +1,7 @@
 import { BotContext, Plugin } from '@botonic/core'
 
 import { AiAgentClient } from './ai-agent-client'
-import { HubtypeClient } from './hubtype-client'
+import { HubtypeApiClient } from './hubtype-client'
 import { MANDATORY_TOOLS } from './tools'
 import {
   AgenticOutputMessage,
@@ -33,7 +33,7 @@ export default class BotonicPluginAiAgents implements Plugin {
         throw new Error('Auth token is required')
       }
 
-      const hubtypeClient = new HubtypeClient(authToken)
+      const hubtypeClient = new HubtypeApiClient(authToken)
       const messages = isProd
         ? await hubtypeClient.getMessages(request, 10)
         : [{ role: 'user', content: request.input.data } as UserMessage]


packages/botonic-plugin-ai-agents/src/hubtype-client.ts
```diff
@@ -0,0 +1,43 @@
+import { BotContext } from '@botonic/core'
+import axios from 'axios'
+
+import { HUBTYPE_API_URL } from './constants'
+import { AgenticInputMessage } from './types'
+
+export class HubtypeClient {
```
vanbasten17: i would name it `HubtypeAPIClient`

vanbasten17 (approved):
Looking promising so far! 💯

packages/botonic-plugin-ai-agents/src/hubtype-client.ts
Iru89: done

packages/botonic-plugin-ai-agents/src/ai-agent-client.ts
```diff
@@ -1,112 +1,73 @@
...
-  constructor() {
-    this.url = `${HUBTYPE_API_URL}/external/v1/ai/agent`
-  }
+  constructor(
+    aiAgentArgs: AiAgentArgs,
+    chatModel: BaseChatModel,
+    tools: StructuredTool[] = []
+  ) {
+    this.name = aiAgentArgs.name
+    this.instructions = aiAgentArgs.instructions
```
elozano98, line 31: extra: maybe this instructions and name could be avoid to be saved since we do not use them

packages/botonic-plugin-ai-agents/src/constants.ts
```diff
@@ -1,2 +1,8 @@
 export const HUBTYPE_API_URL =
   process.env.HUBTYPE_API_URL || 'https://api.hubtype.com'
+
+export const AZURE_OPENAI_API_KEY = process.env.AZURE_OPENAI_API_KEY // pragma: allowlist secret
+export const AZURE_OPENAI_API_DEPLOYMENT_NAME =
+  process.env.AZURE_OPENAI_API_DEPLOYMENT_NAME
+export const AZURE_OPENAI_API_VERSION = process.env.AZURE_OPENAI_API_VERSION
+export const AZURE_OPENAI_API_BASE = process.env.AZURE_OPENAI_API_BASE
```
elozano98, line 8: We could define some default values for **some** this env variables in case they are not passed:

```
export const AZURE_OPENAI_API_DEPLOYMENT_NAME = process.env.AZURE_OPENAI_API_DEPLOYMENT_NAME || 'gpt-4o-mini_p1'
export const AZURE_OPENAI_API_VERSION = process.env.AZURE_OPENAI_API_VERSION || '2024-12-01-preview'
```

packages/botonic-plugin-ai-agents/src/hubtype-client.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  constructor(authToken: string) {
+    this.authToken = authToken
+  }
+
+  async getMessages(
+    request: BotContext,
+    memoryLength: number
+  ): Promise<AgenticInputMessage[]> {
+    const url = `${HUBTYPE_API_URL}/external/v1/ai/agent/message_history/`
```
elozano98, line 18: We could move the url to a constant of HubtypeApiClient

packages/botonic-plugin-ai-agents/src/index.ts
```diff
@@ -17,7 +26,31 @@ export default class BotonicPluginAiAgents implements Plugin {
...
-    return this.aiAgentClient.getInference(request, aiAgentArgs)
+  ): Promise<AgenticOutputMessage | undefined> {
+    try {
+      const authToken = isProd ? request.session._access_token : this.authToken
+      if (!authToken) {
+        throw new Error('Auth token is required')
+      }
+
+      const hubtypeClient = new HubtypeApiClient(authToken)
+      const messages = isProd
```
elozano98, line 37: We should only initialize the HubtypeApiClient only if its prod, otherwise it's always initialized even if it's not used.

packages/botonic-plugin-ai-agents/src/utils.ts
```diff
@@ -0,0 +1,24 @@
+import { AzureChatOpenAI } from '@langchain/openai'
+
+import {
+  AZURE_OPENAI_API_BASE,
+  AZURE_OPENAI_API_DEPLOYMENT_NAME,
+  AZURE_OPENAI_API_KEY,
+  AZURE_OPENAI_API_VERSION,
+} from './constants'
+
+export function loadChatModel(provider: 'azureOpenAI') {
```
elozano98, line 10: Maybe we can create sort of an enum with the available providers, not sure if this is usually done on typescript.

packages/botonic-plugin-flow-builder/src/types.ts
```diff
@@ -93,7 +93,8 @@ export interface KnowledgeBaseResponse {
 }
 
 export interface AiAgentResponse {
-  message: { role: string; content: string }
+  role: string
```
elozano98, line 96: This is not always like that for roles: exit, tool

elozano98 (approved):

packages/botonic-plugin-ai-agents/src/ai-agent-client.ts
Iru89: applied in the following PR

packages/botonic-plugin-ai-agents/src/constants.ts
Iru89: applied in the following PR

PR #3035, opened 2025-06-12
botonic-cli: get tools from bot to show on flow builder frontend bot config #BLT-1665

Iru89:
## Description

We need to save in the backend along with the botConfig the tools, payloads and webviews that the bot can use.

## Context

To achieve this, the bot must have a file in src/bot-config.ts that exports a botConfig object that follows the following interface:

```typescript
interface Tool {
  name: string
  description: string
}

interface Webview {
  name: string
}

interface BotConfigJSON {
  tools: Tool[]
  payloads: string[]
  webviews: Webview[]
}
```

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-cli/src/util/bot-config.ts
+++ b/packages/botonic-cli/src/util/bot-config.ts
@@ -13,15 +13,29 @@ const NPM_DEPTH_0 = 0
 // Also get alpha and beta versions
 const versionRegex = /@botonic\/[^@]+@(\d+\.\d+\.\d+(-[a-zA-Z]+\.\d+)?)/
 
+interface BuildInfo {
+  node_version: string
+  npm_version: string
+  botonic_cli_version: string
+}
+
 type BotonicDependencies = Record<string, { version: string }>
 
+interface Tool {
+  name: string
+  description: string
+}
+
+interface Webview {
+  name: string
+}
+
 export interface BotConfigJSON {
-  build_info: {
-    node_version: string
-    npm_version: string
-    botonic_cli_version: string
-  }
+  build_info: BuildInfo
   packages: BotonicDependencies
+  tools?: Tool[]
+  payloads?: string[]
+  webviews?: Webview[]
 }
 export class BotConfig {
   static async get(appDirectory: string): Promise<BotConfigJSON> {
@@ -37,6 +51,10 @@ export class BotConfig {
     )
     const botonicCliVersion =
       botonicCli.match(botonicCliVersionRegex)?.[1] || ''
+
+    const configLoaded = await this.loadBotConfig(appDirectory)
+    console.log('configLoaded', configLoaded)
+
     spinner.succeed()
 
     return {
@@ -46,7 +64,30 @@ export class BotConfig {
         botonic_cli_version: botonicCliVersion,
       },
       packages,
+      // tools: configLoaded.tools,
+      // payloads: configLoaded.payloads,
+      // webviews: configLoaded.webviews,
     }
+  }
+  static async getTools(appDirectory: string) {
+    try {
+      const botConfigPath = path.join(appDirectory, '.src/bot-config.ts')
+      const output = await this.getOutputByCommand(
+        `npx ts-node ${botConfigPath}`
+      )
+      console.log('output', output)
+    } catch (err: any) {
+      console.error(`Error: ${err.message}`)
+    }
+    return []
+  }
+
+  static async loadBotConfig(appDirectory: string) {
+    const configPath = path.resolve(appDirectory, 'src/bot-config.ts')
+
+    const { botConfig } = await import(configPath)
+
+    return botConfig
   }
 
   private static async getBotonicDependencies(
--- a/packages/botonic-cli/src/util/bot-config.ts
+++ b/packages/botonic-cli/src/util/bot-config.ts
@@ -53,7 +53,6 @@ export class BotConfig {
       botonicCli.match(botonicCliVersionRegex)?.[1] || ''
 
     const configLoaded = await this.loadBotConfig(appDirectory)
-    console.log('configLoaded', configLoaded)
 
     spinner.succeed()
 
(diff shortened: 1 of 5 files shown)


elozano98 (approved):

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -13,15 +13,29 @@ const NPM_DEPTH_0 = 0
...
-  build_info: {
-    node_version: string
-    npm_version: string
-    botonic_cli_version: string
-  }
+  build_info: BuildInfo
   packages: BotonicDependencies
+  tools?: Tool[]
+  payloads?: string[]
+  webviews?: Webview[]
```
vanbasten17: good! but just a question, are these values used along with tools, or are more generic so webviews and payloads can be used also for future configs?
also, should we name tools like `agent_tools` or similar so it's clearer when we read the bot config? or can we use an object to define this like... `agents_config?: {tools, etc.}`

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -46,6 +63,49 @@ export class BotConfig {
...
       },
       packages,
+      tools: configLoaded.tools,
+      payloads: configLoaded.payloads,
+      webviews: configLoaded.webviews,
+    }
+  }
+  static async getTools(appDirectory: string) {
+    try {
+      const botConfigPath = path.join(appDirectory, '.src/bot-config.ts')
```
vanbasten17: we must use path utils to create this path: `'.src/bot-config.ts'`, otherwise could not work in all operative systems. Use constants for bot config file. Was `bot-config.ts` or `bot.config.ts`?

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -46,6 +63,49 @@ export class BotConfig {
...
+      const output = await this.getOutputByCommand(
+        `npx ts-node ${botConfigPath}`
+      )
+      console.log('output', output)
```
vanbasten17: do we want this console log?

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -46,6 +63,49 @@ export class BotConfig {
...
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      require('ts-node').register({
+        transpileOnly: true,
+        compilerOptions: {
+          module: 'commonjs',
+          target: 'es2018',
+          esModuleInterop: true,
+          allowSyntheticDefaultImports: true,
+        },
+      })
```
vanbasten17: put this ts config in an object so we can handle it better if we need to change it later

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -46,6 +63,49 @@ export class BotConfig {
...
+
+      const configPath = path.resolve(appDirectory, 'src/bot-config.ts')
+
+      // Clean cache to reload if necessary
+      delete require.cache[require.resolve(configPath)]
```
vanbasten17, line 90: what does it do?

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -46,6 +63,49 @@ export class BotConfig {
...
+
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const { botConfig } = require(configPath)
+      return botConfig
+    } catch (error) {
+      console.error('Error cargando bot config:', error)
```
vanbasten17: change copy to english

vanbasten17 (approved):
Approved, but I think we must apply part of the feedback, as this will be used a lot

packages/botonic-cli/src/util/bot-config.ts
Iru89: I removed this function because I wasn't using it at all

packages/botonic-cli/src/util/bot-config.ts
Iru89: I have already merged the backend PR with the parser. These objects are only used here for configuration; the objects that define the tools with their functions are imported from the ai-agents plugin

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -46,6 +63,49 @@ export class BotConfig {
...
+      const botConfigPath = path.join(appDirectory, '.src/bot-config.ts')
+      const output = await this.getOutputByCommand(
+        `npx ts-node ${botConfigPath}`
+      )
+      console.log('output', output)
+    } catch (err: any) {
+      console.error(`Error: ${err.message}`)
+    }
+    return []
+  }
```
asastre: Doubt: Is this function used at all? If it is not used you can remove it.

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -46,6 +63,49 @@ export class BotConfig {
...
+
+      // Clean cache to reload if necessary
+      delete require.cache[require.resolve(configPath)]
+
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const { botConfig } = require(configPath)
+      return botConfig
+    } catch (error) {
+      console.error('Error cargando bot config:', error)
+      throw error
```
asastre: Doubt: Shouldn't we check first if the path exists and if it doesn't exist don't do nothing instead of throwing an error if the file doesn't exist? The bot may not have this file and the deploy shouldn't fail if it doesn't exist.

packages/botonic-cli/src/util/bot-config.ts
Iru89: removed

packages/botonic-cli/src/util/bot-config.ts
Iru89: good point, I will review it and ask you to review it again

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -46,6 +63,46 @@ export class BotConfig {
...
+        transpileOnly: true,
+        compilerOptions: {
+          module: 'commonjs',
+          target: 'es2018',
+          esModuleInterop: true,
+          allowSyntheticDefaultImports: true,
+        },
+      }
+      // Register ts-node to load TypeScript in real time
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
```
Iru89: add `require.extensions['.png'] = () => '';` before require('ts-node')

Iru89: require.extensions is deprecated

PR #3036, opened 2025-06-16
plugin flow-builder & ai-agents: rename tools to active_tools in flow ai-agent node

Iru89:
## Description

Rename tools to active_tools in ai-agent node(plugin-flow-builder)
Rename tools to activeTools in aiAgentArgs (plugin-ai-agents)

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/content-fields/flow-ai-agent.tsx
+++ b/packages/botonic-plugin-flow-builder/src/content-fields/flow-ai-agent.tsx
@@ -7,12 +7,14 @@ export class FlowAiAgent extends ContentFieldsBase {
   public code: string = ''
   public name: string = ''
   public instructions: string = ''
+  public activeTools?: { name: string }[]
   public text: string = ''
 
   static fromHubtypeCMS(component: HtAiAgentNode): FlowAiAgent {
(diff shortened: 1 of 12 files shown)


packages/botonic-plugin-flow-builder/src/content-fields/flow-ai-agent.tsx
```diff
@@ -7,12 +7,14 @@ export class FlowAiAgent extends ContentFieldsBase {
   public code: string = ''
   public name: string = ''
   public instructions: string = ''
+  public activeTools?: { name: string }[]
```
vanbasten17, line 10: is it worth create the interface `interface Tool { name: string }`? It is redefined in more parts of the code

elozano98 (approved):

packages/botonic-plugin-flow-builder/src/content-fields/flow-ai-agent.tsx
Iru89: We already have the CustomTool interface when we create it inside the plugin-ai-agents. Here it is only what is received from the node to execute it. For now I will leave it this way if later we need to pass more attributes through the flow builder node I will create an interface.

PR #3037, opened 2025-06-17
plugin-flow-builder: update system locale when plugin flow builder resolve locale #BLT-1717 

Iru89:
## Description

- Update system locale when plugin-flow-builder resolve a locale with a value different from the session's value
- Do not use FlowBuilderAction inside the Multichannel because it adds a child to the React tree.

## Context

- We need to update the system_locale whenever the flow builder plugin uses a different one than the one that exists in the session to correctly receive @botonic/plugin-hubtype-analytics events.
- Multichannel does not work as expected if we add one more level of children in the React tree. Multichannel expects the children to be directly @botonic/react messages such as Text, Carousel, etc.

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/action/index.tsx
+++ b/packages/botonic-plugin-flow-builder/src/action/index.tsx
@@ -77,13 +77,15 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
   render(): JSX.Element | JSX.Element[] {
     const { contents, webchatSettingsParams } = this.props
     const request = this.context as ActionRequest
+    const shouldSendWebchatSettings =
+      (isWebchat(request.session) || isDev(request.session)) &&
+      !!webchatSettingsParams
 
     return (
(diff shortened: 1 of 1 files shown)


packages/botonic-plugin-flow-builder/src/action/index.tsx
```diff
@@ -92,9 +92,16 @@ export class FlowBuilderAction extends React.Component<FlowBuilderActionProps> {
 
 export class FlowBuilderMultichannelAction extends FlowBuilderAction {
   render(): JSX.Element | JSX.Element[] {
+    const { contents, webchatSettingsParams } = this.props
+    const request = this.context as ActionRequest
+
```
vanbasten17, line 102: I would add a `const shouldSendWebchatSettings = (isWebchat(request.session) || isDev(request.session)) && !!webchatSettingsParams`, then use it below

vanbasten17 (approved):

PR #3050, opened 2025-06-30
react: Detect automatically if the webchat is running in iOS/Android webview

asastre:
## Description
Set the default target parameter in links depending if the webchat is opened within a iOS/Android webview (`_self`) or not (`_blank`)

## Context
If the webchat is displayed within an iOS webview (inside an iOS App), the links weren't working because the default target was `_blank` and on iOS webviews the links only work if the target is `_self`.

Create a new `isInWebView` function to detect if the environment is a webview in iOS/Android.

## Testing

### DIFF
--- a/packages/botonic-react/src/util/environment.js
+++ b/packages/botonic-react/src/util/environment.js
@@ -55,3 +55,19 @@ function normalizeArray(parts, allowAboveRoot) {
   }
   return res
 }
+
+export function isInWebView() {
+  const userAgent = window.navigator.userAgent || ''
+  const standalone = window.navigator.standalone
+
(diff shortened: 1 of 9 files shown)


packages/botonic-react/src/util/environment.js
```diff
@@ -55,3 +55,19 @@ function normalizeArray(parts, allowAboveRoot) {
   }
   return res
 }
+
+export function isInWebView() {
```
Iru89: Can we use a more descriptive name, for example isInWebviewApp?

asastre: Changed!

Iru89 (approved):

PR #3060, opened 2025-07-14
Feat: Implement retry on AI Agent call

elozano98:
## Description

Implement retry system when running AI Agent if fails.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25536,7 +25536,7 @@
     },
     "packages/botonic-plugin-ai-agents": {
       "name": "@botonic/plugin-ai-agents",
-      "version": "0.36.0-alpha.7",
+      "version": "0.36.0-alpha.8",
       "dependencies": {
         "@botonic/core": "^0.36.0",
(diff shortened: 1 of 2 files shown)


packages/botonic-plugin-ai-agents/src/runner.ts
```diff
@@ -43,14 +44,33 @@ export class AIAgentRunner {
...
+      })
+      const result = await runner.run(this.agent, messages)
+      return result.finalOutput?.messages || []
+    } catch (error) {
+      if (!(error instanceof ModelBehaviorError)) {
+        throw error
+      }
+      if (attempt > maxRetries) {
+        throw error
+      }
```
Iru89, line 71: You can use a single if

```suggestion
      if (!(error instanceof ModelBehaviorError || attempt > maxRetries)) {
        throw error
      }
```

elozano98: I would prefer to keep it separate to emphasize the two reasons for not performing the retry.

PR #3063, opened 2025-07-16
plugin-flow-builder: rating flow #BLT-1745

Iru89:
## Description

Add rating flow to resolve new rating node from flow builder.
When resolve rating node in WhatsApp use a button list otherwise use a text with buttons.

## Context

Each button has a payload with `agent-rating|${button.id}`.

When plugin receive a payload that starts with `agent-rating` takes the button.id and search the rating node and the this button in the flow. Then use the fields in the button rating node to track and store the rating.

Captura de pantalla 2025-07-16 a las 11 17 36

## Testing

The pull request...

- has unit tests
- has integration tests


packages/botonic-plugin-flow-builder/src/action/payload.ts
```diff
@@ -19,3 +26,46 @@ export async function getContentsByPayload({
...
+  context: FlowBuilderContext
+): Promise<FlowContent[]> {
+  const { cmsApi, flowBuilderPlugin, request } = context
+
+  if (!request.input.payload) {
+    return []
+  }
+
+  const id = request.input.payload
+  const buttonId = id?.split(SEPARATOR)[1]
```
vanbasten17, line 40: alternatively, you can also do: `const [, buttonId] = id?.split(SEPARATOR) ?? []`

vanbasten17 (approved):

Davidhidalgo (approved):

PR #3070, opened 2025-07-25
ai-plugin-agents: update README.md

Iru89:
## Description

Update README.md


packages/botonic-plugin-ai-agents/README.md
```diff
@@ -2,13 +2,155 @@
...
+    }),
+    new rspack.ProgressPlugin(),
+  ]
+```
+
+## Custom Tools
+
+- It is possible to add tools from the bot code. To do this:
+- Add the `zod` dependency.
+- Create and export the definition of the tools in a file, for example: `src/server/tools.ts`.
```
elozano98, line 46: I think this is not optional, the tools need to be defined here. If not, the bot config will not be updated with the custom tools. right?

packages/botonic-plugin-ai-agents/README.md
```diff
@@ -2,13 +2,155 @@
...
+    aiAgents: getAiAgentsConfig(),
+    knowledgeBases: getKnowledgeBasesConfig(),
+  },
+  ...
+}
+```
+
+## Share customTools with flow builder frontend
+
+- Create new `src/bot-config.ts` file and add customTools. This file is used so that every time a bot is deployed, the Flow Builder frontend can know which tools are in the deployed bot and enable or disable tools for each AI Agent node.
```
elozano98, line 133: Okay, I see there is no mandatory folder where tools need to be created since you need to define here from where you collect the tools. 👌🏽

packages/botonic-plugin-ai-agents/README.md
Iru89: They can be created in a different file, the important thing is that they are then used in the src/bot-config.ts file and within the plugin-ai-agents configuration.

PR #3076, opened 2025-08-04
Feat: Use contact_info on AI Agent System Prompt #BLT-1779

elozano98:
## Description

Add the `contact_info` of the user (provided by the bot session) as part of the system prompt of the AI Agent.

### DIFF
--- a/packages/botonic-plugin-ai-agents/src/agent-builder.ts
+++ b/packages/botonic-plugin-ai-agents/src/agent-builder.ts
@@ -33,28 +33,25 @@ export class AIAgentBuilder {
     initialInstructions: string,
     contactInfo: ContactInfo
   ): string {
-    let instructions = `<instructions>\n${initialInstructions}\n</instructions>`
-    instructions = this.addContactInfo(instructions, contactInfo)
-    instructions = this.addMetadata(instructions)
-    return this.addOutputInstructions(instructions)
+    const metadataInstructions = this.getMetadataInstructions()
+    const contactInfoInstructions = this.getContactInfoInstructions(contactInfo)
+    const outputInstructions = this.getOutputInstructions()
+    return `${initialInstructions}\n\n${metadataInstructions}\n\n${contactInfoInstructions}\n\n${outputInstructions}`
   }
 
-  private addContactInfo(
-    instructions: string,
-    contactInfo: ContactInfo
-  ): string {
+  private getContactInfoInstructions(contactInfo: ContactInfo): string {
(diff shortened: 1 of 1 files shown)


packages/botonic-plugin-ai-agents/src/agent-builder.ts
```diff
@@ -2,16 +2,22 @@ import { Agent } from '@openai/agents'
...
+  constructor(
+    name: string,
+    instructions: string,
+    tools: Tool[],
+    contactInfo: ContactInfo
+  ) {
     this.name = name
-    this.instructions = this.addExtraInstructions(instructions)
+    this.instructions = this.addExtraInstructions(instructions, contactInfo)
+    console.log('AI AgentInstructions:\n', this.instructions)
```
vanbasten17: remove console.log?

packages/botonic-plugin-ai-agents/src/agent-builder.ts
```diff
@@ -24,8 +30,32 @@ export class AIAgentBuilder {
     })
   }
 
-  private addExtraInstructions(instructions: string): string {
+  private addExtraInstructions(
```
vanbasten17, line 32: optional: no need to split so much:
```
private addExtraInstructions(
  initialInstructions: string,
  contactInfo: ContactInfo
): string {
  const instructionsSection = `<instructions>\n${initialInstructions}\n</instructions>`

  const contactInfoSection = `<contact_info>\n${
    Object.entries(contactInfo)
      .map(([key, value]) => `${key}: ${value}`)
      .join('\n')
  }\n</contact_info>`

  const metadataSection = `<metadata>\nCurrent Date: ${new Date().toISOString()}\n</metadata>`

  const outputInstructionsSection = this.addOutputInstructions(
    `${instructionsSection}\n\n${contactInfoSection}\n\n${metadataSection}`
  )

  return outputInstructionsSection
}
```

vanbasten17 (approved):

packages/botonic-plugin-ai-agents/src/agent-builder.ts
Davidhidalgo: Half agree, i think is more legible the approach of Marc but the abstraction for each section feels good too. Easy to test and maintain ;)

Davidhidalgo: The unnecesary thing is to modify the string on each abstraction. Just build all the parts on this function and let the parts abstracted into the functions

Davidhidalgo (approved):

PR #3077, opened 2025-08-05
botonic-react: input panel buttons no need enable attribute set to true to be used

Iru89:
## Description

Change behaviour to use custom buttons in input panel component. If you define a custom component for some button (Attachments, Emojis or SendButton) you don't need to put enable attribute set to true. If you define a custo button but you set enable to false, the custom button is not used.

### DIFF
--- a/packages/botonic-plugin-flow-builder/package.json
+++ b/packages/botonic-plugin-flow-builder/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/plugin-flow-builder",
-  "version": "0.37.0",
+  "version": "0.37.1-alpha.1",
   "main": "./lib/cjs/index.js",
   "module": "./lib/esm/index.js",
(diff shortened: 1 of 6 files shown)


packages/botonic-plugin-flow-builder/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@botonic/plugin-flow-builder",
-  "version": "0.37.0",
+  "version": "0.37.1-alpha.1",
```
asastre, line 3: Doubt: Why updating the version of the Flow Builder plugin?

Iru89: Because I had published an alpha and it was not in any PR. 
But when I test alphas I like that the bot has only one version of @botonic/react and the plugin-flow-builder uses @botonic/react internally.

asastre (approved):

PR #3079, opened 2025-08-06
botonic-react: create updateCustomMessageProps function inside WebchatContext #BLT-1787

Iru89:
## Description

Create `updateCustomMessageProps` inside WebchatContext. This function update the props that receive all custom messages. This is a util to keep the state in a custom message when minimize the webchat or reload the browser page. 

## Context

The props.id is the message id, this is used to update the custom message props. This id is undefined in local development.
To keep the state between differents renders  you should init the internal state from your custom message using the props received from props and update this props using the new function `updateCustomMessageProps`

Example how use this hook in the CustomRating component.

```typescript
interface CustomRatingMessageProps {
  payloads: string[]
  messageText: string
  buttonText: string
  ratingType: RatingType
  id?: string // This id not exist in local development
  valueSent?: number
}

const CustomRatingMessage: React.FC<CustomRatingMessageProps> = props => {
  const { payloads, messageText, buttonText, ratingType, id, valueSent } = props
  const { updateCustomMessageProps, sendInput } = useContext(WebchatContext)

  const theme = useContext(ThemeContext)
  const color = theme?.brand?.color ?? ''

  const [ratingValue, setRatingValue] = useState(valueSent ? valueSent : -1)
  const [showRating, setShowRating] = useState(true)

  const ratingChanged = (newRating: number) => {
    setRatingValue(newRating)
  }

  const handleButtonSend = () => {
    if (ratingValue === -1) return

    const json = {
      valueSent: ratingValue,
    }
    updateCustomMessageProps(json, id)


    setShowRating(false)
    const payload = payloads[ratingValue - 1]

    const input = {
      type: INPUT.POSTBACK as InputType,
      payload,
    }
    void sendInput(input)
  }

  const disabled = ratingValue === -1

  return (
    <div className='rating-message-container'>
      <MessageBubble className='rating-message-bubble'>
        {messageText}
        <RatingSelector
          color={color}
          onRatingChange={ratingChanged}
          ratingValue={ratingValue}
          ratingType={ratingType}
          valueSent={valueSent}
        />
      </MessageBubble>
      {!props?.valueSent && showRating && (
        <Button
          autodisable={true}
          disabled={disabled}
          onClick={handleButtonSend}
        >
          {buttonText}
        </Button>
      )}
    </div>
  )
}

  name: 'custom-rating',
  component: CustomRatingMessage,
})

```

### DIFF
--- a/packages/botonic-react/src/webchat/custom-messages/rating/index.tsx
+++ b/packages/botonic-react/src/webchat/custom-messages/rating/index.tsx
@@ -34,12 +34,10 @@ const CustomRatingMessage: React.FC<CustomRatingMessageProps> = props => {
   const handleButtonSend = () => {
     if (ratingValue === -1) return
 
-    updateCustomMessageProps(
-      {
-        valueSent: ratingValue,
-      },
-      id
-    )
+    const json = {
+      valueSent: ratingValue,
+    }
+    updateCustomMessageProps(json, id)
 
     setShowRating(false)
     const payload = payloads[ratingValue - 1]


packages/botonic-react/src/webchat/context/index.tsx
```diff
@@ -54,6 +54,9 @@ export const WebchatContext = createContext<WebchatContextProps>({
   togglePersistentMenu: () => {
     return
   },
+  updateCustomJsonMessage: () => {
```
vanbasten17: rename to `updateCustomMessageJSON` so its clear that it references `CustomMessage`'s

packages/botonic-react/src/webchat/context/messages-reducer.ts
```diff
@@ -14,6 +16,8 @@ export const messagesReducer = (
       return addMessageComponent(state, action)
     case WebchatAction.UPDATE_MESSAGE:
       return updateMessageReducer(state, action)
+    case WebchatAction.UPDATE_CUSTOM_JSON_MESSAGE:
+      return updateCustomJsonMessageReducer(state, action)
```
vanbasten17: same

vanbasten17 (approved):

packages/botonic-react/src/webchat/context/index.tsx
Iru89: applied

packages/botonic-react/src/webchat/custom-messages/rating/index.tsx
```diff
@@ -33,37 +33,23 @@ const CustomRatingMessage: React.FC<CustomRatingMessageProps> = props => {
...
-    if (ratingValue !== -1) {
-      setShowRating(false)
+    if (ratingValue === -1) return
 
-      const payload = payloads[ratingValue - 1]
+    if (props.json?.messageId) {
+      updateCustomMessageJSON(props.json.messageId, {
+        valueSent: ratingValue,
+      })
+    }
```
asastre: Doubt: Shouldn't the props contain the message ID in an `id` attribute? Shouldn't we be able to send the message ID directly [here](https://github.com/hubtype/botonic/blob/master-lts/packages/botonic-react/src/components/custom-message.tsx#L85) as props?

asastre (changes requested):
Review my comments @Iru89

packages/botonic-react/src/webchat/custom-messages/rating/index.tsx
Iru89: I refactor all PR to keep state from props instead of props.json

packages/botonic-react/src/webchat/custom-messages/rating/index.tsx
```diff
@@ -14,56 +13,41 @@ interface CustomRatingMessageProps {
   messageText: string
   buttonText: string
   ratingType: RatingType
-  json: { messageId?: string; valueSent?: number }
+  id?: string // This id not exist in local development
```
asastre, line 16: Doubt: If this `id` does not exist in local development, then we won't be able to test the custom message properly locally?

vanbasten17 (approved):

packages/botonic-react/src/webchat/custom-messages/rating/index.tsx
```diff
@@ -14,56 +13,41 @@ interface CustomRatingMessageProps {
...
-    if (ratingValue !== -1) {
-      setShowRating(false)
+    if (ratingValue === -1) return
 
-      const payload = payloads[ratingValue - 1]
+    if (id) {
+      updateCustomMessageProps(id, {
+        valueSent: ratingValue,
+      })
+    }
```
asastre: Doubt: Would it be better to move this conditional inside the `updateCustomMessageProps`? If we don't do it we will have to add this condition in all of the custom messages that we create, right?

packages/botonic-react/src/webchat/custom-messages/rating/index.tsx
Iru89: no

packages/botonic-react/src/webchat/custom-messages/rating/index.tsx
Iru89: yes I can move this condition inside the function

Iru89: applied

PR #3083, opened 2025-08-07
plugin-ai-agents & plugin-flow-builder: create InferenceResponse for ai-agents #BLT-1803

Iru89:
## Description

plugin-ai-agents return an object with interface InferenceResponse. With this object we can know the tools executed or if ai agents not respond (exit is true). In the future we include in this response the guardrails info.

```typescript
export type InferenceResponse = {
  messages: AgenticOutputMessage[]
  toolsExecuted: string[]
  exit: boolean
  inputGuardrailTriggered: boolean
  outputGuardrailTriggered: boolean
} | undefined
```

plugin -flow-builder use this object to generate response, in the future PR use this info to track ai-agents responses

## Test

Update mockAiAgentResponse to pass tests

### DIFF
--- a/packages/botonic-plugin-ai-agents/src/index.ts
+++ b/packages/botonic-plugin-ai-agents/src/index.ts
@@ -8,10 +8,10 @@ import { setUpOpenAI } from './openai'
 import { AIAgentRunner } from './runner'
 import {
   AgenticInputMessage,
-  AgenticOutputMessage,
   AiAgentArgs,
   Context,
   CustomTool,
+  InferenceResponse,
   PluginAiAgentOptions,
   Tool,
 } from './types'
@@ -33,7 +33,7 @@ export default class BotonicPluginAiAgents implements Plugin {
   async getInference(
     request: BotContext,
     aiAgentArgs: AiAgentArgs
-  ): Promise<AgenticOutputMessage[] | undefined> {
+  ): Promise<InferenceResponse> {
     try {
       const authToken = isProd ? request.session._access_token : this.authToken
       if (!authToken) {
@@ -58,6 +58,8 @@ export default class BotonicPluginAiAgents implements Plugin {
       return await runner.run(messages, context)
     } catch (error) {
       console.error('error plugin returns undefined', error)
+      // Here we can return a InferenceResponse as a exit
+      // but indicate that the inference failed
       return undefined
     }
   }
--- a/packages/botonic-plugin-ai-agents/src/runner.ts
+++ b/packages/botonic-plugin-ai-agents/src/runner.ts
@@ -1,6 +1,12 @@
 import { Runner, RunToolCallItem } from '@openai/agents'
 
-import { AgenticInputMessage, AIAgent, Context, RunResult } from './types'
+import {
+  AgenticInputMessage,
+  AgenticOutputMessage,
+  AIAgent,
+  Context,
+  RunResult,
+} from './types'
 
 export class AIAgentRunner {
   private agent: AIAgent
@@ -38,7 +44,9 @@ export class AIAgentRunner {
       return {
         messages: hasExit
           ? []
-          : outputMessages.filter(message => message.type !== 'exit'),
+          : (outputMessages.filter(
+              message => message.type !== 'exit'
+            ) as AgenticOutputMessage[]),
         toolsExecuted,
         exit: hasExit,
         inputGuardrailTriggered: false,
--- a/packages/botonic-plugin-ai-agents/src/runner.ts
+++ b/packages/botonic-plugin-ai-agents/src/runner.ts
@@ -1,10 +1,11 @@
-import { Runner } from '@openai/agents'
+import { Runner, RunToolCallItem } from '@openai/agents'
 
 import {
   AgenticInputMessage,
   AgenticOutputMessage,
   AIAgent,
   Context,
+  RunResult,
 } from './types'
 
 export class AIAgentRunner {
@@ -18,7 +19,7 @@ export class AIAgentRunner {
     messages: AgenticInputMessage[],
     context: Context,
     maxRetries: number = 1
-  ): Promise<AgenticOutputMessage[]> {
+  ): Promise<RunResult> {
     return this.runWithRetry(messages, context, maxRetries, 1)
   }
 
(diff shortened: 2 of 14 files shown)


packages/botonic-plugin-ai-agents/src/index.ts
```diff
@@ -58,6 +58,8 @@ export default class BotonicPluginAiAgents implements Plugin {
       return await runner.run(messages, context)
     } catch (error) {
       console.error('error plugin returns undefined', error)
+      // Here we can return a InferenceResponse as a exit
```
Iru89, line 61: @elozano98

packages/botonic-plugin-ai-agents/src/runner.ts
```diff
@@ -27,13 +28,34 @@ export class AIAgentRunner {
...
-      return result.finalOutput?.messages || []
+      console.log('result', result)
+
+      const outputMessages = result.finalOutput?.messages || []
+      const hasExit = outputMessages.some(message => message.type === 'exit')
+      const toolsExecuted =
+        result.newItems
+          ?.filter(item => item instanceof RunToolCallItem)
+          .map(item => {
+            if (item.rawItem.type === 'function_call') {
```
vanbasten17: can these items be typed also?

vanbasten17: you can simplify by using `flatMap` if available in ts version
```
const toolsExecuted = result.newItems?.flatMap(item =>
  item instanceof RunToolCallItem && item.rawItem.type === 'function_call'
    ? [item.rawItem.name]
    : []
) || [];
```

packages/botonic-plugin-ai-agents/src/runner.ts
```diff
@@ -27,13 +28,34 @@ export class AIAgentRunner {
...
+            return ''
+          })
+          .filter(item => item !== '') || []
+
+      return {
+        messages: hasExit ? [] : outputMessages,
+        toolsExecuted,
+        exit: hasExit,
+        inputGuardrailTriggered: false,
+        outputGuardrailTriggered: false,
```
vanbasten17, line 53: is this expected to be always false?

packages/botonic-plugin-ai-agents/src/types.ts
```diff
@@ -37,3 +37,13 @@ export interface AiAgentArgs {
 
 export type AgenticInputMessage = AgentInputItem
 export type AgenticOutputMessage = OutputMessage
+
+export interface RunResult {
+  messages: AgenticOutputMessage[]
```
elozano98, line 43: The `AgenticOutputMessage` should be a `Exclude<OutputMessage, ExitMessage>`

packages/botonic-plugin-flow-builder/src/types.ts
```diff
@@ -140,12 +140,15 @@ export interface CarouselMessage extends OutputBaseMessage {
   }
 }
 
-export interface ExitMessage extends OutputBaseMessage {
-  type: 'exit'
-}
-
 export type AgenticOutputMessage =
```
elozano98, line 143: Why we need to redefine here these types? Could we import them from the `plugin-ai-agent`?

packages/botonic-plugin-flow-builder/src/types.ts
```diff
@@ -140,12 +140,15 @@ export interface CarouselMessage extends OutputBaseMessage {
...
   | TextMessage
   | TextWithButtonsMessage
   | CarouselMessage
-  | ExitMessage
+
+export interface InferenceResponse {
```
elozano98, line 148: same here

packages/botonic-plugin-ai-agents/src/runner.ts
```diff
@@ -27,13 +28,34 @@ export class AIAgentRunner {
...
     attempt: number
-  ): Promise<AgenticOutputMessage[]> {
+  ): Promise<RunResult> {
     try {
       const runner = new Runner({
         modelSettings: { temperature: 0 },
       })
       const result = await runner.run(this.agent, messages, { context })
-      return result.finalOutput?.messages || []
+      console.log('result', result)
```
elozano98: extra: remove the console log

packages/botonic-plugin-ai-agents/src/runner.ts
```diff
@@ -27,13 +28,34 @@ export class AIAgentRunner {
...
+
+      const outputMessages = result.finalOutput?.messages || []
+      const hasExit = outputMessages.some(message => message.type === 'exit')
```
elozano98: if the previous line fails and end up with an empty array on `outputMessages` this hasExit would not be true (should be)

packages/botonic-plugin-ai-agents/src/runner.ts
```diff
@@ -1,10 +1,11 @@
...
 
 import {
   AgenticInputMessage,
   AgenticOutputMessage,
   AIAgent,
   Context,
+  RunResult,
 } from './types'
 
 export class AIAgentRunner {
```
elozano98, line 11: we should add tests for this class since the logic inside has increased:
- what if no messages are returned
- what if exit message on the response
- what if no exit messages on the response
- what if tools in the response
- what if tools in the response but are not function_calls
- what if no tools in the response

packages/botonic-plugin-flow-builder/src/types.ts
Iru89: We should not import things from one plugin to another. At some point we can take the types to @botonic/core and import them always from there

packages/botonic-plugin-flow-builder/src/types.ts
Iru89: Here I will never get an exit tool as it is removed inside the plugin-ai-agents

packages/botonic-plugin-ai-agents/src/runner.ts
Iru89: ok I will check it

packages/botonic-plugin-ai-agents/src/runner.ts
Iru89: item can be of more than one type, so I make the first filter and only make the map on those that are instances of RunToolCallItem

packages/botonic-plugin-ai-agents/src/runner.ts
Iru89: For now, yes, it's something Eric will need to change in future PR

packages/botonic-plugin-ai-agents/src/runner.ts
Iru89: applied

Iru89:
@elozano98 I added config for jest and a simple test, after merge this PR the tests are executed in every commit

PR #3086, opened 2025-08-08
plugin-hubtype-analytics & plugin-flow-builder: create and send ai agent event #BLT-1804

Iru89:
## Description

**@botonic/core**
Add type of ai-agent-event
Update InferenceResponse of ai agent with an error bolean

**@botonic/plugin-ai-agents**
Add error boolean in the response of the inference

**@botonic/plugin-hubtype-analytics**
Add ai agent event

**@botonic/plugin-flow-builder** 
Use all attributes of inference response to track event.
Send ai agent event after inference
Update mock to use InferenceResponse

## Testing

Add tests in both plugins. 
- In hubtype analytics to check that event has all attributes.
- In flow builder to check that event is send with all attributes after inference.

### DIFF
--- a/packages/botonic-core/CHANGELOG.md
+++ b/packages/botonic-core/CHANGELOG.md
@@ -20,6 +20,16 @@ All notable changes to Botonic will be documented in this file.
 
 </details>
 
+## [0.40.x] - 2025-mm-dd
+
+### Added
+
+- [BLT-1804](https://github.com/hubtype/botonic/pull/1804): Add AI Agent events types.
+
+### Changed
+
+### Fixed
+
 ## [0.38.0] - 2025-08-20
 
 ## Added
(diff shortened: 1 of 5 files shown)


packages/botonic-plugin-flow-builder/src/action/ai-agent.ts
```diff
@@ -45,3 +51,48 @@ export async function getContentsByAiAgent({
...
+    flowId: flowId,
+    flowName: flowName,
+    flowNodeId: aiAgentContent.id,
+    flowNodeContentId: aiAgentContent.code,
+    flowNodeIsMeaningful: true,
+    toolsExecuted: aiAgentResponse.toolsExecuted,
+    exit: aiAgentResponse.exit,
+    inputGuardrailTriggered: [], //aiAgentResponse.inputGuardrailTriggered,
+    outputGuardrailTriggered: [], //aiAgentResponse.outputGuardrailTriggered,
+    error: false, // aiAgentResponse.error,
```
Iru89: Update this attributes when plugin ai agents return real values

vanbasten17 (approved):

packages/botonic-plugin-flow-builder/src/action/ai-agent.ts
Iru89: applied

PR #3087, opened 2025-08-08
refactor: create hubtype-analytics interfaces and enums in @botonic/core #BLT-1806

Iru89:
## Description

Refactor hubtype-analytics, pass all interfaces and enums to @botonic/core to be able to import them in @botonic/plugin-flow-builder and @botonic/react to type events

Import from @botonic/core interfaces and enums and export from @botnic/plugin-hubtype-analytics, this way it is not necessary to change the imports in the bots.

In the last commit I create a functions to track custom events a feedback events from webchat.

### DIFF
--- a/packages/botonic-core/src/models/hubtype-analytics.ts
+++ b/packages/botonic-core/src/models/hubtype-analytics.ts
@@ -1,0 +1,157 @@
+export enum EventAction {
+  FeedbackCase = 'feedback_case',
+  FeedbackMessage = 'feedback_message',
+  FeedbackConversation = 'feedback_conversation',
+  FeedbackKnowledgebase = 'feedback_knowledgebase',
+  FeedbackWebview = 'feedback_webview',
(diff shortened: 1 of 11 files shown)


packages/botonic-core/src/models/hubtype-analytics.ts
```diff
@@ -0,0 +1,157 @@
+export enum EventAction {
```
vanbasten17, line 1: so moving them here is just to avoid having analytics as an extra dependency in fbplugin and react right?

Iru89: Yes, to be able to use the types in @botonic/react and @botonic/plugin-flowbuilder, without having the dependency on @botonic/plugin-hubtype-analytics

killev: Fixed: replaced type assertion with errors.As for proper wrapped error handling

PR #3088, opened 2025-08-13
plugin-knowledge-bases and plugin-ai-agents types in core

Iru89:
## Description

Just like the hubtype-analytics types, pass the knowledge bases and AI agents types to @botonic/core. This way, we have all the types in one place and they can be imported 

## Testing

Update types used in tests to pass have all in green


elozano98 (approved):

packages/botonic-core/src/models/ai-agents.ts
```diff
@@ -0,0 +1,54 @@
+export interface BaseMessage {
+  type: 'text' | 'textWithButtons' | 'carousel' | 'exit'
```
vanbasten17, line 2: can we change `textWithButtons ` --> `buttonMessage` as we have in backend?

packages/botonic-core/src/models/knowledge-bases.ts
```diff
@@ -0,0 +1,7 @@
+export interface KnowledgeBaseResponse {
```
vanbasten17: I would name the file `knowledge-base.ts` to match `KnowledgeBaseResponse`

packages/botonic-plugin-flow-builder/src/types.ts
```diff
@@ -1,4 +1,14 @@
...
+import {
+  AgenticOutputMessage,
+  BotContext,
+  CarouselMessage,
+  InferenceResponse,
+  KnowledgeBaseResponse,
+  PluginPreRequest,
+  ResolvedPlugins,
+  TextMessage,
+  TextWithButtonsMessage,
```
vanbasten17, line 10: refactor to be `ButtonMessage`

packages/botonic-core/src/models/index.ts
```diff
@@ -1,4 +1,6 @@
+export * from './ai-agents'
 export * from './events'
 export * from './hubtype-analytics'
+export * from './knowledge-bases'
 export * from './legacy-types'
```
vanbasten17, line 4: `events` and `user` can be removed but ensuring we don't break code.

vanbasten17 (approved):

packages/botonic-core/src/models/knowledge-bases.ts
Iru89: but they are the types of the plugin-knowledge-bases.
Do I rename it to plugin-knowledge-bases?
Do I do the same with the files ai-agents and hubtype-analytics?

packages/botonic-core/src/models/ai-agents.ts
Iru89: if I apply this change it will be a breaking change in the flow builder plugin. Do we do it when we release a 0.38.0 version?

packages/botonic-core/src/models/knowledge-bases.ts
vanbasten17: so then just `KnowledgeBasesResponse`

Iru89: renamed KnowledgeBaseResponse to KnowledgeBasesResponse

packages/botonic-core/src/models/index.ts
Iru89: removed

PR #3091, opened 2025-08-14
feat: plugin-flow-builder return text messages instead of whatsapp me…

Iru89:
## Description

I first tried to make the change in the Multichannel, but it does not work well and does not chain the changes. For example if you send a list to Webchat it works correctly and sends a text with buttons but if you send a list to Facebook it doesn't work well because the Multichannel should make two changes:
1) change the list to a text with buttons.
2) change the buttons for quick replies. (here fails)

Second approach make the plugin-flow-builder  not to use specific WhatsApp messages when the channel is not WhatsApp. This way when the message arrives to the Multichannel it will be a normal text message and multichannel only apply change in buttons (character limits, change buttons for quick replies), markdown 

Captura de pantalla 2025-08-06 a las 18 51 19

Captura de pantalla 2025-08-06 a las 18 52 10

- With this second approach works as expected. In Facebook recibe a text message with 4 quick replies when bot send a WhastAppButtonList

Captura de pantalla 2025-08-14 a las 13 05 14

### DIFF
--- a/packages/botonic-plugin-flow-builder/CHANGELOG.md
+++ b/packages/botonic-plugin-flow-builder/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to Botonic will be documented in this file.
 ### Changed
 
 - [PR-3091](https://github.com/hubtype/botonic/pull/3091): plugin-flow-builder not to use specific WhatsApp messages when the channel is not WhatsApp.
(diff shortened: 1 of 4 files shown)


vanbasten17 (approved):

asastre (approved):
👏🏼

PR #3096, opened 2025-09-01
botonic-cli & dx-bundler-rspack: change how bot config is created and cli read it #BLT-1878

Iru89:
## Description

- update BotConfig.get function to get bot-config.js after build from dist folder and remove it before the deploy
- update rspack/baseline with build process for bot-config.ts file


packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -53,6 +53,8 @@ export class BotConfig {
       botonicCli.match(botonicCliVersionRegex)?.[1] || ''
 
     const configLoaded = await this.loadBotConfig(appDirectory)
+    fs.rm(path.join(appDirectory, 'dist', 'bot-config.js'))
+    fs.rm(path.join(appDirectory, 'dist', 'bot-config.js.map'))
```
vanbasten17: sourcemap not needed if we only need this to generate the config

packages/botonic-dx-bundler-rspack/baseline/rspack.config.ts
```diff
@@ -360,6 +362,32 @@ function botonicServerConfig(mode: string): Configuration {
   }
 }
 
+function botonicBotConfig(mode: Mode): Configuration {
+  return {
+    optimization: {
+      sideEffects: true, // critical so that tree-shaking discards browser code from @botonic/react
+      minimize: true,
+      minimizer: mode === Mode.production ? [minimizerPlugin] : [],
+    },
```
vanbasten17: set minimize to false, as we are going to remove this file in any case and we can save some time

packages/botonic-cli/src/util/bot-config.ts
Iru89: applied

packages/botonic-dx-bundler-rspack/baseline/rspack.config.ts
Iru89: applied

packages/botonic-cli/CHANGELOG.md
```diff
@@ -10,12 +10,14 @@ All notable changes to Botonic will be documented in this file.
...
   </summary>
   
-## [0.37.x] - 2025-mm-dd
+## [0.39.x] - 2025-mm-dd
 
 ### Added
 
 ### Changed
 
+- [PR-3096](https://github.com/hubtype/botonic/pull/3096): botonic deploy get bot config from after build from dist folder and remove it before the deploy.
```
asastre: It looks like this phrase was rewriten?
```suggestion
- [PR-3096](https://github.com/hubtype/botonic/pull/3096): get bot config after the build from the `dist` folder and remove it before the deploy.
```

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -53,6 +53,7 @@ export class BotConfig {
       botonicCli.match(botonicCliVersionRegex)?.[1] || ''
 
     const configLoaded = await this.loadBotConfig(appDirectory)
+    fs.rm(path.join(appDirectory, 'dist', 'bot-config.js'))
```
asastre: Doubt: What happens if `@botonic/cli` is updated to the latest version and the `rspack.config.js` file is not updated with the needed changes? Will this removal fail and exit the deploy process?

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -71,34 +72,20 @@ export class BotConfig {
...
+      const botConfig = require(path.resolve(configPath))
       return {
         tools: botConfig?.tools || [],
         payloads: botConfig?.payloads || [],
         webviews: botConfig?.webviews || [],
       }
     } catch (error) {
       console.warn(
-        'Error loading src/bot-config.ts this file is not required but is used to share config with flow builder frontend'
+        `Error loading dist/bot-config.js this file is not required but is used to share config with flow builder frontend.
```
asastre: Nit: 
```suggestion
        `Error loading dist/bot-config.js. This file is not required but is used to share config with flow builder frontend. 
```

packages/botonic-dx-bundler-rspack/baseline/rspack.config.ts
```diff
@@ -360,6 +362,30 @@ function botonicServerConfig(mode: string): Configuration {
...
 
+function botonicBotConfig(mode: Mode): Configuration {
+  return {
+    optimization: {
+      minimize: false,
+    },
+    context: ROOT_PATH,
+    devtool: false,
+    target: 'node',
+    entry: path.resolve(WEBPACK_ENTRIES_DIRNAME, WEBPACK_ENTRIES.BOT_CONFIG),
```
asastre: Change: Since we're using rspack and not webpack, the dirname should be changed from `webpack-entries` to `rspack-entries`.

asastre (changes requested):
Some small changes and doubts before approving.

packages/botonic-cli/src/util/bot-config.ts
Iru89: I have checked it, if this file does not exist it gives an error and the deploy is not done. I refactor the code to remove the file inside the try catch. This way if it fails we will see the log but the deploy will work the same.

PR #3101, opened 2025-09-18
fix: payloads coming from AI Agents structured outputs to work with current flow

vanbasten17:
## Description
* Fix payloads coming from generated buttons with AI Agents to work with flow builder
* AI Generated buttons to work in whatsapp.


Davidhidalgo (approved):

packages/botonic-plugin-ai-agents/src/structured-output/carousel.ts
```diff
@@ -14,7 +14,8 @@ export const CarouselSchema = z
           image: z.string(),
           button: z.object({
             text: z.string(),
-            url: z.string(),
+            url: z.string().nullable().optional(),
+            // TODO: Add payload?
```
asastre: I would add the payload in Carousels as well.

PR #3118, opened 2025-10-29
feat(plugin-flow-builder): Boolean condition of a bot variable conditional on a non boolean variable

asastre:
## Description
Allow a Bot variable conditional to execute a boolean condition over a non boolean variable so it returns `true` when the variable exists and have any value and returns `false` when it is `undefined` or `null`.

## Context
In Flow Builder you can't check if a variable has a value or not because the conditions are checking only their values (string, number or boolean), but using the boolean condition on non boolean variables should check if it has a value (`true`) or not (`false`).

@Iru89 In the future this could be reviewed and refactored to extract this logic into a new condition type to check only if a variable exists or not.

## Testing

The pull request has integration tests.

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/functions/conditional-bot-variable.ts
+++ b/packages/botonic-plugin-flow-builder/src/functions/conditional-bot-variable.ts
@@ -2,9 +2,9 @@ import { ActionRequest } from '@botonic/react'
 
 import { getValueFromKeyPath } from '../utils'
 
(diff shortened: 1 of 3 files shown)


vanbasten17 (approved):

asastre:
Closing this PR in favour of [this other one](https://github.com/hubtype/botonic/pull/3132) that contains all the latest changes done in the Flow Builder plugin.

PR #3198, opened 2026-04-13
plugin-flow-builder: use user locale to resolve flow locale

Iru89:
## Description

Create a class to wrap logic to resolve flow locale using userLocale as a priority or more specific.

Rules:
- If the systemLocale is `es-MX` and the userLocale is `es`, we keep `es-MX` (more specific locale is preserved).
- If the systemLocale is `es` and the userLocale is `es-MX`, we update to `es-MX` (more specific locale takes priority).
- If the systemLocale is a `es-ES` and the userLocale is a generic language different for example `en` the one in the locale, we update to the generic language.

## Context

Previously, only the system locale was used to determine which of the available languages in the workflow would be used to respond to the user. Now, user.locale is checked first, and if this does not match, the old logic is used instead.

## Testing

Add tests to cover new FlowLocale class

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/utils/flow-locale.ts
+++ b/packages/botonic-plugin-flow-builder/src/utils/flow-locale.ts
@@ -19,7 +19,7 @@ export class FlowLocale {
     }
 
     const defaultLocale = this.resolveAsDefaultLocale()
-    this.botContext.setSystemLocale(defaultLocale)
+    this.setSystemLocale(defaultLocale)
     return defaultLocale
   }
 
@@ -28,13 +28,13 @@ export class FlowLocale {
 
     const resolvedUserLocale = this.resolveAsLocale(userLocale)
     if (resolvedUserLocale) {
-      this.botContext.setSystemLocale(resolvedUserLocale)
+      this.setSystemLocale(resolvedUserLocale)
       return resolvedUserLocale
     }
 
     const resolvedUserLanguage = this.resolveAsLanguage(userLocale)
     if (resolvedUserLanguage) {
-      this.botContext.setSystemLocale(resolvedUserLanguage)
+      this.setSystemLocale(resolvedUserLanguage)
       return resolvedUserLanguage
     }
 
@@ -51,7 +51,7 @@ export class FlowLocale {
 
     const language = this.resolveAsLanguage(systemLocale)
     if (language) {
-      this.botContext.setSystemLocale(language)
+      this.setSystemLocale(language)
       return language
     }
     return undefined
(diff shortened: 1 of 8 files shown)


vanbasten17 (approved):

Davidhidalgo (approved):

packages/botonic-plugin-flow-builder/src/utils/flow-locale.ts
```diff
@@ -0,0 +1,90 @@
+import type { BotContext } from '@botonic/core'
```
vanbasten17, line 1: i found interesting this proposal to make it more readable (as this is a very important point from now on):
```
import type { BotContext } from '@botonic/core'

export class FlowLocale {
  constructor(
    private readonly botContext: BotContext,
    private readonly flowLocales: string[],
    private readonly defaultLocaleCode: string
  ) {}

  resolve(): string {
    const priorityLocale = this.getPriorityLocale()

    if (priorityLocale) {
      const exactMatch = this.matchExactLocale(priorityLocale)
      if (exactMatch) return this.applyLocale(exactMatch)

      const languageMatch = this.matchLanguage(priorityLocale)
      if (languageMatch) return this.applyLocale(languageMatch)
    }

    return this.applyLocale(this.getDefaultLocale())
  }

  /**
   * Rules:
   * - If user and system languages differ, user locale takes priority.
   * - If both share the same language, the more specific locale wins.
   * - If both have the same specificity, user locale wins.
   */
  private getPriorityLocale(): string | undefined {
    const userLocale = this.botContext.getUserLocale()
    const systemLocale = this.botContext.getSystemLocale()

    if (!userLocale || !systemLocale) return undefined

    const userLanguage = this.getLanguage(userLocale)
    const systemLanguage = this.getLanguage(systemLocale)

    if (userLanguage !== systemLanguage) return userLocale

    const userIsSpecific = this.isSpecificLocale(userLocale)
    const systemIsSpecific = this.isSpecificLocale(systemLocale)

    if (userIsSpecific && !systemIsSpecific) return userLocale
    if (!userIsSpecific && systemIsSpecific) return systemLocale

    return userLocale
  }

  private matchExactLocale(locale: string): string | undefined {
    return this.flowLocales.includes(locale) ? locale : undefined
  }

  private matchLanguage(locale: string): string | undefined {
    const language = this.getLanguage(locale)
    return this.flowLocales.includes(language) ? language : undefined
  }

  private getDefaultLocale(): string {
    return this.defaultLocaleCode || 'en'
  }

  private applyLocale(locale: string): string {
    this.botContext.setSystemLocale(locale)
    this.botContext.session.user.system_locale_updated = true
    return locale
  }

  private getLanguage(locale: string): string {
    return locale.split('-')[0]
  }

  private isSpecificLocale(locale: string): boolean {
    return locale.includes('-')
  }
}
```

vanbasten17 (approved):

packages/botonic-plugin-flow-builder/src/utils/flow-locale.ts
Iru89: applied

PR #3199, opened 2026-04-15
plugin-ai-agents: add an await when call track LLM runs

Iru89:
## Description

Add an await when call tracking LLM runs

### DIFF
--- a/packages/botonic-plugin-ai-agents/src/guardrails/input.ts
+++ b/packages/botonic-plugin-ai-agents/src/guardrails/input.ts
@@ -53,7 +53,7 @@ export function createInputGuardrail(
       const result = await runner.run(agent, [lastMessage], { context })
       const endTime = Date.now()
 
-      void sendGuardrailLlmRunTracking(
+      await sendGuardrailLlmRunTracking(
         result,
         trackingContext,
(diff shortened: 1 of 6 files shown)


packages/botonic-plugin-ai-agents/src/guardrails/input.ts
```diff
@@ -97,7 +97,10 @@ async function sendGuardrailLlmRunTracking(
   const durationPerCall = Math.round(totalDuration / rawResponses.length)
   const temperature =
     (llmConfig.modelSettings.temperature as number | undefined) ?? 0
-  const apiVersion = OPENAI_PROVIDER === 'azure' ? AZURE_OPENAI_API_VERSION : ''
+  const apiVersion =
```
vanbasten17: i would create a helper function resolveApiVersion and use it in both of them and to throw if there is no apiVersion

vanbasten17 (approved):

packages/botonic-plugin-ai-agents/src/guardrails/input.ts
Iru89: applied

PR #3211, opened 2026-05-13
botonic-react: Assure urls are always opened via _self on WebView apps

askalatskiy:
Stop defaulting url buttons to _blank when target is omitted; use the explicit target prop so callers control window.open behavior.

## Description

We were facing an issue on Alsa that in their app (which is implemented via WebView) the url buttons generated via AI Agents wasn't opening.

With @asastre we found that the issue was that when they generated via AI Agent by default they have `_blank` as default, and was skipping the `isInWebviewApp` check.

- Assure that always when a button is rendered on a WebView app it sets the target to `_self` no matters what comes set from upper level. 

## Testing

The pull request...

- has unit tests
- has integration tests

### DIFF
--- a/packages/botonic-react/src/components/button/index.tsx
+++ b/packages/botonic-react/src/components/button/index.tsx
@@ -44,8 +44,8 @@ export const Button = (props: ButtonProps) => {
             payload: props.payload,
           })
     } else if (props.url) {
-      const defaultTarget = isInWebviewApp() ? '_self' : '_blank'
-      window.open(props.url, props.target || defaultTarget)
+      const target = isInWebviewApp() ? '_self' : props.target
+      window.open(props.url, target)
     }
 
     if (props.onClick) {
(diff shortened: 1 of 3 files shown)


asastre (approved):

packages/botonic-react/src/components/button/index.tsx
```diff
@@ -44,8 +44,8 @@ export const Button = (props: ButtonProps) => {
             payload: props.payload,
           })
     } else if (props.url) {
-      const defaultTarget = isInWebviewApp() ? '_self' : '_blank'
-      window.open(props.url, props.target || defaultTarget)
+      const target = isInWebviewApp() ? '_self' : props.target
```
Iru89: Maybe we should check whether `props.target` is undefined, and if it is, use `_blank`. If not, you can change the behaviour of the webchat buttons.

```suggestion
      const defaultTarget = props.target || '_blank'
      const target = isInWebviewApp() ? '_self' : defaultTarget
```

asastre: @Iru89 the default value for the target in `window.open` is `_blank` (https://developer.mozilla.org/en-US/docs/Web/API/Window/open), so I think we won't change anything if nothing is passed.

Iru89 (approved):

vanbasten17 (approved):

PR #3218, opened 2026-05-21
plugin-ai-agents: add LiteLLM provider support #BLT-2353

Davidhidalgo:
## Summary

- Replace static `OPENAI_PROVIDER`/`OPENAI_MODEL`/`OPENAI_API_KEY` constants with dynamic provider selection driven by `LITELLM_API_URL` in `BotSettings`
- Add LiteLLM OpenAI-compatible client with `x-litellm-tags` header (`bot_id`, `org_id`) for per-organisation request tracking in LiteLLM dashboards
- Add `modelSettings?: Record<string, unknown>` to `BotSettings` as a fallback for unrecognised model names (required for arbitrary LiteLLM-proxied models like Claude, Gemini, etc.)
- Remove unused `verbosity` param from `LLMConfig`; provider and API version now resolved via `llmConfig.getProviderName()` / `llmConfig.getApiVersion()`
- Rewrite `llm-config.test.ts` to cover all new paths (14 tests)

## Test plan

- [ ] All `llm-config` tests pass (`npm test` inside `packages/botonic-plugin-ai-agents`)
- [ ] LiteLLM client sends `x-litellm-tags: bot_id:<id>,org_id:<id>` header on requests
- [ ] Azure path unaffected (no `defaultHeaders`, same deployment/apiVersion behaviour)
- [ ] Unrecognised model name falls back to `BotSettings.modelSettings` when set, throws otherwise

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/botonic-core/src/models/legacy-types.ts
```diff
@@ -347,6 +347,7 @@ export interface BotSettings {
   AZURE_OPENAI_API_VERSION: string
   LANGUAGE_DETECTION_ENABLED?: boolean
   CUSTOM_SHORT_URL_HOST: string | null
+  modelSettings?: Record<string, unknown>
```
vanbasten17: BotSettings is the contract received from the lambda — adding modelSettings there means it would need to be configured as an actual bot setting, which isn't the right place for this.
Maybe we should receive them from the backend as LITELLM_MODEL_SETTINGS ?

packages/botonic-plugin-ai-agents/src/runners/base-runner.ts
vanbasten17: do api_version: this.llmConfig.getApiVersion() if not used anymore in this body function. same in guardrail-input

packages/botonic-plugin-ai-agents/src/index.ts
```diff
@@ -132,7 +132,10 @@ export default class BotonicPluginAiAgents<
       this.maxRetries,
       this.timeout,
       aiAgentArgs.model,
-      aiAgentArgs.verbosity
+      botContext.settings,
+      botContext.secrets,
+      botContext.session.bot.id,
+      botContext.session.organization_id
```
vanbasten17: you can modify LLMConfig to just accept botContext and pick these values insde so you don't need all these positional arguments

packages/botonic-plugin-ai-agents/src/llm-config.ts
```diff
@@ -11,14 +11,17 @@ import {
...
-  OPENAI_MODEL,
-  OPENAI_PROVIDER,
 } from './constants'
 
+export type LLMProvider = 'litellm' | 'azure'
+
 export class LLMConfig {
   private readonly maxRetries: number
   private readonly timeout: number
+  private readonly settings: BotSettings
```
vanbasten17: initialize with botContext, keep it simple

vanbasten17: or pass an object

packages/botonic-plugin-ai-agents/src/llm-config.ts
```diff
@@ -40,6 +51,16 @@ export class LLMConfig {
     return await this.modelProvider.getModel(this.modelName)
   }
 
+  getProviderName(): LLMProvider {
+    return this.settings.LITELLM_API_URL ? 'litellm' : 'azure'
```
vanbasten17: don't use magic strings, better to have them isolated in an enum or constant object

packages/botonic-plugin-ai-agents/src/llm-config.ts
```diff
@@ -49,28 +70,42 @@ export class LLMConfig {
...
     }
 
     return this.getAzureClient()
   }
 
-  private getOpenAIClient(): OpenAI {
+  private buildLiteLLMTags(): { 'x-litellm-tags': string } | undefined {
+    const parts: string[] = []
+    if (this.botId) {
+      parts.push(`bot_id:${this.botId}`)
```
vanbasten17: put `bot_id` and `org_id` particle in constants

vanbasten17: + a separator

vanbasten17 (approved):

packages/botonic-plugin-ai-agents/src/index.ts
Iru89: I agree, it’s better to pass the entire botContext

packages/botonic-plugin-ai-agents/src/index.ts
```diff
@@ -132,7 +132,10 @@ export default class BotonicPluginAiAgents<
       this.maxRetries,
       this.timeout,
       aiAgentArgs.model,
-      aiAgentArgs.verbosity
```
Iru89, line 139: This value is the Verbosity selected in the Flow Builder frontend, you cannot delete it

packages/botonic-plugin-ai-agents/src/index.ts
```diff
@@ -193,7 +196,10 @@ export default class BotonicPluginAiAgents<
...
+      botContext.settings,
+      botContext.secrets,
+      botContext.session.bot.id,
+      botContext.session.organization_id
```
Iru89: Same here, better pass all botContext

packages/botonic-plugin-ai-agents/src/llm-config.ts
```diff
@@ -11,14 +11,17 @@ import {
...
 
+export type LLMProvider = 'litellm' | 'azure'
+
 export class LLMConfig {
   private readonly maxRetries: number
   private readonly timeout: number
+  private readonly settings: BotSettings
+  private readonly secrets: BotSecrets
+  private readonly botId?: string
+  private readonly orgId?: string
```
Iru89: As Marc said, it’s best to save the entire botContext and then retrieve the values you need wherever you need them

```suggestion
  private readonly botContext: BotContext
```

packages/botonic-plugin-ai-agents/src/index.ts
```diff
@@ -193,7 +196,10 @@ export default class BotonicPluginAiAgents<
       this.maxRetries,
       this.timeout,
       aiAgentArgs.model,
-      aiAgentArgs.verbosity
```
Iru89, line 200: Same here this is de verbosity set in Flow Builder Frontend

Iru89 (changes requested):
You are removing the “verboisty” value that was being passed to the LLMConfig configured in the Flow Builder frontend.

packages/botonic-plugin-ai-agents/src/index.ts
Davidhidalgo: Done —  now accepts  as the last argument and extracts , , , and  internally. The positional , , ,  params are gone.

packages/botonic-plugin-ai-agents/src/llm-config.ts
Davidhidalgo: Done — `LLMConfig` now stores `private readonly botContext: BotContext` and extracts all needed values internally.

Davidhidalgo: Done — `LLMConfig` now stores `private readonly botContext: BotContext` and extracts all needed values internally.

packages/botonic-plugin-ai-agents/src/llm-config.ts
Davidhidalgo: Done — added `LLM_PROVIDERS = { LITELLM: 'litellm', AZURE: 'azure' } as const` to `constants.ts` and updated `getProviderName()` and `getApiVersion()` to use it.

packages/botonic-plugin-ai-agents/src/llm-config.ts
Davidhidalgo: Done — added `LITELLM_TAG_KEYS = { BOT_ID: 'bot_id', ORG_ID: 'org_id', SEPARATOR: ',' } as const` to `constants.ts`. `buildLiteLLMTags()` now uses these constants for both the key names and the join separator.

Davidhidalgo: Done — `LITELLM_TAG_KEYS.SEPARATOR` is used in `parts.join(...)`.

packages/botonic-plugin-ai-agents/src/runners/base-runner.ts
Davidhidalgo: Done — removed the `const apiVersion` variable and inlined `this.llmConfig.getApiVersion()` directly in the map, same in `guardrails/input.ts`. Also changed `getApiVersion()` to return `string | undefined` (returns `undefined` for LiteLLM instead of the placeholder string), and made `LlmRunData.api_version` optional.

packages/botonic-plugin-ai-agents/src/llm-config.ts
Davidhidalgo: Done — applied the suggestion.

packages/botonic-plugin-ai-agents/src/index.ts
Davidhidalgo: Done.

packages/botonic-plugin-ai-agents/src/index.ts
Davidhidalgo: Done — both `executeWorkerAIAgent` and `executeRouterAIAgent` now pass `botContext` directly.

packages/botonic-plugin-ai-agents/src/index.ts
Davidhidalgo: Already restored in commit `0e19aacb` — `aiAgentArgs.verbosity` is passed to `LLMConfig` in both call sites.

packages/botonic-plugin-ai-agents/src/index.ts
Davidhidalgo: Same as above — already restored in commit `0e19aacb`.

packages/botonic-core/src/models/legacy-types.ts
Davidhidalgo: `modelSettings` was already removed from `BotSettings` in a previous commit. It is not a bot setting — it is derived inside `LLMConfig` from the model name and verbosity, so it stays fully internal to the plugin.

packages/botonic-plugin-ai-agents/src/runners/base-runner.ts
vanbasten17: i meant to call it inlined, no need of caching apiVersion if not used anywhere else:
```
api_version: this.llmConfig.getApiVersion()
```

vanbasten17 (approved):

Iru89 (approved):

PR #3224, opened 2026-05-27
fix: input guardrail handling in BaseRunner and RouterRunner

Iru89:
## Description

This pull request fixes the AI router response when an input guardrail blocks a request.
Router runs now preserve router-specific metadata, including the router agent name and available specialists, even when the OpenAI Agents SDK raises an input guardrail tripwire before a normal run result is built.

## Context

When an input guardrail is triggered, `BaseRunner` builds an early `RunResult` from the exception. Before this change, that fallback result used empty agent names and no specialist list, so router-based AI agent responses could lose important routing context.

This affected consumers such as Flow Builder, where the router response should still identify the router agent as the starting and last agent when no specialist transfer happened.

The input guardrail tripwire handling has been moved into a protected `BaseRunner` method so specialized runners can reuse and customize the fallback result.

`RouterRunner` now overrides that handler to enrich guardrail-triggered results with router metadata:

- `startingAgentName` and `lastAgentName` are set to the router agent name.
- `availableSpecialists` is derived from the router handoffs.
- `isTransferredToSpecialist` remains `false` because the guardrail stops the run before a handoff.

The existing specialist extraction logic was also centralized in `RouterRunner#getAvailableSpecialists()` so normal router results and guardrail-triggered results stay consistent.

Flow Builder now falls back to the router content field name for `startingAgentName` and `lastAgentName` when an AI agent response does not provide those values.

### DIFF
--- a/packages/botonic-plugin-flow-builder/src/content-fields/flow-ai-agent-router.tsx
+++ b/packages/botonic-plugin-flow-builder/src/content-fields/flow-ai-agent-router.tsx
@@ -165,8 +165,8 @@ export class FlowAiAgentRouter extends FlowAiAgentBase {
         this.aiAgentResponse?.inputGuardrailsTriggered ?? [],
       outputGuardrailsTriggered:
         this.aiAgentResponse?.outputGuardrailsTriggered ?? [],
-      startingAgentName: this.aiAgentResponse?.startingAgentName ?? '',
-      lastAgentName: this.aiAgentResponse?.lastAgentName ?? '',
+      startingAgentName: this.aiAgentResponse?.startingAgentName ?? this.name,
+      lastAgentName: this.aiAgentResponse?.lastAgentName ?? this.name,
       availableSpecialists: this.aiAgentResponse?.availableSpecialists ?? [],
       isTransferredToSpecialist,
     }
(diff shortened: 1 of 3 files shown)


Davidhidalgo (approved):

vanbasten17 (approved):

packages/botonic-plugin-flow-builder/src/content-fields/flow-ai-agent-router.tsx
```diff
@@ -165,8 +165,8 @@ export class FlowAiAgentRouter extends FlowAiAgentBase {
         this.aiAgentResponse?.inputGuardrailsTriggered ?? [],
       outputGuardrailsTriggered:
         this.aiAgentResponse?.outputGuardrailsTriggered ?? [],
-      startingAgentName: this.aiAgentResponse?.startingAgentName ?? '',
-      lastAgentName: this.aiAgentResponse?.lastAgentName ?? '',
+      startingAgentName: this.aiAgentResponse?.startingAgentName ?? this.name,
+      lastAgentName: this.aiAgentResponse?.lastAgentName ?? this.name,
```
vanbasten17, line 169: use `||`?

PR #3228, opened 2026-05-28
feat(BLT-2353): add LiteLLM provider support

Davidhidalgo:
## Summary

- Adds `LITELLM` to `LLM_PROVIDERS` and a new `LITELLM_TAG_KEYS` constant (`bot_id`, `org_id`, separator)
- Detects LiteLLM via `botContext.settings.LITELLM_API_URL` (platform injection) or `LLM_API_URL` env var (local dev) — no new env var needed
- Routes LiteLLM requests through the OpenAI client with a custom `baseURL` and `LITELLM_API_KEY`
- Attaches `x-litellm-tags: bot_id:<id>,org_id:<id>` header on every request for per-org tracking in the LiteLLM dashboard
- Returns `undefined` for `api_version` when provider is LiteLLM (no API version concept)
- Falls back to reasoning model settings for unknown models (supports arbitrary proxied models like Claude, Gemini, etc.)

## Test plan

- [ ] 29 tests added in `llm-config.test.ts`: provider detection (via botContext + env), LiteLLM client config, `x-litellm-tags` variations (both/one/none), `getApiVersion` returns `undefined` for LiteLLM, model settings fallback for unknown models
- [ ] All 118 tests pass (`npx jest --no-coverage`)
- [ ] Biome check clean on modified files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/botonic-plugin-ai-agents/src/services/types.ts
+++ b/packages/botonic-plugin-ai-agents/src/services/types.ts
@@ -17,7 +17,7 @@ export interface LlmRunData {
   deployment_name: string
   model_name: string
   feature: string
-  api_version: string | undefined
+  api_version?: string
   num_prompt_tokens: number
   num_completion_tokens: number
   duration_in_milliseconds: number
--- a/packages/botonic-plugin-ai-agents/src/services/types.ts
+++ b/packages/botonic-plugin-ai-agents/src/services/types.ts
@@ -17,7 +17,7 @@ export interface LlmRunData {
   deployment_name: string
   model_name: string
   feature: string
-  api_version: string
+  api_version: string | undefined
   num_prompt_tokens: number
   num_completion_tokens: number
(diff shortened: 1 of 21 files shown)


packages/botonic-plugin-ai-agents/src/services/types.ts
```diff
@@ -17,7 +17,7 @@ export interface LlmRunData {
   deployment_name: string
   model_name: string
   feature: string
-  api_version: string
+  api_version: string | undefined
```
vanbasten17: do api_version?: string instead

vanbasten17 (approved):

packages/botonic-plugin-ai-agents/src/services/types.ts
Davidhidalgo: done

PR #3234, opened 2026-06-15
fix: update variable pattern and improve variable replacement logic

Iru89:
## Description

This pull request fixes variable replacement in Flow Builder content and tightens the variable matching regex to avoid unsafe patterns on uncontrolled text. It also aligns WhatsApp button list processing with other content types and ensures capture-user-input tracking uses the resolved text or audio transcript.

## Context

Flow Builder replaces placeholders such as `{name}` with values from the bot context. The previous implementation had [two issues](https://github.com/hubtype/botonic/security/code-scanning/33):

1. **Variable replacement logic** — `replaceVariables` collected matches and replaced them one by one with `String.replace(match, ...)`. When multiple variables appeared in the same string, only the first occurrence of each unique match was reliably replaced, and edge cases with repeated or overlapping patterns could produce incorrect output.
2. **Variable regex** — The pattern `/{([^}]+)}/g` could match across nested or malformed brace sequences in user-controlled content, which is both incorrect for rendering and risky when applied to uncontrolled input.

EXTRA: Not related to the issue, but detected in this PR

3. **WhatsApp button list** — `FlowWhatsappButtonList` did not call `filterContent` during `processContent`, unlike other content field types, so variable-based filtering was skipped for that node.
4. **Capture user input tracking** — Analytics events used `request.input.data` directly instead of `textOrTranscript`, so audio messages were tracked with empty or incorrect input values.

### Variable pattern (`constants.ts`)

The global variable pattern was updated from `/{([^}]+)}/g` to `/{([^{}]+)}/g`. The new pattern only matches a single level of braces and excludes nested `{` or `}` inside the placeholder, preventing false matches on sequences like `{{|{{|{{|{{|`.

### Variable replacement (`content-fields-base.ts`)

`replaceVariables` now uses `text.replace(VARIABLE_PATTERN_GLOBAL, callback)` so every match is processed in a single pass. The callback resolves the key path, keeps access-token placeholders unchanged, and returns the stringified value when valid or the original match otherwise. This correctly handles multiple variables in one message and removes the previous TODO about stale values on repeated renders.

### WhatsApp button list (`flow-whatsapp-button-list.tsx`)

`processContent` now calls `filterContent(botContext, this)` after tracking, consistent with `FlowText`, `FlowCarousel`, and other content field implementations.

### Capture user input (`capture-user-input-api.ts`)

The `CaptureUserInput` tracking event now sends `this.textOrTranscript` as `userInput`, which is the same value used for capture and API calls, including transcribed audio input.

### Release

Version bumped to `0.50.1` with an entry in `CHANGELOG.md`.

No API changes. Existing Flow Builder flows that use `{variable}` placeholders in text, buttons, or WhatsApp button lists benefit automatically.

Example — multiple variables in one text node:

```ts
// Bot context: { name: 'Ada', email: 'ada@example.com' }
// Template: 'Hello {name}, your email is {email}'
// Result:   'Hello Ada, your email is ada@example.com'
```

Example — literal braces are preserved:

```ts
// Template: '{{|{{|{{|{{| Hello {name}'
// Result:   '{{|{{|{{|{{| Hello Ada'
```

## Testing

The pull request...

- [x] has unit tests


packages/botonic-plugin-flow-builder/src/content-fields/content-fields-base.ts
```diff
@@ -63,25 +63,19 @@ export abstract class ContentFieldsBase {
   }
 
   replaceVariables(text: string, botContext: BotContext): string {
-    const matches = text.match(VARIABLE_PATTERN_GLOBAL)
-
-    let replacedText = text
-    if (matches && botContext) {
-      matches.forEach(match => {
+    if (botContext) {
```
vanbasten17: i would invert the condition now the "replacedText" is removed, so:
```
text: string --> rawText
if !botContext { return rawText}
return rawText.replace...
```
and is more explicit and readable

vanbasten17 (approved):
as this is small, i would apply it also in botonic's hubtype-product

PR #3236, opened 2026-06-18
feat: add new custom conditional v2

Iru89:
## Description

Adds support for the new **Custom Condition V2** flow node (`custom-condition`) in `botonic-plugin-flow-builder`. The node evaluates a bot variable (string, number, or boolean) against a list of declarative conditions with operators, resolves the last matching condition's target, and tracks the result via the existing `ConditionalCustom` analytics event.

## Context

The previous custom conditional implementation (`FlowCustomConditional`) relies on function-based result mapping and exact variable matching, which limits expressiveness and is harder to maintain. Flow Builder now exposes a V2 node with typed operators (e.g. `contains`, `greaterThan`, `between`, `isTruthy`) that must be evaluated on the Botonic side.

This PR implements the runtime evaluation logic so bots can route flows based on rich conditional rules defined in the CMS, without custom function code per condition.

The implementation is split into three focused modules and one flow content class:

1. **`hubtype-fields/custom-conditional-v2.ts`** — Type definitions for the CMS payload: `VariableFormat`, condition operators, and `HtCustomConditionalV2Node`.
2. **`custom-conditional-v2-evaluators.ts`** — Pure evaluation functions per variable type and operator. Uses lookup tables (`STRING_CONDITION_EVALUATORS`, `NUMBER_CONDITION_EVALUATORS`, `BOOLEAN_CONDITION_EVALUATORS`) for clarity and testability.
3. **`custom-conditional-v2-resolver.ts`** — Shared resolution logic:
   - Iterates conditions in order and keeps the **last** match (same semantics as sequential if/else-if chains).
   - Skips conditions without a `target`.
   - Falls back to `default_target` when no condition matches.
   - Derives `customResult` from the matched condition value (or `'default'` / `'false'` for fallbacks).
4. **`flow-custom-condition-v2.tsx`** — `FlowCustomConditionalV2` content field. Resolves the follow-up target at construction time (in `fromHubtypeCMS`) by reading the variable from `key_path` via `getValueFromKeyPath`, then delegates to the evaluator and resolver modules.

The node is registered in `FlowFactory` under `HtNodeWithContentType.CUSTOM_CONDITION` and exported from the content-fields index.

### Supported operators

| Type    | Operators |
|---------|-----------|
| String  | `contains`, `notContains`, `equalsTo`, `notEqualsTo`, `startsWith`, `notStartsWith`, `endsWith`, `notEndsWith` |
| Number  | `equalsTo`, `notEqualsTo`, `greaterThan`, `lessThan`, `greaterThanOrEqualTo`, `lessThanOrEqualTo`, `between`, `notBetween` |
| Boolean | `isTruthy` |

## Testing

The pull request...

- [x] has unit tests
- [ ] has integration tests

Unit tests in `tests/custom-conditional-v2.test.ts` cover:

- All string operators with expected `customResult` and `followUp` resolution.
- All number operators, including `between` / `notBetween` with `min` / `max`.
- Boolean `isTruthy` for `true` and default fallback for `false`.
- Default target usage when no condition matches (string, number, and boolean).


packages/botonic-plugin-flow-builder/src/content-fields/hubtype-fields/nodes.ts
```diff
@@ -3,6 +3,7 @@ import type { HtAiAgentRouterNode } from './ai-agent-router'
 import type { HtBotActionNode } from './bot-action'
 import type { HtCaptureUserInputNode } from './capture-user-input'
 import type { HtCarouselNode } from './carousel'
+import type { HtCustomConditionalV2Node } from './custom-conditional-v2'
 // import { HtChannelConditionalNode } from './channel-conditional'
 // import { HtCountryConditionalNode } from './country-conditional'
 // import { HtCustomConditionalNode } from './custom-conditional'
```
vanbasten17, line 9: can referenced commented code be removed then?

packages/botonic-plugin-flow-builder/tests/custom-conditional-v2.test.ts
```diff
@@ -0,0 +1,359 @@
+import { INPUT } from '@botonic/core'
```
vanbasten17, line 1: again unexpected .test.ts

vanbasten17 (approved):

packages/botonic-plugin-flow-builder/tests/custom-conditional-v2.test.ts
Iru89: In this repo, we've always used `.test.ts` 😅

Davidhidalgo (approved):

PR #3239, opened 2026-06-29
feat(botonic-cli): add variables in bot config in each deployment #BLT-2449 

Iru89:
## Description

This pull request extends the bot deployment payload with a new `variables` field in `BotConfigJSON`, allowing bots to declare session variable definitions that are sent to the Hubtype platform on every deploy. Bot developers can now list the variables their bot uses (for example, `session.user.locale`) in `bot-config.ts`, and the Botonic CLI will include them automatically in the deployment bundle metadata.

## Context

Flow Builder and the Hubtype platform need to know which bot session variables exist and how they are typed, so they can expose them in the Flow Builder UI and align runtime behavior with bot configuration. Previously, `BotConfigJSON` only carried tools, payloads, and webviews. There was no standardized way for a bot to declare its variable schema during deployment.

This change is part of **BLT-2449** and prepares the Botonic ecosystem for `bot_variable_definitions` support in Flow Builder data (`HtFlowBuilderData`).

### 1. Core types (`@botonic/core`)

- Added `VariableConfigJSON` with `key_path` (required) and optional `type` (`'string' | 'number' | 'boolean'`).
- Extended `BotConfigJSON` with an optional `variables` field so existing bots remain backward compatible.

### 2. Botonic CLI (`@botonic/cli`)

- Extended the CLI's internal `BotConfigJSON` interface with a required `variables: VariableConfigJSON[]` field in the deploy payload.
- Updated `BotConfig.loadBotConfig()` to read `variables` from `dist/bot-config.js` and default to `[]` when:
  - The field is missing or `undefined`
  - `dist/bot-config.js` cannot be loaded
- Refactored `BotConfig` from static methods to instance methods to improve testability (mocks can now spy on instance methods).
- Updated the deploy command to instantiate `BotConfig` instead of calling static `BotConfig.get()`.

### 3. Flow Builder plugin (`@botonic/plugin-flow-builder`)

- Added `HtBotVariableDefinition` and `bot_variable_definitions` to `HtFlowBuilderData` to type the variable definitions returned by the Hubtype API, including an `is_from_bot_config` flag to distinguish bot-declared variables from platform-defined ones.

### 4. Example

- Updated `examples/flow-builder-typescript/src/bot-config.ts` with sample variable declarations for locale-related session fields.

### Data flow

```
bot-config.ts (source)
    → build → dist/bot-config.js
    → BotConfig.loadBotConfig() reads variables
    → BotConfig.get() includes variables in deploy payload
    → Hubtype platform receives variable definitions on deploy
```

Bot developers can declare variables in their `bot-config.ts` file:

```typescript
import type { BotConfigJSON } from '@botonic/core'

export const botConfig: BotConfigJSON = {
  tools: [],
  payloads: [],
  webviews: [],
  variables: [
    { key_path: 'session.user.locale', type: 'string' },
    { key_path: 'session.user.country', type: 'string' },
    { key_path: 'session.user.extra_data.isPremium', type: 'boolean' },
    { key_path: 'session.user.extra_data.score' }, // type is optional
  ],
}
```

After building the bot (`dist/bot-config.js` must be generated), running `botonic deploy` will include the `variables` array in the deployment metadata. Bots that do not declare `variables` will send an empty array (`[]`).

## Testing

The pull request...

- [x] has unit tests


examples/flow-builder-typescript/src/bot-config.ts
```diff
@@ -12,4 +12,9 @@ export const botConfig: BotConfigJSON = {
   tools: aiAgentTools,
   payloads: [],
   webviews: [],
+  variables: [
+    { keyPath: 'session.user.locale', type: 'string' },
+    { keyPath: 'session.user.country', type: 'string' },
```
Davidhidalgo: **Bug: `keyPath` should be `key_path`.**

The `VariableConfigJSON` interface in `@botonic/core` declares the field as `key_path` (snake_case), but the example uses `keyPath` (camelCase). TypeScript will actually catch this with excess property checking — meaning the example does not compile as-is. More importantly, it is the wrong shape to show to bot developers: if they copy it, their variable definitions will be sent with the wrong field name and ignored by the platform.

packages/botonic-cli/tests/util/bot-config.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    })
+
+    const result = await botConfig.loadBotConfig('/any')
+
+    expect(result.variables).toEqual([])
+  })
+
+  it('returns declared variables intact when present', async () => {
+    const variables = [
+      { keyPath: 'session.user.locale', type: 'string' as const },
```
Davidhidalgo: **Same `keyPath` vs `key_path` mismatch as in the example.**

The mock returns objects with `keyPath`, but the actual `VariableConfigJSON` interface uses `key_path`. This test passes because `getBotConfig` is spied on as `any`, so TypeScript does not enforce the shape of the returned value. As a result, the *"returns declared variables intact"* test is verifying the wrong shape — production code receives `{ key_path: ... }` objects from `dist/bot-config.js`, but the assertion checks for `{ keyPath: ... }` ones.

Fix by using `key_path` in the fixture to match the actual contract.

packages/botonic-cli/src/util/bot-config.ts
```diff
@@ -31,29 +31,36 @@ interface WebviewConfigJSON {
   name: string
 }
 
+interface VariableConfigJSON {
```
Davidhidalgo, line 34: **`VariableConfigJSON` is duplicated from `@botonic/core`.**

`@botonic/core` already exports this type from its public barrel. Redefining it locally creates a maintenance trap: if the core definition gains new fields or renames `key_path`, this copy won't follow and the two will diverge silently.

There is already a subtle inconsistency: core uses `type?: VariableType` (an enum), while this copy uses `type?: 'string' | 'number' | 'boolean'` (a string union). Structurally equivalent today, but not guaranteed to stay that way.

Replace with:
```ts
import type { VariableConfigJSON } from '@botonic/core'
```

packages/botonic-plugin-flow-builder/src/content-fields/hubtype-fields/common.ts
```diff
@@ -17,6 +17,7 @@ export interface HtFlowBuilderData {
   flows: HtFlows[]
   webviews: HtFlowWebview[]
   campaigns: HtCampaign[]
+  bot_variable_definitions: HtBotVariableDefinition[]
 }
```
Davidhidalgo, line 21: **`bot_variable_definitions` should be optional.**

This field is required (`HtBotVariableDefinition[]`, no `?`), but bots deployed before this PR was shipped will not include it in the API response. Any `feature-flow-builder` code that reads `HtFlowBuilderData` will get `undefined` at runtime where a required array is expected, breaking existing bots.

Mark it optional until the backend guarantees a default or all existing deploy records are backfilled:
```ts
bot_variable_definitions?: HtBotVariableDefinition[]
```

packages/botonic-core/package.json
```diff
@@ -51,7 +51,8 @@
     "axios": "^1.15.2",
     "pako": "^2.1.0",
     "process": "^0.11.10",
-    "pusher-js": "^5.1.1"
+    "pusher-js": "^5.1.1",
+    "zod": "^4.4.3"
```
Davidhidalgo, line 55: **Adding Zod as a production dependency of `botonic-core` adds it to every bot's bundle.**

All bots depend on `@botonic/core`, so they will all bundle Zod at runtime even if they never call `getSessionVariables`. Zod 4 is not trivially tree-shakeable across all bundler configurations.

If `getSessionVariables` is meant as an optional helper for bot developers who already use Zod, consider one of:
- Moving it to a separate opt-in package (e.g. `@botonic/zod`), keeping `core` dependency-light.
- Declaring `zod` as a `peerDependency` / `optionalDependency`.
- Accepting a plain JSON Schema object instead of a `ZodType`, so the caller does the Zod conversion and `core` does not need the import.

packages/botonic-core/src/utils/bot-config.ts
```diff
@@ -0,0 +1,124 @@
...
+  { key_path: 'session.user.locale', type: VariableType.String },
+  {
+    key_path: 'session.user.country',
+    type: VariableType.String,
+  },
+  { key_path: 'session.user.system_locale', type: VariableType.String },
+]
+
+export function getSessionVariables(
+  extraDataSchema?: ZodType
```
Davidhidalgo, line 18: **`getSessionVariables` is exported but has no consumer in this PR.**

Is this intended as infrastructure for a follow-up PR (e.g. in `feature-flow-builder` or `plugin-flow-builder`)? If so, a short comment linking to the ticket would help reviewers confirm it is not dead code — and protect it from future cleanup passes.

If there is no imminent consumer, consider deferring the export until it is actually used. Unused exports in a published package become part of the public API surface and are harder to remove or rename later.

packages/botonic-cli/src/util/bot-config.ts
Iru89: I'm applying the change to the types so they're the same, but I don't want to add the @botonic/core dependency to the @botonic/cli

packages/botonic-core/src/utils/bot-config.ts
Iru89: This should be used in the bot's code to create the `variables` attribute within `botConfig`. This way, with each deployment, the feature-flow-builder will know which variables from the bot can be used.
It is used this way in `bot-config.ts` file:

```typescript
import type {
  BotConfigJSON,
  getSessionVariables
} from '@botonic/core'
import { getToolsForBotConfig } from '@botonic/plugin-ai-agents'

import { botConfigWebviews } from './client/webviews'
import { PAYLOADS } from './server/constants'
import { userDataSchema } from './server/domain/user-data'
import { customTools } from './server/tools'

const aiAgentTools = getToolsForBotConfig(customTools ?? [])
export const botSessionVariables = getSessionVariables(userDataSchema)

export const botConfig: BotConfigJSON = {
  tools: aiAgentTools,
  payloads: Object.values(PAYLOADS),
  webviews: botConfigWebviews,
  variables: botSessionVariables,
}

```

packages/botonic-plugin-flow-builder/src/content-fields/hubtype-fields/common.ts
Iru89: But the backend adds an empty array if it hasn't been saved

Davidhidalgo (approved):

PR #3259, opened 2026-07-28
feat(plugin-ai-agents): review that new models work for ai agents #BLT-2493 

Iru89:
## Description

Simplifies `LLMConfig.getModelSettings()` so new models routed through LiteLLM work without model-specific branches, and removes leftover debug `console.log` statements from the AI agents plugin.

## Context

Part of [BLT-2493](https://hubtype.atlassian.net/browse/BLT-2493) — review that new models work for AI agents.

The previous `LLMConfig.getModelSettings()` implementation hard-coded branches for `chat`, `luna`, `gpt-5` and `gpt-4` model name substrings, defaulting any unknown model to `ReasoningEffort.None`. That meant every new model added through LiteLLM needed an ad-hoc branch, and models that don't accept `reasoning.effort` could receive an invalid setting. Two debug `console.log` statements had also been left in the production path (`llm-config.ts` constructor and `getSpecialistAgentAndTools` in `index.ts`).

`LLMConfig.getModelSettings()` is reduced to three cases:

- `gpt-4*` models keep the existing deterministic settings: `temperature: 0` and `text.verbosity: VerbosityLevel.Medium`, ignoring both `verbosity` and `reasoningEffort`.
- For any other model, when `reasoningEffort` is provided by the caller, the settings are `{ reasoning: { effort: reasoningEffort }, temperature: 1, text: { verbosity } }`.
- For any other model without `reasoningEffort`, the settings fall back to `{ temperature: 1, text: { verbosity } }` with no `reasoning` block, letting the provider apply its own defaults.

This removes the `chat`/`luna`/`gpt-5`/fallback branches and the implicit `ReasoningEffort.None` default, so new LiteLLM-proxied models work out of the box and only opt into reasoning when explicitly requested.

Two debug `console.log` statements are removed from `llm-config.ts` and `index.ts`.

No public API changes. Behavior change callers may notice:

- Models whose name does not contain `gpt-4` no longer receive `reasoning: { effort: 'none' }` by default; they only get a `reasoning` block when `reasoningEffort` is explicitly passed to `LLMConfig`.

## Testing

The pull request...

- has unit tests

Updated `llm-config.test.ts` to cover the new `gpt-4*` (ignores verbosity and `reasoningEffort`) and non-`gpt-4*` (with/without `reasoningEffort`) branches.

[BLT-2493]: https://hubtype.atlassian.net/browse/BLT-2493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/botonic-plugin-ai-agents/src/llm-config.ts
+++ b/packages/botonic-plugin-ai-agents/src/llm-config.ts
@@ -1,4 +1,8 @@
-import { type BotContext, ReasoningEffort, VerbosityLevel } from '@botonic/core'
+import {
+  type BotContext,
+  type ReasoningEffort,
+  VerbosityLevel,
+} from '@botonic/core'
 import {
   type Model,
   type ModelProvider,
@@ -54,7 +58,6 @@ export class LLMConfig {
       verbosity,
       reasoningEffort
     )
-    console.log('new LLMConfig reasoningEffort', reasoningEffort)
   }
 
   async getModel(): Promise<Model> {
@@ -147,30 +150,6 @@ export class LLMConfig {
     verbosity: VerbosityLevel,
     reasoningEffort?: ReasoningEffort
   ): ModelSettings {
-    // By default, we use low reasoning effort for chat models because they not accept the reasoning effort set as none.
-    if (model.includes('chat')) {
-      return {
-        reasoning: { effort: reasoningEffort ?? ReasoningEffort.Low },
-        temperature: 1,
-        text: { verbosity },
-      }
-    }
-
-    if (model.includes('luna')) {
-      return {
-        temperature: 1,
-        text: { verbosity },
-      }
-    }
-
-    if (model.includes('gpt-5')) {
-      return {
-        reasoning: { effort: reasoningEffort ?? ReasoningEffort.None },
-        temperature: 1,
-        text: { verbosity },
-      }
-    }
-
     if (model.includes('gpt-4')) {
       return {
         temperature: 0,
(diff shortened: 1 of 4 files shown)


packages/botonic-plugin-ai-agents/src/llm-config.ts
```diff
@@ -147,40 +150,22 @@ export class LLMConfig {
...
-    if (model.includes('chat')) {
-      return {
-        reasoning: { effort: reasoningEffort ?? ReasoningEffort.Low },
-        temperature: 1,
-        text: { verbosity },
-      }
-    }
-
-    if (model.includes('luna')) {
+    if (model.includes('gpt-4')) {
```
vanbasten17, line 153: suggestion (potential improvements): I think that it might make sense to have a domain class HTModel like...
```
export class HTModel {

  constructor(private readonly value: string) {}

  get name(): string {

    return this.value

  }

  get isGPT4(): boolean {

    return this.value.includes('gpt-4')

  }

  get supportsReasoning(): boolean {

    return !this.isGPT4

  }

}
```
so we avoid spreading very specific logic within the code.
just a suggestion for improvement

vanbasten17 (approved):

packages/botonic-plugin-ai-agents/src/llm-config.ts
Iru89: I don't see a need for it right now; we'll probably remove gpt-4.1 soon

PR #3264, opened 2026-08-11
fix: avoid sending too much typing on/off events

vanbasten17:
## Summary

Webchat v1 was sending excessive `typing_on` / `typing_off` events to the platform. Each event triggers a full webhook POST, which became problematic once platform-side rate limiting was introduced.

This PR fixes the root cause and adds a network-level safety net so a typing session produces at most one `typing_on` and one `typing_off` webhook call.

## Problem

In `textarea.tsx`, typing state (`isTyping`, `typingTimeout`) was stored in plain `let` variables inside the component body. Those reset on every re-render.

Because `Textarea` subscribes to the full `WebchatContext`, it re-renders frequently (bot messages, agent typing indicator, emoji picker, focus changes, etc.). After each re-render, `isTyping` was `false` again, so every subsequent keystroke could send another `typing_on`.

Additionally:

- `stopTyping()` always sent `typing_off`, even when not actually typing
- No `typing_off` was sent on input blur
- Each event was forwarded immediately to `HubtypeService.postMessage` with no deduplication
- Async races could produce duplicate or out-of-order signals (e.g. send finishing after a new typing session had already started)

## Approach

Two complementary layers:

### Layer 1 — UI session (`use-typing-session.ts` + `textarea.tsx`)

`Textarea` delegates typing lifecycle to `useTypingSession`, which persists state across re-renders with refs:

- `isTyping` — whether the UI considers the user in an active typing session
- `intentVersion` — bumps whenever the user starts or stops typing, so stale async work is ignored
- `idleTimer` — 20s (`TYPING_IDLE_MS`) inactivity before requesting `typing_off`
- `pendingTypingOn` — shares one in-flight `typing_on` request while delivery is pending

Guards and triggers:

- `startTyping()` sends `typing_on` only once per session; retries if delivery failed
- `stopTyping()` only sends `typing_off` when the user's latest intent is still to stop
- `stopTyping()` on blur, Enter, clear, idle timeout, and unmount cleanup
- Enter decouples `stopTyping()` from `sendTextAreaText()` so a slow/failed send cannot kill a newer typing session

### Layer 2 — Network sender (`typing-network-sender.ts` + `use-typing-chat-event-sender.ts`)

`InputPanel` wires `Textarea` to `useTypingChatEventSender`, which owns a `TypingNetworkSender` instance:

- `typing_on` — sent immediately, but skipped if the server already knows the user is typing; cancels any pending `typing_off`; concurrent starts share a single in-flight request
- `typing_off` — debounced by 2s (`TYPING_OFF_DEBOUNCE_MS`) to coalesce rapid stop signals (blur + clear + send)
- `typing_off` always uses the same session context as its matching `typing_on`
- `dispose()` on unmount sends `typing_off` immediately; if `typing_on` is still pending, it sends `typing_off` as soon as that request resolves
- Failed `onUserInput` calls do not mark the session as active, allowing retry on the next keystroke

### Normal typing session

```
User types first character
  -> useTypingSession: startTyping()
  -> TypingNetworkSender: typing_on sent (1 POST)

User keeps typing (many keystrokes + re-renders)
  -> useTypingSession: startTyping() skipped
  -> TypingNetworkSender: typing_on skipped
  -> no extra POSTs

User stops typing for 20s
  -> useTypingSession: stopTyping() on idle timeout
  -> TypingNetworkSender: schedules typing_off (2s debounce)
  -> typing_off sent (1 POST)
```

Total: **2 POSTs**

### Blur then refocus quickly

```
User types
  -> typing_on sent

User blurs input
  -> useTypingSession: stopTyping()
  -> TypingNetworkSender: typing_off scheduled (2s)

User refocuses and types again within 2s
  -> TypingNetworkSender: pending typing_off cancelled
  -> typing_on skipped (still active)

Total: still only 1 typing_on so far
```

### Send message with Enter

```
User types
  -> typing_on sent

User presses Enter
  -> textarea: stopTyping() + sendTextAreaText() independently
  -> TypingNetworkSender: typing_off scheduled (2s)
  -> typing_off sent after debounce

Total: 2 POSTs
```

### Start next message before send completes

```
User types "hi" and presses Enter
  -> stopTyping() requested, sendTextAreaText() starts (slow)

User immediately starts typing the next message
  -> intentVersion invalidates the pending stop
  -> existing typing session stays active
  -> no premature typing_off / typing_on pair
```

### Before vs after (buggy re-render scenario)

```
BEFORE:
  keystroke -> typing_on
  bot update -> re-render -> isTyping = false
  keystroke -> typing_on again
  Result: N typing_on events for one message

AFTER:
  keystroke -> typing_on
  re-render -> isTypingRef preserved
  keystroke -> skipped
  Result: 1 typing_on + 1 typing_off
```

## Files changed

- `textarea.tsx` — thin UI wrapper; delegates typing lifecycle to `useTypingSession`
- `use-typing-session.ts` — ref-backed typing session with idle timeout, intent guards, and cleanup
- `typing-network-sender.ts` — deduplicated network delivery with debounced `typing_off`
- `use-typing-chat-event-sender.ts` — React hook that bridges `onUserInput` to `TypingNetworkSender`
- `input-panel/index.tsx` — wires `sendChatEvent` through the new hook
- `tests/webchat/textarea-typing.test.js` — textarea + integration coverage
- `tests/webchat/typing-chat-event.test.js` — network sender + hook coverage
- `tests/webchat/use-typing-session.test.js` — async race coverage for UI session

## Test plan

- [ ] Open webchat, type a message slowly — verify only one `typing_on` in network tab
- [ ] Keep typing while bot/agent events arrive — verify no additional `typing_on` calls
- [ ] Stop typing for 20s — verify one `typing_off` fires ~2s after idle timeout
- [ ] Blur the input while typing — verify `typing_off` is sent
- [ ] Blur and quickly refocus + type — verify no duplicate `typing_on` / `typing_off` pair
- [ ] Send message with Enter — verify `typing_off` follows `typing_on`
- [ ] Clear input with Backspace — verify `typing_off` is sent when field becomes empty
- [ ] Press Enter and immediately start typing the next message — verify typing session is not reset by the previous send
- [ ] Close/unmount webchat while typing — verify `typing_off` is sent immediately


packages/botonic-react/src/webchat/input-panel/textarea.tsx
```diff
@@ -1,107 +1,68 @@
...
-  const clearTimeoutWithReset = (reset: boolean) => {
-    const waitTime = 20 * 1000
-    if (typingTimeout) {
-      clearTimeout(typingTimeout)
-    }
-    if (reset) {
-      typingTimeout = setTimeout(stopTyping, waitTime)
+      // Stop typing immediately; do not wait for sendTextAreaText to finish.
+      void stopTyping()
+      void sendTextAreaText()
```
Iru89: Is the function `async`? Why are you using `void` instead of `await`?

packages/botonic-react/src/webchat/input-panel/textarea.tsx
```diff
@@ -1,107 +1,68 @@
...
       <TextareaAutosize
         ref={(ref: HTMLTextAreaElement) => {
           textareaRef.current = ref
         }}
-        onFocus={handleFocus}
-        onBlur={handleBlur}
+        onFocus={() => setIsInputFocused(true)}
+        onBlur={() => {
+          setIsInputFocused(false)
+          void stopTyping()
```
Iru89: same

packages/botonic-react/src/webchat/input-panel/typing-network-sender.ts
```diff
@@ -0,0 +1,164 @@
...
+    this.cancelPendingOff()
+
+    if (!this.isTypingOnServer || !this.activeContext) {
+      return
+    }
+
+    // Unmount should not leave the server thinking the user is still typing.
+    const context = this.activeContext
+    this.resetTypingState()
+    void this.sendToServer(Typing.Off, context)
```
Iru89: same here, why is the reason to use void and not await?

vanbasten17:
feedback applied @Iru89

Iru89 (approved):

